PR #6, opened 2016-08-13
Elasticsearch Support

hernandev:
Initial Elasticsearch support, tests on the way.

Some options may differ and I may make a custom config key for customizing the query parameter, Elastic can be really complex but this setup should cover most scenarios.


tomschlick:
Why close this?

hernandev:
https://github.com/laravel/scout/pull/7

PR #7, opened 2016-08-13
Initial Elasticsearch support

hernandev:
Reopened since squashing would include other poeples commits from the upstream sync.

Initial Elasticsearch support, tests on the way.

Some options may differ and I may make a custom config key for customizing the query parameter, Elastic can be really complex but this setup should cover most scenarios.

Working right now:

- Searching
- Searching with where conditions
- Pagination
- Unit Testing :)

TODO:
- For future implementations, for both algolia and elastic, may the performSearch be a replaceable callback,so people can extend and customize it, gonna send a separate PR on it

### DIFF
--- a/src/EngineManager.php
+++ b/src/EngineManager.php
@@ -11,7 +11,7 @@ class EngineManager extends Manager
     /**
      * Get a driver instance.
      *
-     * @param  string  $driver
+     * @param  string  $name
      * @return mixed
      */
     public function engine($name = null)
@@ -31,6 +31,11 @@ class EngineManager extends Manager
         ));
     }
 
+    /**
+     * Create an Elasticsearch engine instance.
+     *
+     * @return Engines\ElasticsearchEngine
+     */
     public function createElasticsearchDriver()
     {
         return new Engines\ElasticsearchEngine(
--- a/src/EngineManager.php
+++ b/src/EngineManager.php
@@ -35,7 +35,8 @@ class EngineManager extends Manager
     {
         return new Engines\ElasticsearchEngine(Elasticsearch::create()
             ->setHosts(config('scout.elasticsearch.hosts'))
-            ->build()
+            ->build(),
+            config('scout.elasticsearch.index')
         );
     }
 
--- a/src/EngineManager.php
+++ b/src/EngineManager.php
@@ -33,9 +33,8 @@ class EngineManager extends Manager
 
     public function createElasticsearchDriver()
     {
-        return new Engines\ElasticsearchEngine(Elasticsearch::create()
-            ->setHosts(config('scout.elasticsearch.hosts'))
-            ->build(),
+        return new Engines\ElasticsearchEngine(
+            Elasticsearch::fromConfig(config('scout.elasticsearch.config')),
             config('scout.elasticsearch.index')
         );
     }
--- a/src/EngineManager.php
+++ b/src/EngineManager.php
@@ -4,6 +4,7 @@ namespace Laravel\Scout;
 
 use Illuminate\Support\Manager;
 use AlgoliaSearch\Client as Algolia;
+use Elasticsearch\ClientBuilder as Elasticsearch;
 
 class EngineManager extends Manager
 {
@@ -28,6 +29,14 @@ class EngineManager extends Manager
         return new Engines\AlgoliaEngine(new Algolia(
             config('scout.algolia.id'), config('scout.algolia.secret')
         ));
+    }
+
+    public function createElasticsearchDriver()
+    {
+        return new Engines\ElasticsearchEngine(Elasticsearch::create()
+            ->setHosts(config('scout.elasticsearch.hosts'))
+            ->build()
+        );
     }
 
     /**
--- a/src/Engines/ElasticsearchEngine.php
+++ b/src/Engines/ElasticsearchEngine.php
@@ -28,14 +28,14 @@ class ElasticsearchEngine extends Engine
      */
     public function update($models)
     {
-        foreach($models as $model) {
+        $models->each(function ($model) {
             $this->elasticsearch->index([
-                'index' =>  'laravel',
-                'type'  =>  $model->searchableAs(),
-                'id'    =>  $model->getKey(),
-                'body'  =>  $model->toSearchableArray(),
+                'index' => 'laravel',
+                'type' => $model->searchableAs(),
+                'id' => $model->getKey(),
+                'body' => $model->toSearchableArray(),
             ]);
-        }
+        });
     }
 
     /**
@@ -46,13 +46,13 @@ class ElasticsearchEngine extends Engine
      */
     public function delete($models)
     {
-        foreach($models as $model) {
+        $models->each(function ($model) {
             $this->elasticsearch->delete([
                 'index' => 'laravel',
                 'type' => $model->searchableAs(),
                 'id'  => $model->getKey(),
             ]);
-        }
+        });
     }
 
     /**
--- a/src/Engines/ElasticsearchEngine.php
+++ b/src/Engines/ElasticsearchEngine.php
@@ -70,10 +70,10 @@ class ElasticsearchEngine extends Engine
      */
     public function search(Builder $query)
     {
-        return $this->performSearch($query, array_filter([
+        return $this->performSearch($query, [
             'filters' => $this->filters($query),
             'size' => $query->limit ?: 10000,
-        ]));
+        ]);
     }
 
     /**
@@ -174,7 +174,9 @@ class ElasticsearchEngine extends Engine
         }
 
         $keys = collect($results['hits']['hits'])
-                        ->pluck('_id')->values()->all();
+                    ->pluck('_id')
+                    ->values()
+                    ->all();
 
         $models = $model->whereIn(
             $model->getKeyName(), $keys
@@ -182,7 +184,6 @@ class ElasticsearchEngine extends Engine
 
 
         return collect($results['hits']['hits'])->map(function ($hit) use ($model, $models) {
-
             return $models[$hit['_source'][$model->getKeyName()]];
         });
     }
--- a/src/Engines/ElasticsearchEngine.php
+++ b/src/Engines/ElasticsearchEngine.php
@@ -1,0 +1,182 @@
+<?php
+
+namespace Laravel\Scout\Engines;
+
+use Laravel\Scout\Builder;
+use Elasticsearch\Client as Elasticsearch;
+use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Support\Collection as BaseCollection;
+
+class ElasticsearchEngine extends Engine
+{
+    /**
+     * Create a new engine instance.
+     *
+     * @param  \Elasticsearch\Client  $elasticsearch
+     * @return void
+     */
+    public function __construct(Elasticsearch $elasticsearch)
+    {
+        $this->elasticsearch = $elasticsearch;
+    }
+
+    /**
+     * Update the given model in the index.
+     *
+     * @param  Collection  $models
+     * @return void
+     */
+    public function update($models)
+    {
+        foreach($models as $model) {
+            $this->elasticsearch->index([
+                'index' =>  'laravel',
+                'type'  =>  $model->searchableAs(),
+                'id'    =>  $model->getKey(),
+                'body'  =>  $model->toSearchableArray(),
+            ]);
+        }
+    }
+
+    /**
+     * Remove the given model from the index.
+     *
+     * @param  Collection  $models
+     * @return void
+     */
+    public function delete($models)
+    {
+        foreach($models as $model) {
+            $this->elasticsearch->delete([
+                'index' => 'laravel',
+                'type' => $model->searchableAs(),
+                'id'  => $model->getKey(),
+            ]);
+        }
+    }
+
+    /**
+     * Perform the given search on the engine.
+     *
+     * @param  Builder  $query
+     * @return mixed
+     */
+    public function search(Builder $query)
+    {
+        return $this->performSearch($query, array_filter([
+            'filters' => $this->filters($query),
+            'size' => $query->limit,
+        ]));
+    }
+
+    /**
+     * Perform the given search on the engine.
+     *
+     * @param  Builder  $query
+     * @param  int  $perPage
+     * @param  int  $page
+     * @return mixed
+     */
+    public function paginate(Builder $query, $perPage, $page)
+    {
+        $from = (($page * $perPage) - $perPage);
+
+        $result = $this->performSearch($query, [
+            'filters' => $this->filters($query),
+            'size' => $perPage,
+            'from' => $from,
+        ]);
+
+        $result['nbPages'] = (int) ceil($result['hits']['total'] / $perPage);
+
+        return $result;
+    }
+
+    /**
+     * Perform the given search on the engine.
+     *
+     * @param  Builder  $query
+     * @param  array  $options
+     * @return mixed
+     */
+    protected function performSearch(Builder $query, array $options = [])
+    {
+        $filterQuery = [
+            "filter" => [
+                "query_string" => [
+                    "query" => "*{$query->query}*",
+                ],
+            ]
+        ];
+
+        if (array_key_exists('filters', $options) && $options['filters']) {
+            $filterQuery = array_merge($filterQuery, [
+                "must" => [
+                    "match" => $options['filters'],
+                ],
+            ]);
+        }
+
+        $searchQuery = [
+            'index' =>  'laravel',
+            'type'  =>  $query->model->searchableAs(),
+            'body' => [
+                'query' => [
+                    "bool" => $filterQuery,
+                ],
+            ],
+        ];
+
+        if (array_key_exists('size', $options)) {
+            $searchQuery = array_merge($searchQuery, [
+                'size' => $options['size'],
+            ]);
+        }
+
+        if (array_key_exists('from', $options)) {
+            $searchQuery = array_merge($searchQuery, [
+                'from' => $options['from'],
+            ]);
+        }
+
+        return $this->elasticsearch->search($searchQuery);
+    }
+
+    /**
+     * Get the filter array for the query.
+     *
+     * @param  Builder  $query
+     * @return array
+     */
+    protected function filters(Builder $query)
+    {
+        return $query->wheres;
+    }
+
+    /**
+     * Map the given results to instances of the given model.
+     *
+     * @param  mixed  $results
+     * @param  \Illuminate\Database\Eloquent\Model  $model
+     * @return Collection
+     */
+    public function map($results, $model)
+    {
+        if (count($results['hits']) === 0) {
+            return Collection::make();
+        }
+
+        $keys = collect($results['hits']['hits'])
+                        ->pluck('_id')->values()->all();
+
+        $models = $model->whereIn(
+            $model->getKeyName(), $keys
+        )->get()->keyBy($model->getKeyName());
+
+
+        return collect($results['hits']['hits'])->map(function ($hit) use ($model, $models) {
+
+            return $models[$hit['_source'][$model->getKeyName()]];
+        });
+    }
+}
--- a/src/Engines/ElasticsearchEngine.php
+++ b/src/Engines/ElasticsearchEngine.php
@@ -72,7 +72,7 @@ class ElasticsearchEngine extends Engine
     {
         return $this->performSearch($query, array_filter([
             'filters' => $this->filters($query),
-            'size' => $query->limit,
+            'size' => $query->limit ?: 10000,
         ]));
     }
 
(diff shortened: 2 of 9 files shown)


tomschlick:
Looks good, only thing is I would suggest making the index configurable. Instead of just `'laravel'` it should go off of environment variables / the config file. My app has a different index per customer so that would be super helpful.

tomschlick:
Also, Taylor will probably suggest that you don't add elasticsearch/elasticsearch to the require block, just to the suggestion block so people using Agolia don't have to download the elasticsearch composer package.

hernandev:
@tomschlick great, thanks for the idea!

hernandev:
@tomschlick unless i'm misteken its on require-dev only

tomschlick:
Oops! The diff viewer was only showing me the line above it and not the block header. Carry on! :) 👍

tomschlick:
FYI you don't have to keep squashing each time you commit anymore since github allows the team to do that at merge :)

hernandev:
Just included pagination, not really beauty since elastic does not gives page information, so some math needs to be done, gonna start refactoring the whole thing now. but looks like the support for it is now complete

hernandev:
thanks @tomschlick didn't knew that

tomschlick:
Regarding the index name: there is a `searchableAs()` method on the Searchable trait that you should be using to determine the index name.

hernandev:
@tomschlick i'm using that to determine the 'type' field on elastic search

tomschlick:
Ah ok. Sorry missed that.

--
Tom Schlick




On Sat, Aug 13, 2016 at 6:28 PM -0400, "Diego Hernandes" <notifications@github.com> wrote:












@tomschlick i'm using that to determine the 'type' field on elastic search

src/Engines/ElasticsearchEngine.php
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * Update the given model in the index.
+     *
+     * @param  Collection  $models
+     * @return void
+     */
+    public function update($models)
+    {
+        foreach($models as $model) {
```
jmarcher: use the each() method form collections

src/Engines/ElasticsearchEngine.php
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * Remove the given model from the index.
+     *
+     * @param  Collection  $models
+     * @return void
+     */
+    public function delete($models)
+    {
+        foreach($models as $model) {
```
jmarcher: use the each() method form collections

hernandev: indeed, sending in a bit

hernandev: done

src/EngineManager.php
```diff
@@ -30,6 +31,14 @@ public function createAlgoliaDriver()
         ));
     }
 
+    public function createElasticsearchDriver()
+    {
+        return new Engines\ElasticsearchEngine(Elasticsearch::create()
```
nWidart: You could use the `ClientBuilder` class from Elasticsearch:

```
ClientBuilder::fromConfig(config('scout.elasticsearch'))
```

adiachenko:
@hernandev Just to add on the topic of configurable index and why it's useful, here is another valid use case I've seen in some apps:

```
'index' => 'search_' . app()->getLocale(),
```

alexbowers:
How would this handle AWS which is on version 1.5, and Elastic.co which is on 2.3?

tomschlick:
However the current elasticsearch/elasticsearch library would handle it. 
If that's not compatible it should be pretty trivial to fork this one and set the library to an earlier compatible version. 

--
Tom Schlick




On Sun, Aug 14, 2016 at 10:13 AM -0400, "Alex Bowers" <notifications@github.com> wrote:












How would this handle AWS which is on version 1.5, and Elastic.co which is on 2.3?

src/EngineManager.php
hernandev: that's client builder, aliased

hernandev:
@alexbowers from the php library readme

> If you are using Elasticsearch 1.x or 2.x, prefer using the Elasticsearch-PHP 2.0 branch. The 1.0 branch is compatible however.

hernandev:
@adiachenko Elasticsearch docs says that for performance, each index becomes a different shard, and that elastic would handle better single index with multiple types, that's why laravel is the prefixed index, gonne make index a configurable key so people can custom it if needed at run time as well, but would be better if everyone just overwrite the seachableAs() method to customized the type key instead of index key

paulocastellano:
Great job!

src/EngineManager.php
nWidart: But you're hard coding the hosts config option. 
There is no way here for me to set additional options on the ES client.

hernandev: ok, noticed that option now, could be a great option for custom setups. thanks @nWidart

nWidart: No problem, good job on the PR! You were faster than me to send it. 😄

hernandev:
@nWidart you comments gonna be hidden since that code is not active anymore, included a config array on the config, so you can pass specific configuration there instead of the hard coded config it was using before.

Also, for those who requested, index name is a separate config option

src/Engines/ElasticsearchEngine.php
```diff
@@ -0,0 +1,189 @@
...
+     * @return mixed
+     */
+    public function paginate(Builder $query, $perPage, $page)
+    {
+        $from = (($page * $perPage) - $perPage);
+
+        $result = $this->performSearch($query, [
+            'filters' => $this->filters($query),
+            'size' => $perPage,
+            'from' => $from,
```
jmarcher: This is the only place where you use the $from variable, no need for it to be a local variable.

jmarcher:
Looking great so far

src/Engines/ElasticsearchEngine.php
hernandev: the math was a little bit hard to read inline, gonna try again

jmarcher: you can reduce it to:
```
($page - 1) * $perPage
```

hernandev: you're dammed right!

src/Engines/ElasticsearchEngine.php
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Remove the given model from the index.
+     *
+     * @param  Collection  $models
+     * @return void
+     */
+    public function delete($models)
+    {
+        $models->each(function ($model) {
+            $this->elasticsearch->delete([
```
intech: No! Only BULK!

tomschlick:
Any update on what still needs to be done to get this rolling? Is it truly just unit tests?

hernandev:
@tomschlick for what i've tested, its fully functional, one little comment on it is that elasticsearch will return 10000 records limit, but no one should even be scanning the whole result set without pagination of  some sort. I'm travelling right now, gonna try to finish the unit test this sunday/monday

tomschlick:
Awesome! Great work on this BTW! 👍

src/Engines/ElasticsearchEngine.php
hernandev: @intech care to elaborate?

intech: @hernandev great pleasure to help you in September. Already 3 years developing a laravel + es

jmarcher:
Any news on this?

src/Engines/ElasticsearchEngine.php
VinceG: @hernandev i think he means that instead of each the bulk operation should be used.
https://www.elastic.co/guide/en/elasticsearch/guide/current/bulk.html

VinceG:
Is there a way to safely merge this into a branch where we can test?

VinceG:
@hernandev there might be an issue with this implementation. Did anyone test this out?

It appears that using this engine vs. Algolia returns incorrect results. For example searching for '5161' which is part of an address returns the attached image, where Algolia returned relevant results.

Elasticsearch searches in all fields, but for some reason it wont search in the String type fields.

screen shot 2016-08-23 at 12 01 52 pm
screen shot 2016-08-23 at 12 01 35 pm

src/Engines/ElasticsearchEngine.php
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * Perform the given search on the engine.
+     *
+     * @param  Builder  $query
+     * @return mixed
+     */
+    public function search(Builder $query)
+    {
+        return $this->performSearch($query, array_filter([
```
jmarcher: Correct me if I'm wrong, but I find this array_filter() pointless because we know the values are not null. performSearch already takes care of this, if this stays some performSearch conditions are pointless

src/Engines/ElasticsearchEngine.php
```diff
@@ -0,0 +1,189 @@
...
+    {
+        if (count($results['hits']) === 0) {
+            return Collection::make();
+        }
+
+        $keys = collect($results['hits']['hits'])
+                        ->pluck('_id')->values()->all();
+
+        $models = $model->whereIn(
+            $model->getKeyName(), $keys
```
jmarcher: Only one argument per line if the function call is multiline

peterfrankjohnson:
Just to let you know I've been using this today and haven't found any problems.

However I had copied it into my own github account and pulled from upstream to ensure that everything was up to date.

There were a couple of merge conflicts in src/SearchableScope.php and src/Engines/ElasticsearchEngine.php but these were trivial to resolve.

@hernandev If you need any help writing the unit tests I would be happy to lend a hand.

hernandev:
@peterfrankjohnson thanks for the awesome feedback, gonna merge and work on the conflicts right away, any help is welcomed

src/Engines/ElasticsearchEngine.php
hernandev: @jmarcher gonna work on the style now, thanks

hernandev:
@VinceG how are you passing the data to the search() method? integer os numeric string? some conversion may be necessary on my implementation, havent tested against a lot of data yet, planning to do so today

src/Engines/ElasticsearchEngine.php
hernandev: @jmarcher indeed, even now that max results are enforced when not present

VinceG:
@hernandev Doing further testing i found out it actually does find the results, but it's not the first one to come up. Since scout by default indexes every single column in the database it searches in all of those columns as well and assigns a better score to some records based on columns that shouldn't be the first one to be considered as a highest score.

For example in my case we have property address and loan number. Property address should be considered in search as well as loan number but property address should be in a higher priority for returned results rather than loan number. For example when searching directly through the console it returns as expected.

I think Algolia somehow knows to do that since it actually returned the correct results, more relevant ones. 

In general i think Scout should have more filters available rather than just simple where() implementation. Algolia and Elasticsearch are different in the way filters are passed in so this needs to be abstracted somehow. 


```
Vincents-MacBook-Pro-2:~ vincentgabriel$ curl -XPOST 'localhost:9200/laravel/_search?pretty' -d '
> {
>   "query": { "match": { "propaddress1": "5161" } },
>   "_source": ["id", "propaddress1"]
> }'
{
  "took" : 4,
  "timed_out" : false,
  "_shards" : {
    "total" : 5,
    "successful" : 5,
    "failed" : 0
  },
  "hits" : {
    "total" : 122,
    "max_score" : 4.9124317,
    "hits" : [ {
      "_index" : "laravel",
      "_type" : "appr_order",
      "_id" : "200024576",
      "_score" : 4.9124317,
      "_source":{"propaddress1":"5161 Lankershim","id":200024576}
    }, {
      "_index" : "laravel",
      "_type" : "appr_order",
      "_id" : "200007603",
      "_score" : 4.7902904,
      "_source":{"propaddress1":"5161 Lankershim","id":200007603}
    }, {
      "_index" : "laravel",
      "_type" : "appr_order",
      "_id" : "200007670",
      "_score" : 4.7902904,
      "_source":{"propaddress1":"5161 Lankershim","id":200007670}
    }, {
      "_index" : "laravel",
      "_type" : "appr_order",
      "_id" : "200008431",
      "_score" : 4.7167063,
      "_source":{"propaddress1":"5161 Lankershim","id":200008431}
    }, {
      "_index" : "laravel",
      "_type" : "appr_order",
      "_id" : "200015744",
      "_score" : 4.6966953,
      "_source":{"propaddress1":"5161 lankershim","id":200015744}
    }, {
      "_index" : "laravel",
      "_type" : "appr_order",
      "_id" : "200045178",
      "_score" : 4.5933485,
      "_source":{"propaddress1":"5161 lankershim","id":200045178}
    }, {
      "_index" : "laravel",
      "_type" : "appr_order",
      "_id" : "141232586",
      "_score" : 3.9299455,
      "_source":{"propaddress1":"5161 Lankershim Blvd","id":141232586}
    }, {
      "_index" : "laravel",
      "_type" : "appr_order",
      "_id" : "200005904",
      "_score" : 3.9299455,
      "_source":{"propaddress1":"5161 Lankershim Blvd","id":200005904}
    }, {
      "_index" : "laravel",
      "_type" : "appr_order",
      "_id" : "200004468",
      "_score" : 3.9299455,
      "_source":{"propaddress1":"5161 Landershim Blvd","id":200004468}
    }, {
      "_index" : "laravel",
      "_type" : "appr_order",
      "_id" : "200005508",
      "_score" : 3.9299455,
      "_source":{"propaddress1":"5161 Lankershim Blvd","id":200005508}
    } ]
  }
}
```

hernandev:
@VinceG i think you're looking for how the relevance on elasticsearch works, may this topic helps you out:

https://www.elastic.co/guide/en/elasticsearch/guide/current/scoring-theory.html

Not so sure about defining the relevance to be queried against the fields, but for sure elasticsearch has more query options, the only think i can think of right now is somehow implementing the query performer as a callback so you can assign a different logic to the client to make your queries, of course not for elastic or algolia the options will not met everyones needs, so we should default on the more general case and let people customize it without having to fork it.

Lets solve one issue at time, first getting this implementation working, then thinking about that, any ideas on it would be appreciated while that of couse. thanks

VinceG:
@hernandev I agree. i was just putting in my two cents. Thanks for the contribution, looking forward to it.

hernandev:
General Update, some style was fixed, merged with upstream for new code conflict solving and now we have Unit Testing.

@taylorotwell as general terms, as I've stated in a edit of the first PR comment, only thing that may be cool of people using it is to customize the performSearch method on a default but extendable callback, so the logic may change between application more easily.

Unless that's something specific that needs to be changed, this is ready to be merged so tested by a wide range of people.

Thanks!

src/Engines/ElasticsearchEngine.php
intech: @VinceG yes, you are right!

intech:
@VinceG @hernandev about the search a little late. It is necessary to consider both the mapping and fuzzing parameters in your queries. Elasticsearch very functional search engine and there are a lot of settings and nuances. He also regularly updated and often something does not work as it should, on this we talk to the developers on the forum. Make changes to releases.

VinceG:
@intech That's right. we currently use elasticsearch already, but we won't be able to just migrate over to Scout without it supporting custom search functions and attributes. I think once this PR is merged we can discuss further implementation of filters. In a perfect world the Eloquent where* should work the same with search and it can be adapted based on the search engine.

So i'd personally love to just do

```
Order::search('5161')->whereIn('...')->whereLike('...')->orderBy('...')->get();
```

hernandev:
@intech thanks for the awesome changes, already merged.

So, my guess is there is no one-size-fits-all approach when using elasticsearch, the only solution i could think of was to bind a callback that the package uses, so you can create your own performSearch method, would that solve the problems you've stated?

intech:
I spent a lot of time @VinceG attempts to create a universal package, studied all the existing ones, helping with their development, but always came to the fact that the requests themselves to SQL and Elastic very different in its architecture.
I'll give you a simple example of life.
We have users, we do have a group of users: administrators, moderators, buyers. In SQL, we would have realized the connection between the tables and the further development of scaled horizontally.
But elastica is wrong. It creates index "users" and types: administration, moderator, customer. And when I'm working with queries, I can work as a separate group /users/customer/_search and with all users /users/_search. This is necessary for proper scaling and high load.

VinceG:
@intech I agree. We have the same issue, i think that if we can make Scout just more flexible in terms of filtering and searching it'll at least make it somewhat useable in the real world applications. As it stands right now most applications just can't use it out of the box. I like the concept and this approach so i'll be definitely helping with this once it's merged.

hernandev:
@intech  @VinceG I know the expectations for more complex and robust options would be nice, but let's try making it functional for general and easy use-cases at first, there are performance, relevance and other topics really important but lets try do do it incrementally instead, agreed?

hernandev:
@intech still gonna review and merge your changes for bulk ioperations on this initial PR

intech:
@hernandev @VinceG now comes first DSL for search queries - https://github.com/ongr-io/ElasticsearchDSL. And here's an interesting moment with mapping in requests depend on it a lot of them on the types of documents field.

VinceG:
@intech that's awesome.

intech:
@VinceG or this https://github.com/sleimanx2/plastic#searching-model-content ;)

hernandev:
@intech looks awesome indeed

sleimanx2:
Hey I've been working on this [package](https://github.com/sleimanx2/plastic) for a while now.

There is a good amount of similarity with scout however plastic covers a lot of elastic specific queries and mappings.

We started a conversation [here](https://github.com/sleimanx2/plastic/issues/19) about this topic If you guys think that we can move plastic to become a scount driver please let me know.

I think at least we can use some common features like the result to model mapping.

VinceG:
@sleimanx2 I love the idea and the advanced functionality. I doubt that'll merge that project into scout as that is ES specific. they need to build these engines abstracted so the functionality remains the same when switching between engines. For example if i use Plastic and then decide to switch to Algolia i'll have to re-write the search code. While Scout tries to solve this problem by abstracting everything, we can "borrow" certain functionality from Plastic but i am not sure if @taylorotwell will merge such a large package into Scout.

$0.02

intech:
@VinceG In this case, we need to convince @taylorotwell that Elasticsearch :factory: much more promising and popular than service Algolia :baby: ))

tomschlick:
I think plastic should be a seperate PR. I say let this one go through with the basic functionality and then create a new driver called plastic and hook in new methods where you need them.

Right now though, we just need to get ES in place so we can actually use it.

jmarcher:
I think we should keep this simple, if someone needs more functionality they can use Plastic.
Anyway instead of asking Taylor to change the actual driver we could just decorate the actual one

sleimanx2:
I'm also not sure if it's a good idea to build elastic DSL queries using SQL syntax like where and whereIn. Using DSL terminology makes it easier to understand and debug the query if you're referencing the official docs or have a prior experience with elastic.

intech:
https://github.com/ErickTamayo/laravel-scout-elastic clone project

VinceG:
@intech i like that one. it uses bulk operations.

VinceG:
@sleimanx2 True. but at least map some of the basic functionality, it'll make things much easier to transition from one engine to another. i am not saying to abstract the whole thing, just the basic where clauses.

tomschlick:
Is this ready to go? 

IMO the stuff above can be a seperate pull request... I just want the ability to use ES now that scout has been released for over a week.

taylorotwell:
I've merged this and people can iterate on it from here. I'm not an Elastic search expert so have to just trust the community on this thing.

hernandev:
thanks @taylorotwell 
I believe people can improve it over time now

mehrdaad:
@hernandev I'm sorry can this be used with elasticsearch 5.x cluster?

jmarcher:
@mehrdaad no, as elasticsearch/elasticsearch 2.2 supports elasticsearch >= 2.0, < 5.0

hernandev:
@mehrdaad, as @jmarcher said, not currently, since there is no stable release of 5.0 yet. So the stable release of php driver for ElasticSearch don't support it

mehrdaad:
@hernandev thanks. One more question, I know scout is mainly to submit a query and get results.  
But what is your approach if you desire to get a more complicated query using DSL?  
for example geosearch, fuzzy searching, etc.  
Right now (although seems seely is that I get it done with elasticsearch php client and leave model syncing to scout.

mehrdaad:
@hernandev I see in ``ElasticSearchEngine``, in ``performSearch`` method, there is a ``filters`` argument. Would you please describe what it does? or provide some examples how it works?

terion-name:
It was merged, but now is not present in the package. Why?

jmarcher:
It is now only supported by external driver

PR #10, opened 2016-08-14
Fixed command docs

jmarcher:


taylorotwell:
Would still need a @return void. I've fixed it.

PR #11, opened 2016-08-15
add support for TNTSearch

nticaric:


taylorotwell:
I would suggest making this a package first that people can pull in. Similar to how notification drivers are being written.

stokic:
@taylorotwell is that an "official" recommendation for other drivers? I was thinking about adding a driver for Sphinx Search unless someone beats me to it :)

PR #14, opened 2016-08-15
New Algolia's features

JanPetr:
Hello Taylor,

Greetings from Algolia and a big thank you for your work!

This PR implements 3 features Algolia that are usually necessery during a search implementation:
- filters (for facet filtering, different compate and logical operators, ...)
- setting index settings directly via Scout
- slaves (for creating replicas with different sorting strategy)

This PR adds two methods:
- `filters()` to the Builder class which accepts a string and sets it to the Builder public variable.
   The filter parameter can handle numericFilters but also non numeric ones (https://www.algolia.com/doc/api-client/php/parameters#filters) which make it more powerful.
- `setSettings` to the Searchable trait which takes model's `$searchSettings` and push it to the index

`setSettings` throws a `BadMethodCallException` when Engine does not support `setSettings`. Not sure if it's the best approach, just let me know if that works for you.

Filters usage example:
```php
Task::search('car')->filters('wheels > 4 AND color: red')->get();
```

Model's settings variable example:
```php
public $searchSettings = [
    'attributesToIndex' => [
        'id',
        'name',
    ],
    'customRanking' => [
        'desc(popularity)',
        'asc(name)',
    ],
    'synonyms' => [
        [
            'objectID' => 'red-color',
            'type'     => 'synonym',
            'synonyms' => ['red', 'another red', 'yet another red']
        ]
    ],
    'slaves' => [
        'my_slave1' => [
            'ranking' => [
                'desc(id)',
                'typo',
                'geo',
                'words',
                'proximity',
                'attribute',
                'exact',
                'custom'
            ],
        ],
        'my_slave2' => [
            'ranking' => [
                'asc(id)',
                'typo',
                'geo',
                'words',
                'proximity',
                'attribute',
                'exact',
                'custom'
            ],
        ],
    ],
];
```

Tests are included.

Let us know what you think about this.
❤️ from Algolia

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -36,6 +36,13 @@ class Builder
     public $wheres = [];
 
     /**
+     * The "filters" constraints added to the query.
+     *
+     * @var string
+     */
+    public $filters = '';
+
+    /**
      * The "limit" that should be applied to the search.
      *
      * @var int
@@ -78,6 +85,20 @@ class Builder
     public function where($field, $value)
     {
         $this->wheres[$field] = $value;
+
+        return $this;
+    }
+
+
+    /**
+     * Add a constraints to the search query.
+     *
+     * @param  string  $filters
+     * @return $this
+     */
+    public function filters($filters)
+    {
+        $this->filters = $filters;
 
         return $this;
     }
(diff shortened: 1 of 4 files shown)


morloderex:
Maybe better to split these settings up into multiple properties?

taylorotwell:
The `filters` thing really would leak implementation details out of the abstraction that I'm not sure we want to be leaking. You have to specify them as a string in a very Algolia specific way in this PR.

taylorotwell:
This will have to be re-worked to not be a leaky abstraction I think. I also think the settings thing should be a separate PR.

PR #19, opened 2016-08-19
SetSettings feature

JanPetr:
Hello there,
as requested in #14, I separed `setSettings` feature to a separate PR.

This PR implements:
- setting index settings directly via Scout
- slaves (for creating replicas with different sorting strategy)

This PR adds two method `setSettings` to the Searchable trait which takes model's `$searchSettings` and pushes it to the index.

`setSettings` throws a `BadMethodCallException` when Engine does not support `setSettings`. 

Model's settings variable example:
```php
public $searchSettings = [
    'attributesToIndex' => [
        'id',
        'name',
    ],
    'customRanking' => [
        'desc(popularity)',
        'asc(name)',
    ],
    'synonyms' => [
        [
            'objectID' => 'red-color',
            'type'     => 'synonym',
            'synonyms' => ['red', 'another red', 'yet another red']
        ]
    ],
    'slaves' => [
        'my_slave1' => [
            'ranking' => [
                'desc(id)',
                'typo',
                'geo',
                'words',
                'proximity',
                'attribute',
                'exact',
                'custom'
            ],
        ],
        'my_slave2' => [
            'ranking' => [
                'asc(id)',
                'typo',
                'geo',
                'words',
                'proximity',
                'attribute',
                'exact',
                'custom'
            ],
        ],
    ],
];
```

Test is included.

### DIFF
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -2,12 +2,16 @@
 
 namespace Laravel\Scout\Engines;
 
+use Illuminate\Database\Eloquent\Model;
 use Laravel\Scout\Builder;
 use AlgoliaSearch\Client as Algolia;
 use Illuminate\Database\Eloquent\Collection;
 
 class AlgoliaEngine extends Engine
 {
+    /** @var \AlgoliaSearch\Client */
+    private $algolia;
+
     /**
      * Create a new engine instance.
      *
@@ -17,6 +21,16 @@ class AlgoliaEngine extends Engine
     public function __construct(Algolia $algolia)
     {
         $this->algolia = $algolia;
+    }
+
+    /**
+     * Gets instance of Algolia's API Client
+     *
+     * @return \AlgoliaSearch\Client
+     */
+    public function getAlgoliaClient()
+    {
+        return $this->algolia;
     }
 
     /**
@@ -154,5 +168,43 @@ class AlgoliaEngine extends Engine
     public function getTotalCount($results)
     {
         return $results['nbHits'];
+    }
+
+    /**
+     * Sets settings to Algolia index.
+     * https://www.algolia.com/doc/api-client/php/settings#set-settings
+     *
+     * @param  \Illuminate\Database\Eloquent\Model  $model
+     * @param  mixed  $settings
+     * @return void
+     */
+    public function setSettings(Model $model, $settings)
+    {
+        $index = $this->algolia->initIndex($model->searchableAs());
+
+        if (isset($settings['synonyms'])) {
+            foreach (array_chunk($settings['synonyms'], 100) as $synonyms) {
+                $index->batchSynonyms($synonyms, true, true);
+            }
+
+            unset($settings['synonyms']);
+        }
+
+        $slavesSettings = [];
+        if (isset($settings['slaves'])) {
+            $slavesSettings = $settings['slaves'];
+            $settings['slaves'] = array_keys($slavesSettings);
+        }
+
+        $index->setSettings($settings, true);
+
+        if (!empty($slavesSettings)) {
+            unset($settings['slaves']);
+
+            foreach ($slavesSettings as $slaveName => $slaveSettings) {
+                $slaveSettings = array_merge($settings, $slavesSettings[$slaveName]);
+                $this->algolia->initIndex($slaveName)->setSettings($slaveSettings);
+            }
+        }
     }
 }
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -3,6 +3,7 @@
 namespace Laravel\Scout;
 
 use Laravel\Scout\Jobs\MakeSearchable;
+use BadMethodCallException;
 use Illuminate\Database\Eloquent\Collection;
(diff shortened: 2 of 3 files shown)


src/Engines/AlgoliaEngine.php
```diff
@@ -134,4 +148,48 @@ public function map($results, $model)
...
+
+        if (isset($settings['synonyms'])) {
+            foreach (array_chunk($settings['synonyms'], 100) as $synonyms) {
+                $index->batchSynonyms($synonyms, true, true);
+            }
+
+            unset($settings['synonyms']);
+        }
+
+        $slavesSettings = [];
```
lucasmichot: Why not just:
```php
if (isset($settings['slaves'])) {
    $slavesSettings = $settings['slaves'];
    $settings['slaves'] = array_keys($slavesSettings);
}
```

?

src/Engines/AlgoliaEngine.php
```diff
@@ -2,17 +2,21 @@
...
 class AlgoliaEngine extends Engine
 {
+    /** @var Client */
+    private $algolia;
+
     /**
      * Create a new engine instance.
      *
-     * @param  \AlgoliaSearch\Client  $algolia
+     * @param Client  $algolia
```
lucasmichot: I think all Laravel projects use FQN in docblocks

src/Engines/AlgoliaEngine.php
```diff
@@ -2,17 +2,21 @@
...
 
+use AlgoliaSearch\Client;
 use Laravel\Scout\Builder;
 use AlgoliaSearch\Client as Algolia;
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Support\Collection as BaseCollection;
 
 class AlgoliaEngine extends Engine
 {
+    /** @var Client */
```
lucasmichot: I think all Laravel projects use FQN in docblocks

src/Searchable.php
```diff
@@ -78,6 +78,23 @@ public static function makeAllSearchable()
...
+     *
+     * @return void
+     */
+    public static function setSettings()
+    {
+        $model = new static();
+        $engine = $model->searchableUsing();
+
+        if (!method_exists($engine, 'setSettings')) {
+            throw new \BadMethodCallException('Search engine "'.get_class($engine).'" does not support setSettings() call.');
```
lucasmichot: `\BadMethodCallException`should be imported.

src/Searchable.php
```diff
@@ -78,6 +78,23 @@ public static function makeAllSearchable()
     }
 
     /**
+     * Save settings to search engine
+     *
+     * @return void
```
lucasmichot: Missing `@throws \BadMethodCallException` docblock.

src/Engines/AlgoliaEngine.php
```diff
@@ -134,4 +148,48 @@ public function map($results, $model)
...
+
+    /**
+     * Sets settings to Algolia index.
+     * https://www.algolia.com/doc/api-client/php/settings#set-settings
+     *
+     * @param  \Illuminate\Database\Eloquent\Model  $model
+     * @param  mixed  $settings
+     * @return void
+     */
+    public function setSettings($model, $settings)
```
lucasmichot: Model should be type-hinted maybe?
`public function setSettings(Model $model, $settings)`

JanPetr:
Thanks for your feedback @lucasmichot, incporporated 🎉

taylorotwell:
Can all of this be defined in the Algolia UI?

taylorotwell:
Would prefer people just use the Algolia UI.

elfeffe:
@taylorotwell Yes, but there is a lot of reasons to use it with the API.
You can use Laravel GUI to add settings, instead of give users acces to Algolia GUI.
Also is really needed to add synonyms dinamically.
This is the reason why I don't use Scout, and keep using Algolia package.

julienbourdeau:
We created a package to manage your settings easily. This allow you to export/import settings from a json file, using an artisan command.
https://github.com/algolia/laravel-scout-settings

@elfeffe Feel free to try and let me know if you see any improvement to make 👍

elfeffe:
@julienbourdeau ok, I will check it

rrrhys:
Just adding a use case for laravel side configuration of Algolia indices, I'm building out a multi-tenant site that uses restricted indices so each tenant can only search their index.

It's unmanageable to add faceting, filtering, ranking parameters for each index manually through Algolia frontend. All tenants will use the same settings.

As with @elfeffe  above I had to revert to the old Algolia library (even though they recommend use of Scout now)

julienbourdeau:
@rrrhys Have you checked this package: https://github.com/algolia/laravel-scout-settings

If you register the `AlgoliaSearch\Client` in the container, setting settings when an index is created should not require much code. 

Feel free to open a thread on http://discourse.algolia.com/ I'd be happy to help 👍

rrrhys:
Thanks Julien,

I worked around it in a way I was satisfied with (using old lib) -- just thought instead of grumbling about Scout I should share my use case :)

PR #20, opened 2016-08-19
Extend search method to receive optional $args

JanPetr:
Hello again from Algolia,

as implementation of filters in #14 leaked abstration, here we implemented `filters` feature of Algolia in more abstract way which should suit more use-cases and do not leak abstraction only because Algolia.

`search()` method of Searchable trait now accepts second optional parameter `$args` which can be used to pass special arguments (like filters) to search. Those `$args` are stored within `Builder` and can be used in search engines while performing search.

Let us know what you think.

❤️ from Algolia

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -22,6 +22,13 @@ class Builder
     public $query;
 
     /**
+     * Additional search arguments
+     *
+     * @var array
+     */
+    public $args;
+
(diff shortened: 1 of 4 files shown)


taylorotwell:
I really want to just keep this opinionated to the main use case of simple full text searching.

PR #23, opened 2016-08-19
Changing config structure

KennedyTedesco:
This is just a proposal to follow: https://github.com/laravel/laravel/blob/master/config/queue.php

To me looks better to read and organize the drivers. The problem is that's a breaking change.

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -15,7 +15,7 @@ return [
     |
     */
 
-    'driver' => env('SCOUT_DRIVER', 'algolia'),
+    'default' => env('SCOUT_DRIVER', 'algolia'),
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
This actually isn't correct. You are hard-coding the "algolia" value in the method that creates the algolia driver. That isn't how cache drivers work for example. This doesn't truly support drivers as the cache system does. If it did you would need to allow arbitrary driver names and not hardcode algolia like that. Hope this makes sense.

KennedyTedesco:
It makes sense for sure. I have passed for a short lapse of bad judgment (or inattention). Sorry for that. 😊

Issue #21, opened 2016-08-19
Pagination issue


wcjonathan:
When retrieving results with pagination, the following error occurs..

`1/1ErrorException in Builder.php line 142:Undefined index: nbPages
in Builder.php line 142
at HandleExceptions->handleError('8', 'Undefined index: nbPages', '/Users/michaelevans/workspace/jargonbuster/jb/vendor/laravel/scout/src/Builder.php', '142', array('perPage' => '10', 'pageName' => 'page', 'page' => '1', 'engine' => object(TNTSearchEngine), 'results' => object(Collection), 'rawResults' => array('ids' => array(), 'execution time' => '7.0291 ms'), 'this' => object(Builder), 'paginator' => object(Paginator))) in Builder.php line 142
at Builder->paginate('10') in PagesController.php line 36
at PagesController->search(object(Request))
at call_user_func_array(array(object(PagesController), 'search'), array(object(Request))) in Controller.php line 52
at Controller->callAction('search', array(object(Request))) in ControllerDispatcher.php line 44
at ControllerDispatcher->dispatch(object(Route), object(PagesController), 'search') in Route.php line 189
at Route->runController() in Route.php line 144
at Route->run(object(Request)) in Router.php line 642
at Router->Illuminate\Routing\{closure}(object(Request)) in Pipeline.php line 53
at Pipeline->Illuminate\Routing\{closure}(object(Request)) in SubstituteBindings.php line 41
at SubstituteBindings->handle(object(Request), object(Closure)) in Pipeline.php line 137
at Pipeline->Illuminate\Pipeline\{closure}(object(Request)) in Pipeline.php line 33
at Pipeline->Illuminate\Routing\{closure}(object(Request)) in VerifyCsrfToken.php line 64
at VerifyCsrfToken->handle(object(Request), object(Closure)) in Pipeline.php line 137
at Pipeline->Illuminate\Pipeline\{closure}(object(Request)) in Pipeline.php line 33
at Pipeline->Illuminate\Routing\{closure}(object(Request)) in ShareErrorsFromSession.php line 49
at ShareErrorsFromSession->handle(object(Request), object(Closure)) in Pipeline.php line 137
at Pipeline->Illuminate\Pipeline\{closure}(object(Request)) in Pipeline.php line 33
at Pipeline->Illuminate\Routing\{closure}(object(Request)) in StartSession.php line 64
at StartSession->handle(object(Request), `

wcjonathan:
Fixed in #26

PR #26, opened 2016-08-21
Pagination Issue: nbPages undefined index fix

wcjonathan:
When requesting results using scout and the paginate method the request resulted in an undefined index on 'nbPages' #21

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -139,7 +139,7 @@ class Builder
         ]));
 
         return $paginator->appends('query', $this->query)
-                         ->hasMorePagesWhen($rawResults['nbPages'] > $page);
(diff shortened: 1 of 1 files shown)


dvlpp:
I dont understand this fix. I mean, `->hasMorePagesWhen(($results->count() / $perPage) > $page);` will always return false, since `$results->count()` is not total count, but the count of returned items...

For the Algolia engine, this should be changed to `->hasMorePagesWhen(($rawResults["nbHits"] / $perPage) > $page)`, but since there is no contract on how the total count should be returned by the engine, it could be different for other drivers. Maybe we can add a `totalCount()` method in the Engine contract?

PR #28, opened 2016-08-21
Fix docblocks and use FQN syntax

lucasmichot:
Fix docblocks and use FQN syntax


taylorotwell:
Has conflicts.

lucasmichot:
Fixed @taylorotwell

Issue #27, opened 2016-08-21
Error when performing a batch import


harrygr:
When trying to perform a batch import (e.g. `php artisan scout:import "App\Post"`) the following error is thrown

> Symfony\Component\Debug\Exception\FatalThrowableError: Type error: Argument 1 passed to Laravel\Scout\SearchableScope::Laravel\Scout\{closure}() must be an instance of Laravel\Scout\Builder, instance of Illuminate\Database\Eloquent\Builder given in /home/vagrant/samarkand-web/vendor/laravel/scout/src/SearchableScope.php:32

JanPetr:
Update to the latest version, the issue was fixed in https://github.com/laravel/scout/pull/22

harrygr:
I have  172d156 installed (2hrs old at time of writing) and it still gives that error

wcjonathan:
Have you include the use statement for the Searchable trait?

harrygr:
Yes, here is the relevant code:

```php
//...
use Laravel\Scout\Searchable;

class Product extends Model
{
    use Searchable;

//...
```

tomschlick:
Give this a shot now... looks like @talv86 fixed it with his PR

harrygr:
Yes looks like #33 did the trick

PR #33, opened 2016-08-21
Fix class names

talv86:
Fixes a bug that causes this error
```
[Symfony\Component\Debug\Exception\FatalThrowableError]                                                   
  Type error: Argument 1 passed to Laravel\Scout\SearchableScope::Laravel\Scout\{closure}() must be an ins  
  tance of Laravel\Scout\Builder, instance of Illuminate\Database\Eloquent\Builder given
```

Since an \Illuminate\Database\Eloquent\Builder doesn't extend Laravel\Scout\Builder to me the previous code will not work, these simple class name updates should fix the problem.

### DIFF
--- a/src/SearchableScope.php
+++ b/src/SearchableScope.php
@@ -29,7 +29,7 @@ class SearchableScope implements Scope
      */
     public function extend(EloquentBuilder $builder)
     {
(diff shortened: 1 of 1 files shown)


talv86:
This should fix #27 too since i had the same problem and don't with the above change

Issue #36, opened 2016-08-26
Result Pagination Type


edcs:
I've just implemented Scout on an API I've been building. I'm using Fractal by the PHP League which has baked in support for Laravel pagination using `\Illuminate\Pagination\LengthAwarePaginator`.

The documentation for Scout states the following (https://laravel.com/docs/5.3/scout#pagination):

> In addition to retrieving a collection of models, you may paginate your search results using the paginate method. This method will return a Paginator instance just as if you had paginated a traditional Eloquent query.

However, Scout returns an instance of `Illuminate\Pagination\Paginator` where as Eloquent returns an instance of `\Illuminate\Pagination\LengthAwarePaginator`. Before I spend too much effort on this, is there any reason why Scout returns a different type of paginator to Eloquent? This seems a bit like an issue since the documentation suggests that it's pagination should be the same as Eloquent's.

EdwardNelson:
Fixed with #37 it just needs to be tagged

EdwardNelson:
Fixed with #62 it just needs to be tagged

PR #37, opened 2016-08-26
Return length aware paginator

edcs:
Changes the paginator to `Illuminate\Pagination\LengthAwarePaginator` so that it matches Eloquent.

Following on from this issue: https://github.com/laravel/scout/issues/36

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Scout;
 
+use Illuminate\Pagination\LengthAwarePaginator;
 use Illuminate\Pagination\Paginator;
 use Illuminate\Database\Eloquent\Collection;
 
@@ -121,7 +122,7 @@ class Builder
      * @param  int  $perPage
      * @param  string  $pageName
      * @param  int|null  $page
-     * @return \Illuminate\Contracts\Pagination\Paginator
+     * @return \Illuminate\Contracts\Pagination\LengthAwarePaginator
      */
     public function paginate($perPage = 15, $pageName = 'page', $page = null)
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
This is not correct because $result->count() is just the results you got from this query. LengthAware paginator wants the total count of all records in the database.

eoghanobrien:
@taylorotwell, I understand why you've closed this, since calling `$result->count()` doesn't return the correct total.

However, it's not obvious how you would pass a total back from `Engine::paginate()` so that `Builder::paginate()` can work with it.

It's not as explicit as I'd like but maybe `Builder::paginate()` could check for a `nbHits` key when from the resulting array from `Engine::paginate()`?

For example, using @edcs's code as a base:

```php
public function paginate($perPage = 15, $pageName = 'page', $page = null)
{
    $engine = $this->engine();

    $page = $page ?: Paginator::resolveCurrentPage($pageName);

    $results = Collection::make($engine->map(
        $rawResults = $engine->paginate($this, $perPage, $page), $this->model
    ));

    // Here's the important part.
    $total = $rawResults['nbHits'] ?: $results->count();

    $paginator = (new LengthAwarePaginator($results, $total, $perPage, $page, [
        'path' => Paginator::resolveCurrentPath(),
        'pageName' => $pageName,
    ]));

    return $paginator->appends('query', $this->query);
}
```

Looking for the `nbHits` key with automatically get the `AlgoliaEngine` working. Other Engines would probably need to make their request first and then format the results array to be returned to `Builder::paginate()`.

What do you think?

edcs:
The docblock in `LengthAwarePaginator` states that `total()` returns:

>Get the total number of items being paginated.

So, surely if `$total` contains the total number of items being paginated, the total number of search results is a valid value?

PR #45, opened 2016-09-02
Use bulk API for elasticsearch inserts and actually test elasticsearch.

Skysplit:
There were errors in api calls for all elastiscsearch versions. This PR fixes it.  
It also adds actual tests against elasticsearch API and uses bulk api for inserts/deletes.

Perhaps even more tests would be required, as I already stumbled across various bugs

Travis Elasticsearch version is `1.4`. On my computer I was testing it using docker with Elasticsearch version `1.5` and `2.3.4`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,12 @@ php:
 
 sudo: false
 
+services:
+    - elasticsearch
+
 install: travis_retry composer install --no-interaction --prefer-dist
+
+before_script:
+    - sleep 10
 
 script: vendor/bin/phpunit --verbose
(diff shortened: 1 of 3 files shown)


Skysplit:
Sadly, build failed because elasticsearch didn't have enough time to process job and did not insert data on time.

intech:
@Skysplit refresh in bulk queries not recommend. Default interval index 1 second.

Skysplit:
Yeah, I was bit torn apart wheter to add `refresh` or not. Finally I decided to add this, as it will allow to instantly search on indexed values without waiting.

Issue #44, opened 2016-09-01
Customize Object ID


foxted:
I have been trying to index different resources into a single index to simplify search. 

By customizing the `toSearchableArray()` method, I did find a way to change the ObjectID stored in Algolia, but the `unsearchable()` method is broken now.

Is there a way to explicitly tell Scout to use a different ObjectID?

dennisoderwald:
+1

PR #47, opened 2016-09-03
Allow for custom IDs in indexes

foxted:
Related to #44.

Not sure if it is the best implementation possible, but I had to give it a try.

### DIFF
--- a/tests/Fixtures/AlgoliaEngineTestModelCustomKey.php
+++ b/tests/Fixtures/AlgoliaEngineTestModelCustomKey.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Tests\Fixtures;
+
+class AlgoliaEngineTestModelCustomKey extends TestModel
+{
+    public function getSearchableKey()
+    {
+        return "custom-{$this->getKey()}";
+    }
+
+    public function getReverseSearchableKey($key = '')
+    {
+        return str_replace('custom-1', '', $key);
+    }
+}
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -43,7 +43,6 @@ class AlgoliaEngine extends Engine
             if (empty($array)) {
                 return;
             }
-            g
             return array_merge(['objectID' => $model->getSearchableKey()], $array);
         })->filter()->values()->all());
     }
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -43,6 +43,7 @@ class AlgoliaEngine extends Engine
             if (empty($array)) {
                 return;
             }
+
             return array_merge(['objectID' => $model->getSearchableKey()], $array);
         })->filter()->values()->all());
     }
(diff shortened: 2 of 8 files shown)


taylorotwell:
Has merge conflicts.

foxted:
I'll fix it ASAP!

foxted:
Fixed :)

taylorotwell:
I don't understand why you need the reverse method?

foxted:
@taylorotwell Right now, without including the ID of the model in the index, Scout will not be able to index or retrieve the models. 

When customizing the searchable key (ObjectID), we can include the ID of the model along with the custom parts, for example, an `App\Post` model could have a ObjectID like `post-1`. The reverse function would allow the parsing of the ObjectID to find the original model ID back. 

I thought it would be useful. I don't mind taking it off, the important thing is being able to index different models under one unique index, to avoid having to query many indices, as the Algolia's Javascript library does not allow multi-indices query simultaneously yet.

tests/Fixtures/AlgoliaEngineTestModelCustomKey.php
```diff
@@ -0,0 +1,16 @@
...
+class AlgoliaEngineTestModelCustomKey extends TestModel
+{
+    public function getSearchableKey()
+    {
+        return "custom-{$this->getKey()}";
+    }
+
+    public function getReverseSearchableKey($key = '')
+    {
+        return str_replace('custom-1', '', $key);
```
taylorotwell: I don't think your tests are functioning how you expect. I can return literally any value from this method and the tests still pass.

foxted: It is pretty much the same test as `test_map_correctly_maps_results_to_models` but with a custom ObjectID. I am not directly testing the `getSearchableKey()` & `getReverseSearchableKey()` methods, as these methods are on the Fixture model class.

Just as the `test_map_correctly_maps_results_to_models` method, the model is mocked and so is the `getReverseSearchableKey` method, which explains why it would always return true even if you change the method on the Fixture. If you change the return of the mocked method though, it will fail:

`$model->shouldReceive('getReverseSearchableKey')->andReturn(2);`

Maybe `test_map_correctly_maps_results_to_models_with_custom_keys` is not needed since we already have `test_map_correctly_maps_results_to_models` as they are pretty much the same.

Also, the map method on the engine is fetching the list of ObjectID from the hits, which could be custom and not match the ID used in the database when the engine runs a query in the DB to fetch all the models within an array of ObjectIDs. 

So either the map method in the Engine class should list the ID instead of ObjectIDs, but the ID would be required in the searchable fields. Currently, the ID field is required when indexing anyway.

Or we need a method on the model to transform an ObjectID into the ID used in the DB.

foxted:
@taylorotwell After trying it out in a project of mine, it does work as expected. But I still have no clue about how to make the tests relevant :/

taylorotwell:
Closing until we can figure out tests. If you figure out the tests please re-submit.

elfeffe:
@foxted Did you make it work?

foxted:
I did, but it was not accepted, so I kind of gave up.

PR #48, opened 2016-09-03
add method getClient() for dynamic queries

intech:


Skysplit:
It would be nice if you implement this method in `AlgoliaEngine` as well

intech:
@Skysplit done!

taylorotwell:
There is an unrelated change here. Also the method should be at the bottom of the class not right under the constructor.

PR #49, opened 2016-09-05
Pagination simplification

jmarcher:
As the local variable is used only one time


taylorotwell:
Why did you change the calculation?

jmarcher:
I mean why use the same variable two times in the same line?

    (A * B) - B = (1 - A) * B

maybe is just a mater of taste but I find the right side better

taylorotwell:
Please use the same calculation.

jmarcher:
done

Issue #42, opened 2016-09-01
Restrict Indexing to a Subset of Your Data?


dennisoderwald:
I think Scout needs a function to select which entries are indexed. The implementation of Algolia's Laravel Plugin has an `indexOnly()` function.

https://github.com/algolia/algoliasearch-laravel#restrict-indexing-to-a-subset-of-your-data

pvanhemmen:
You can set the `hidden` property on the model for all fields that should not be indexed. At least this works for me in combination with the elasticsearch driver. All hidden fields will not be converted to JSON or arrays and thus not be pushed to the index. The `visible`property should work as well, but I haven't tested it yet.

https://laravel.com/docs/5.3/eloquent-serialization#hiding-attributes-from-json

driesvints:
You can use the `toSearchableArray` method.

PR #50, opened 2016-09-05
Restrict indexing of a model

dchristophersen:
This pr is fixing #42.

The engines `update` method will look for an `indexOnly` method on the model. If this method is present the model will only be indexed when `indexOnly` returns `true`

### DIFF
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -30,9 +30,25 @@ class AlgoliaEngine extends Engine
     {
         $index = $this->algolia->initIndex($models->first()->searchableAs());
 
-        $index->addObjects($models->map(function ($model) {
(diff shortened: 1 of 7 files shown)


taylorotwell:
This could be implemented by just having searchableAs return an empty array.

dennisoderwald:
@taylorotwell This does not solve the issue. Scout create a record only with a `objectID`. But that does not make sense, since it unnecessarily consumes storage.

taylorotwell:
No I’m saying you should submit a PR that doesn’t index the object if you return an empty array. I wasn’t saying it works that way now.

dennisoderwald:
Ah @taylorotwell, misunderstanding. :D I talk with @dchristophersen and we submit a new PR, when it's ready.

PR #56, opened 2016-09-08
Switch to simple_query_string for better exception handling in Elasticsearch

ameliaikeda:
Right now, if you query for, say, `category:foo`, you're gonna see an instance of `Elasticsearch\Common\Exceptions\BadRequest400Exception` thrown because `*category:foo*` is a syntax error in elasticsearch query strings.

All-in-all, `simple_query_string` (which cannot error and silently ignores bad syntax, returning 0 results on a proper non-transport-related error instead) seems to be a much better fit for "it just works".

### DIFF
--- a/src/Engines/ElasticsearchEngine.php
+++ b/src/Engines/ElasticsearchEngine.php
@@ -145,7 +145,7 @@ class ElasticsearchEngine extends Engine
         if (array_key_exists('filters', $options) && $options['filters']) {
             foreach ($options['filters'] as $field => $value) {
                 $searchQuery[] = [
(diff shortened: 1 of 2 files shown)


taylorotwell:
@tomschlick any opinions on this?

ameliaikeda:
Relevant docs page for this:

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html

tomschlick:
👍  on that. Otherwise we have to try / catch every call, this seems much easier and is officially supported.

PR #61, opened 2016-09-09
ElasticsearchEngine modifications to follow elasticsearch best practices

brino:
Corrected several common misconceptions about using elasticsearch:

- NEVER use query_string query ... it's not intended to be opened up to the end user! This allows the end user direct access to the underlying Lucene instance, with which they can easily  (in theory) do a query that would suck up all free JAVA_HEAP and crash elastic.
- Used match query for fulltext searches, and term filter for (numeric) filters
- Used "fuzzy" parameter on main match query to allow fuzzy matching and misspellings
- Used Bool query instead of Filtered query. the filtered query has been depricated as of Elasticsearch 2.0. Bool query is probably the best combo-query to use for most things
- Changed 'index' to be the $model->searchableAs() value prefixed with 'index' value, then set 'type' to static value 'docs'. Elasticsearch Indices should have a single type, unless you have a REALLY good reason to combine multiple types within the same index. In elasticsearch 2.x types are held within the same Apache Lucene instance, thus, you cannot have two fields with the same name but different data types within the same index. This means that an application could have 2 different models that both define the same field, but with different data types. The 2nd model would never index documents because of the field not matching the type elastic is expecting. It also means that, currently, all documents for all models are going into the same index. Which means you have to search through all docs from all models just to find one from a single model.

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -71,12 +71,8 @@ return [
     */
 
     'elasticsearch' => [
-        'index' => env('ELASTICSEARCH_INDEX', 'laravel'),
-
         'config' => [
-            'hosts' => [
-                env('ELASTICSEARCH_HOST')
-            ],
+            'hosts' => explode(',',env('ELASTICSEARCH_HOST','localhost')),
         ],
     ],
 
--- a/src/EngineManager.php
+++ b/src/EngineManager.php
@@ -42,8 +42,7 @@ class EngineManager extends Manager
     public function createElasticsearchDriver()
     {
         return new ElasticsearchEngine(
-            Elasticsearch::fromConfig(config('scout.elasticsearch.config')),
-            config('scout.elasticsearch.index')
+            Elasticsearch::fromConfig(config('scout.elasticsearch.config'))
         );
     }
 
--- a/src/Engines/ElasticsearchEngine.php
+++ b/src/Engines/ElasticsearchEngine.php
@@ -17,23 +17,14 @@ class ElasticsearchEngine extends Engine
     protected $elasticsearch;
 
     /**
-     * The index name.
-     *
-     * @var string
-     */
-    protected $index;
-
-    /**
      * Create a new engine instance.
      *
      * @param  \Elasticsearch\Client  $elasticsearch
      * @return void
      */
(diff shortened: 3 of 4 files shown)


brino:
I'd rather have the terms appear under a filter clause instead of must, but I think that only works for elasticsearch 2.x and so it fails the travis build

Skysplit:
Currently all results are being scored. I'm not an ES expert, so I'll just ask; do you think it is better to score every document than filtering documents first? `filtered` indeed is deprecated but works in 1.x as well as in 2.x.

Also, it would be nice if brackets for condition statements and lopps would follow psr coding standards:)

Overall - nice job!

brino:
Yes, it makes sense to filter wherever possible. That way you reduce the result set before fulltext searching. Seems like kind of a tossup as to which way to approach this though. The Filtered Query will be removed in the next major release of Elasticsearch. At which point there will not be a way to achieve backwards compatibility between the newest and oldest versions.

I suppose the best thing to do is use the filtered query as it DOES work just fine for now. Then, deal with the deprecation issues when the time comes.

I also changed the brackets to be on the same line as the if/foreach. :) As well as, added a fix for the deep paging error in elastic 2.x when from + size > 10000.

Thanks for the input!

brino:
I'd also like to point out that this query change allows for adding strings in a ->where() (not just numeric values like the scout docs say). It's not quite perfect, but it will work the vast majority of the time.

Also you can use dot notated names for any fields indexed while eager loaded. So for example ...

$articles has: 

protected $with = ['user'];

then you can: 

$articles->search('laravel')->with('user.name','billy')->get();

in contrast you could also just search for billy, and that would match in the user.name field as well, since the match query is against the "_all" field, user and all it's nested fields are included.

taylorotwell:
I have no idea if any of this is correct or not because I don't use ElasticSearch. Someone more knowledgable will have to review it.

brino:
That's fine, I was also thinking I could close this one, and submit several smaller PRs, as some of the things aren't all that controversial. Then everything could be discussed separately.

@taylorotwell Would you rather I did that?

taylorotwell:
That's probably a better idea.

PR #63, opened 2016-09-11
Return a Eloquent\Collection.

lucasmichot:
Returns the expected `\Illuminate\Database\Eloquent\Collection`.


taylorotwell:
Neither of the things you are collecting are actually arrays of Eloquent models. Only the final result is.

lucasmichot:
Fixed @taylorotwell

PR #64, opened 2016-09-11
Provide suggested packages versions

lucasmichot:
Provide suggested packages versions

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,8 +44,8 @@
         }
     },
     "suggest": {
-        "algolia/algoliasearch-client-php": "Required to use the Algolia engine.",
-        "elasticsearch/elasticsearch": "Required to use the Elasticsearch engine."
+        "algolia/algoliasearch-client-php": "Required to use the Algolia engine (^1.10).",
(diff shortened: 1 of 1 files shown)


brino:
Would this mean scout 'requires' elasticsearch 2.2 or greater, or just suggests it?

Point being, if elasticsearch 2.x is a requirement, we do a better job of crafting the query... namely, NOT using the deprecated Filtered query in favor of a Bool query with a Filter clause.

See PR #61 where I've been going back and forth with this.

lucasmichot:
> Would this mean scout 'requires' elasticsearch 2.2 or greater, or just suggests it?

It suggests them

> Point being, if elasticsearch 2.x is a requirement...

This is a dev requirement, anyone can use another driver, e.g. Algolia

> See PR #61 where I've been going back and forth with this.

Not sure these 2 PR are related

brino:
The point was just asking weather or not people are allowed to use elasticsearch 1.x ... or at least, at what point do we cut off support for older versions?

PR #70, opened 2016-09-13
Remove elasticsearch_index, use $model->searchableAs() instead

brino:
Removed elasticsearch_index, Elasticsearch [officially suggests] (https://www.elastic.co/blog/index-vs-type) using one single Type per Index unless you have a very good reason not to. This is because fields must be consistent between types. Therefor, you could never have 2 fields with the same name but different field types within two different models that use the Searchable trait. In that case elastic would silently fail to index the document with the NON-matching field type.

So, it makes more sense to use `$model->searchableAs()` as the "index" name, and then basically any string as the "type" name.

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -71,8 +71,6 @@ return [
     */
 
     'elasticsearch' => [
-        'index' => env('ELASTICSEARCH_INDEX', 'laravel'),
-
         'config' => [
             'hosts' => [
                 env('ELASTICSEARCH_HOST'),
--- a/src/Engines/ElasticsearchEngine.php
+++ b/src/Engines/ElasticsearchEngine.php
@@ -17,11 +17,9 @@ class ElasticsearchEngine extends Engine
     protected $elasticsearch;
 
     /**
-     * The index name.
-     *
      * @var string
      */
-    protected $index;
+    protected $type = 'doc';
 
     /**
      * Create a new engine instance.
@@ -29,11 +27,9 @@ class ElasticsearchEngine extends Engine
      * @param  \Elasticsearch\Client  $elasticsearch
      * @return void
      */
-    public function __construct(Elasticsearch $elasticsearch, $index)
+    public function __construct(Elasticsearch $elasticsearch)
     {
         $this->elasticsearch = $elasticsearch;
-
-        $this->index = $index;
     }
 
     /**
@@ -55,8 +51,8 @@ class ElasticsearchEngine extends Engine
 
             $body->push([
                 'index' => [
-                    '_index' => $this->index,
-                    '_type' => $model->searchableAs(),
+                    '_index' => $model->searchableAs(),
+                    '_type' => $this->type,
                     '_id' => $model->getKey(),
                 ],
             ]);
@@ -83,8 +79,8 @@ class ElasticsearchEngine extends Engine
         $models->each(function ($model) use ($body) {
             $body->push([
                 'delete' => [
-                    '_index' => $this->index,
-                    '_type' => $model->searchableAs(),
+                    '_index' => $model->searchableAs(),
+                    '_type' => $this->type,
                     '_id'  => $model->getKey(),
                 ],
             ]);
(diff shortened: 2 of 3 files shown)


mtdavidson:
While I agree with you that being able to use different index names makes more sense. I think there are lots of times where you'd want to define a type. A usecase that comes to mind is if you have an index with multiple types in some kind of parent child relationship ( https://www.elastic.co/guide/en/elasticsearch/guide/current/parent-child-mapping.html ) it would be good to be able to query both the parent and the child. Though I do appreciate this might be an advanced use case being able to keep things flexible would be great.

brino:
Good point @mtdavidson  very true ... Parent->child relationships MUST use different types within the same Index.

However, for the purposes of Scout, users are currently unable to do any kind of parent/child query or even index documents with special parent/child relationships (altho, you could probably get it to work OUTSIDE of scout). Furthermore, I'm not sure that Algolia offers any similar feature. So it might be a little awkward to have things in the builder/engine that some engines support and others don't.

That being said, I think it would be easy enough to implement a `setType()` method (or something) on the ElasticEngine class if and when the time came. Then you could use the `searchableAs()` method on the child model to force the same index as the "parent" model and `setType()` on both models to set the appropriate types.

From what I've seen @taylorotwell say, Scout is intended to be a quick and easy way to add full-text searching to your models. It doesn't currently seem to be part of the "vision" to add all kinds of advanced features and things.

Basically all I'm trying to say is, I believe reasons for making this change still stand, with or without Parent/Child. The models should each have their own Index in the vast majority of cases.

Thanks for the input, and I'm still very willing to discuss other approaches. Maybe @taylorotwell  can shed some light on where he sees scout going in the future?

taylorotwell:
This sounds fine for the basic use case. Scout is definitely supposed to be a basic search package. I have no intention of supporting every feature supported by Elastic since its a pretty vast product.

taylorotwell:
Would users need to update their applications after this change? Isn't it a breaking change and would require some preparation on the developer's part?

brino:
@taylorotwell The config changes, it no longer needs the index defined there. 
And, the data would need to be re-indexed ... `artisan scout:import 'App\Model'`

So yeah, I suppose that would be breaking?

hernandev:
Recommended reading on the subject: https://www.elastic.co/blog/index-vs-type

brino:
@hernandev is dead on... that's actually the same article I linked in the first comment.

Here is the important part: 

> **Which one should I use?**
> I will try to summarize the questions you should ask yourself to make a decision:
>
> - Are you using parent/child? If yes this can only be done with two types in the same index.
> - Do your documents have similar mappings? If no, use different indices.
> - If you have many documents for each type, then the overhead of Lucene indices will be easily amortized so you can safely use indices, with fewer shards than the default of 5 if necessary.
> - Otherwise you can consider putting documents in different types of the same index. Or even in the same type.

So basically what they are saying is... use different Indices, unless you have a specific use case NOT to. And, while @mtdavidson is correct about the Parent/child mappings, it probably isn't a use case Scout would handle out of the box.

taylorotwell:
Closing this since its breaking. All of the Elastic related PRs are gradually encouraging me to just discontinue official support for it and let the community maintain the drivers. It's just going to be more headache than its worth.

hernandev:
@taylorotwell as I did the initial support, I would encourage removal also. This PR is really simple actually but people are trying to support many Elastic features that goes against Scount idea.

Just keep in mind some easy to register drivers way, I could totally move the Elastic support to a separated packaged.

brino:
@taylorotwell I wholly agree with @hernandev I love the vision of scout and I think it's very useful in a lot of cases. Elastic fills the roll of a driver just fine. If users want more elastic functionality they can fork Scout or any one of the other packages that are already built around elastic.

As far as elastic being a scout driver though, this PR is kind of important IMHO. Field type collisions are a very possible issue without it.

What if I changed it to NOT require a config change? People would just have to reindex their data after updating... is that still considered breaking?

hernandev:
@brino not breaking means adding new features that does not change anything to the existing applications or fixing bugs, so I would say it is breaking.

The @taylorotwell problem is (I guess) not with your fix, while I'm inclined to agree on your approach, the main problem is several PR's trying to support the world of possibilities.

So, I'm still +1 on the removal, we both can work in a separated driver.

As soon this is here, there is work to manager and being polite saying no to all advanced features. 

IMHO.

brino:
I totally agree with all the advanced feature requests being out of scope. Technically this PR is removing functionality lol. 

Anyway, I'm inclined to agree that this would be breaking. I just hope it can be addressed at some point.

PR #76, opened 2016-09-20
Use "_id" property as a model key

dam1r89:
Use "_id" property as a model key instead of pulling it from the index body

This is solving a problem where toSearchableArray doesn't return a primary key.
Implementation is already passing model's key to the _id, so it is easy change.


taylorotwell:
toSearchableArray always need to return a primary key.

PR #77, opened 2016-09-22
Pass additional parameters to search engine

### LINKED ISSUE #12: Geo Search Option?
I was wondering if people would be interested in possibly a trait to say setup this entry to be able to be geo-searched on. As Algolia and Elasticsearch are separate in the way they do this, so it would require some setup in terms of how to get things setup for the array. If there is interest I can create a PR for this functionality.

EdwardNelson:
This PR allows additional parameters to be passed to the search engine. The reason I wanted this was to allow me to pass geo-search options to Algolia, I have therefore only implemented this for Algolia. I know Taylor has mentioned abstraction leakage in some earlier PRs but I really believe many features that make Algolia and Elasticsearch great were missing - such as Geo-search and tag search, this PR adds in the additional flexibility that is much needed. If this is wanted, I can look into Elasticsearch.

fixes #12

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -44,16 +44,25 @@ class Builder
     public $limit;
 
     /**
+     * Options that should be passed to the engine.
+     *
+     * @var array
+     */
+    public $options;
+
+    /**
      * Create a new search builder instance.
      *
      * @param  \Illuminate\Database\Eloquent\Model  $model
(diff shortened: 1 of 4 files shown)


taylorotwell:
I don't like these types of option arrays. Either use the Algolia SDK or add real support for more algolia features with real names.

PR #78, opened 2016-09-22
Add location method to search builder

EdwardNelson:
In response to feedback from #77. I have simply made this PR to see what it would feel like to use and what other people think. 

In the other PR I spoke about additional features that Algolia and Elasticsearch provide. When I had a think and stripped them down to the essentials and what is compatible between Algolia and Elasticsearch I found it was only Geosearch that was key.

I don't expect this to be merged, I just want this to start a discussion about whether scout is already supporting the maximum features it plans to.

### DIFF
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -105,6 +105,8 @@ class AlgoliaEngine extends Engine
      */
     protected function performSearch(Builder $builder, array $options = [])
     {
+        $options = $this->mergeLocation($builder, $options);
+
         return $this->algolia->initIndex(
             $builder->index ?: $builder->model->searchableAs()
         )->search($builder->query, $options);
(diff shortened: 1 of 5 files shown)


src/Engines/AlgoliaEngine.php
```diff
@@ -162,4 +164,19 @@ public function getTotalCount($results)
...
     }
+
+    /**
+     * Merge location data with options array to pass to Algolia.
+     *
+     * @param \Laravel\Scout\Builder $builder
+     * @param array $options
+     * @return array
+     */
+    private function mergeLocation (Builder $builder, array $options)
```
taylorotwell: Why is there a space after the function name?

EdwardNelson: My bad, fixed

EdwardNelson:
See #12 for more discussion

taylorotwell:
Holding off on this for now.

julienbourdeau:
What about now? It's an interesting feature I think.

Issue #58, opened 2016-09-09
Paginator returned by Builder hardcodes URL query key


mikeu:
In the `paginate` function of `Laravel\Scout\Builder`, the current search query is added to the paginator's URLs with the hardcoded key of `'query'`:

    return $paginator->appends('query', $this->query)

It would be great to be able to customize this key. For example, this would allow paginating search results with `...?q=terms` or `...?search=terms`.

----

One way to do this is by adding a fluent `withQueryKey` method to set a `$query_key` property on the `Builder`, allowing code such as

    $search_results = $model->search($search_query)
                            ->withQueryKey('q')
                            ->paginate(15);

I can create a PR implementing that, if there is interest.

powelski:
This is great idea, but I think it would be even better to make it work with a closure. This way, we could make our URLs whatever we want. I like having `/search/my-dog` URLs. Some may argue that it's not standard compatible, but I like it.

driesvints:
I can see this being useful. I'll try to send in a PR for this.

driesvints:
Sent in a PR for this. Please provide feedback https://github.com/laravel/scout/pull/311

jenky:
> Actually @nunomaduro just made the remark that you can just call `->appends('query', "overwrite here")` and it would overwrite the query parameter. In addition you can do something like `->appends('q', $query)` if you want a different key.

It returns `LengthAwarePaginator` so `appends` should work, but why does it have to be hardcoded like that. If page name can be customized then this `query` should not be added by default.

minkmk94:
> Actually @nunomaduro just made the remark that you can just call `->appends('query', "overwrite here")` and it would overwrite the query parameter. In addition you can do something like `->appends('q', $query)` if you want a different key.

In View i have Input name=search, and i use ->appends('search', request()->get('search'))->links() in bottom of view. But it still have ?query=something&search=something in my url, how can i remove param 'query' or customize it to 'search' ?

markomilivojevic:
Another solution is to remove `query` from query string by calling `->appends('query', null)` on your `->paginate()` method.

Harry-Harrison:
Thank you everyone in this thread!

So to anyone else, if you want to use `q` for your query parameter, for instance if `query` is disallowed by any other packages... *cough...* You need to do this.

```php
ModelName::search($request->q)->paginate(15)->appends('query', null)->appends('q', $request->q);
```

This will return your pagination urls without the `query` parameter and with the q parameter with the value of the request.

PR #82, opened 2016-09-27
Add option to override 'query' key when paginating

mandrean:
Related to issue #58.

Let's you override the `?query=` param in `next_page_url`/`prev_page_url` of the paginator if say your Laravel/Lumen app uses a different query parameter name for searches.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -133,7 +133,8 @@ class Builder
         $page = $page ?: Paginator::resolveCurrentPage($pageName);
 
         $results = Collection::make($engine->map(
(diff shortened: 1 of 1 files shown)


taylorotwell:
Would rather just stick to a convention than bloat up the parameters.

mandrean:
Any suggestions of an other way to override 'query' then, that is more OK?

Because surely it can't stay hardcoded to ' query' forever, right? I mean, some projects use `q=` or `search=` or whatever.

And search-replacing the response is rather ugly.

ajaybhargav:
👍 if you can add this option

Issue #83, opened 2016-09-27
Scout search doesn't map nested indexed relations in response


mandrean:
Let's say you index a model with some relations to Algolia or Elastic. Nested relations instead of separate indices.

If you search directly using Algolia/Elastic, your response will contain records with the entire structure (nested and all).

But Scout will only map the raw results to the main model class and not include its relations in the response.

mandrean:
PR: https://github.com/laravel/scout/pull/84

PR #84, opened 2016-09-27
Add support for mapping nested relational models in the response

mandrean:
Fix for #83.

Lets one specify relations to be mapped in the response if your index contains nested relations.

Usage:

```php
$relations = ['products', 'media'];
$orders = App\Order::search('Star Trek')->with($relations)->paginate();
```

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -16,6 +16,13 @@ class Builder
     public $model;
 
     /**
+     * Relations of the $model
+     *
+     * @var array
+     */
+    public $relations;
+
+    /**
      * The query expression.
      *
      * @var string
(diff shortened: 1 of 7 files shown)


taylorotwell:
Call the `->load` method on the collection to load the relations you want.

mandrean:
I'm sorry to hear that @taylorotwell.

Because then the returned object is no longer a `LengthAwarePaginator`, but a `Collection`. Fits badly if your project enforces `LengthAwarePaginator` as the strict return type.

A `->with()` would have been cleaner imho. Similar syntax to how `Illuminate\Database\Eloquent\Builder` already works. Would rather just stick to a convention, as you said 😉

But you're the boss!

PR #87, opened 2016-10-03
allow searching for all documents in the index

macku99:
Enable the ability to search across all documents in the index when using elasticsearch, inline with algolia engine.


taylorotwell:
No plans to continue updating the Elastic driver. Plan to let the community break it off and maintain it because it's too much confusion and overhead for me.

PR #92, opened 2016-10-08
enable the ability to eager load relations

macku99:
Enable the ability to eager load relations when mapping results to the given model.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -16,6 +16,13 @@ class Builder
     public $model;
 
     /**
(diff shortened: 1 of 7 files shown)


taylorotwell:
You can do it after you get your results by calling the `->load` method on the Collection.

macku99:
:) I wish I read the code more carefully. Maybe is worth to add this one on your trello board with items that need documentation.

I learned something new from this exercise.

Thanks Taylor.

PR #113, opened 2016-10-22
Any plan to support Lumen?

appkr:
The box content was not working under a Lumen project. I manage to make it work, so I'd like to share it with others. Thanks.

### DIFF
--- a/src/ScoutServiceProvider.php
+++ b/src/ScoutServiceProvider.php
@@ -2,9 +2,12 @@
 
 namespace Laravel\Scout;
 
+use Illuminate\Foundation\Application as LaravelApplication;
 use Illuminate\Support\ServiceProvider;
+use Laravel\Lumen\Application as LumenApplication;
 use Laravel\Scout\Console\FlushCommand;
 use Laravel\Scout\Console\ImportCommand;
+
 
 class ScoutServiceProvider extends ServiceProvider
 {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you explain why it doesn't work under Lumen in the first place?

appkr:
@taylorotwell `config_path()` helper was not available in Lumen. I think that was the root cause.

```sh
~/lumen-product $ php artisan
[Symfony\Component\Debug\Exception\FatalThrowableError]
  Call to undefined function Laravel\Scout\config_path()
```

taylorotwell:
I still don't want to run these commands if the application is not in console mode.

taylorotwell:
I don't really like this change to be honest. I don't like specificially detecting Lumen in the code like that.

appkr:
Ok then, how about adding the `config_path()` helper on Lumen framework? There was no way for Lumen users to make the scout work without that helper. Thanks~

appkr:
Let me post this for others who are researching on how to use scout in lumen project. The right solution was...

```php
// bootstrap/app.php

$app->register(Laravel\Scout\ScoutServiceProvider::class);
$app->configure('scout');
```

That's it.

RudyJessop:
This solution would work great in scenarios where you would need a laravel/external package, lumen is a mico-framework. Its meant to be lightweight initially.

PR #118, opened 2016-10-27
Return Algolia Search Highlights

stephenoldham:
Was thinking it'd be good to have access to the highlighted text in Algolia Search results. Especially now that the callback has been added allowing us to pass options in the `performSearch` method. This change conditionally adds the `_highlightResult` array to each result only if:

**a)** The array exists
**b)** A boolean var `$enableSearchHighlights` is explicitly set in the searchable model.

Signed-off-by: Steve Oldham stevo@vdotgood.com


taylorotwell:
I don't really like how this leaks Algolia stuff out into the model.

stephenoldham:
I thought that may be the case. Is there an alternative route you'd want to take?

PR #123, opened 2016-11-06
Add orderBy support to Builder

damiantw:
Adds an orders property to Builder. Generally you'll want to order by relevancy but not always. This allows drivers to be more flexible with ordered queries.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -51,6 +51,13 @@ class Builder
     public $limit;
 
     /**
+     * The "order" that should be applied to the search.
+     *
+     * @var int
+     */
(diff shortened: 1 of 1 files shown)


taylorotwell:
Why would you not want to sort by relevance?

damiantw:
A person made a request on my Laravel Scout MySQL driver package (https://github.com/damiantw/laravel-scout-mysql-driver/issues/4)

Looks like they wanted users to be able to search for products and sort them by price, but dates seems to be the use case that comes to mind. For example If one wanted to do a full text search on posts then sort by created_at so newer posts show first.

PR #125, opened 2016-11-09
$this->models is a collection : Should check via isEmpty()

lukepolo:


src/Jobs/MakeSearchable.php
```diff
@@ -35,7 +36,7 @@ public function __construct($models)
      */
     public function handle()
     {
-        if (empty($this->models)) {
+        if (!$this->models->count()) {
```
jhoff: `$this->models->isEmpty()` would read better IMO.

src/Jobs/MakeSearchable.php
```diff
@@ -13,17 +14,17 @@ class MakeSearchable implements ShouldQueue
     /**
      * The models to be made searchable.
      *
-     * @var \Illuminate\Database\Eloquent\Collection
+     * @var Collection
```
jhoff: Docblocks should have fully qualified class names.

lukepolo:
@jhoff thanks for opinions , updated~

jhoff:
Your PR tile is outdated now

Issue #122, opened 2016-11-06
elasticsearch : parsing_exception: no [query] registered for [filtered]


wardwar:
please help. how to solve this error?

another error

```json
{
	"error": {
		"root_cause": [{
			"type": "parsing_exception",
			"reason": "no [query] registered for [filtered]",
			"line": 1,
			"col": 22
		}],
		"type": "parsing_exception",
		"reason": "no [query] registered for [filtered]",
		"line": 1,
		"col": 22
	},
	"status": 400
}
```

kfchai:
I have just setup Elasticsearch 5.0 and Laravel Scout then came across this as well when I trigger get() function of the query builder. Realized that Elasticsearch 5.0 doesn't support **filtered** query, instead it has been replaced by **bool** query. 

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-filtered-query.html

I did a little modification in ElasticSearchEngine.php (line 177) to make it works:
```
$searchQuery = [
            'index' =>  $this->index,
            'type'  =>  $query->model->searchableAs(),
            'body' => [
                'query' => [
                    'bool' => [
                        'filter' => $termFilters,
                         'must' => $matchQueries,
                    ],
                ],
            ],
        ];
```

lamungu:
With ElasticSearch 5.0 released, should the driver be rightfully updated to abide to those changes?

moxoder:
Elasticsearch 5+ is not supported properly, please fix it.

rmff:
@wardwar 

Use temp fix suggested from @kfchai 

Inside: vendor\laravel\scout\src\Engines\ElasticsearchEngine.php
Method: performSearch
Replace $query by:
```
        $query = [
            'index' =>  $this->index,
            'type'  =>  $builder->model->searchableAs(),
            'body' => [
                'query' => [
                    'bool' => [
                        'filter' => $filters,
                        'must' => $matches,
                    ],
                ],
            ],
        ];
```

Gist for changed file [vendor\laravel\scout\src\Engines\ElasticsearchEngine.php](https://gist.github.com/rmff/f262ad5195b8fb9bf9aae9e61e323ced#file-vendor-laravel-scout-src-engines-elasticsearchengine-php)

jmarcher:
ping @hernandev
Any ideas?

moxoder:
Use your driver on the basis of basic Elasticsearch Engine through DI (service provider)

sl0wik:
[...]

PR #127, opened 2016-11-10
Add support to use Bool Query of Elasticsearch 5.0

kfchai:
Responding to issue #122, Elasticsearch 5.0 will responded "no [query] registered for [filtered]" with status 400 bad request. Added support to send the query body as bool query instead of filtered query depends on the version specified in scout.php.


taylorotwell:
I'm not going to make any further updates to the ElasticSearch driver. Make a community driver if you want different behavior and this one will be extracted. It's too much of a contentious thing to maintain in the core.

PR #134, opened 2016-11-17
Add ability to run MakeSearchable on non-default queue

zactax:
This PR allows the MakeSearchable job to run outside of the default queue. Adds a config setting called "on_queue" which is "default" by default.


taylorotwell:
Would rather it just be a method like the connection configuration. I don't like configuration bloat.

PR #136, opened 2016-11-23
Add getModel() to Scout Builder

mandrean:


pierrocknroll:
The getModel method should be in scout, could you make a pull request ?

Issue #138, opened 2016-12-10
Bypass mapping to Eloquent object and return keys of search results


lioreshai:
I'm working with an implementation of Scout where I immediately cache the results of a search (with Algolia  as the engine) and then feed it back to the user by using the whereIn and orderByRaw methods of the query builder as follows:

$query = $query->whereIn('id', $search_result_ids)->orderByRaw('FIELD(`products`.`id`,' . implode(',', $search_result_ids) . ')');

The reason I'm doing this is so I can put further constraints/run queries on the search results and still preserve the order of the results (as they were returned by Algolia).

As far as I can tell it is a great solution, but I would love to hear if anyone thinks there is a better one.

I noticed that there was no way to return just the result keys out of the box, since both the get() and paginate() messages map the search result keys to their respective Eloquent models. I didn't want to have results mapped on every search query, only for the app to pluck the ids and cache an array of keys.

I made the following changes to the package:

In Builder.php

    /**
     * Get the results of the search.
     *
     * @return \Illuminate\Database\Eloquent\Collection
     */
    public function get()
    {
        return $this->engine()->get($this);
    }

    /**
     * Get the keys of search results.
     *
     * @return \Illuminate\Database\Eloquent\Collection
     */
    public function getKeys()
    {
        return $this->engine()->getKeys($this);
    }

In Engines.php

    /**
     * Get the results of the given query mapped onto models.
     *
     * @param  \Laravel\Scout\Builder  $builder
     * @return \Illuminate\Database\Eloquent\Collection
     */
    public function get(Builder $builder)
    {
        return Collection::make($this->map(
            $this->search($builder), $builder->model
        ));
    }

    /**
     * Get the results of the query as a Collection of keys.
     *
     * @param  \Laravel\Scout\Builder  $builder
     * @return \Illuminate\Support\Collection
     */
    public function getKeys(Builder $builder)
    {
        return $this->getIds($this->search($builder));
    }

Each engine would need its own getIds method, like the following one I wrote for Algolia:

    /**
     *
     * Pluck and return the primary keys of the results.
     *
     * @param  mixed  $results
     * @return \Illuminate\Database\Eloquent\Collection
     */
    public function getIds($results) {

        return collect($results['hits'])
            ->pluck('objectID')->values()->all();

    }

If there's any interest in this feature I can package it up nicely and create a pull request. Either way, looking forward to hearing thoughts on this.

[...]

PR #139, opened 2016-12-10
Bypass mapping of Eloquent objects and return just the keys of results

lioreshai:
This is the getKeys feature I suggested here: https://github.com/laravel/scout/issues/138

It would be nice to have the option to return only ids of search results as opposed to the entire Eloquent collection.


lioreshai:
Awesome, thanks!

PR #141, opened 2016-12-15
Advanced where constraints on the Builder class.

tlaverdure:
This PR adds additional methods to the builder class that will allow advanced constraints to be implemented respectively by the engine driver. 

**Where methods:**
- `where($field, $value)`
- `orWhere($field, $value)`
- `whereBetween($field, array $value)`
- `orWhereBetween($field, array $value)`
- `whereIn($field, array $value)`
- `whereNot($field, $value)`
- `whereNot($field, $value)`
- `whereNotIn($field, array $value)`
- `whereNull($field)`
- `whereNotNull($field)`

Each of these where conditions are captured in arrays like so:

```php
$this->where['and'][] = [$field, $value];
``` 

``` php
[
    'featured' => 1,
    'topic' => 'security'
]
```

This would be most helpful and shouldn't cause any breaking changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,6 @@ php:
 
 sudo: false
 
-services:
-    - elasticsearch
-
 install: travis_retry composer install --no-interaction --prefer-dist
 
 before_script:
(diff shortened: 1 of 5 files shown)


taylorotwell:
I think I would rather just be a package somehow. Elastic is already going to be extracted.

tlaverdure:
So no to the where conditions?

PR #142, opened 2016-12-16
Added ability to search for/only soft deleted models

bencomeau:
Two options to interact with soft deleted models:

Include soft deleted models
`App\User::search('search query string')->withTrashed()->get();`

Return only soft deleted models
`App\User::search('search query string')->onlyTrashed()->get();`

Models must be deleted within the `withoutSyncingToSearch` callback to avoid the indices being updated during `delete()`.

**Note:** I am not sure if you would like to automate the `withoutSyncingToSearch` aspect by adding to the `ModelObserver` (checking if the model uses `SoftDeletes`, confirm not `forceDeleting`, etc.). If so, I will add that.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -58,6 +58,20 @@ class Builder
     public $orders = [];
 
     /**
+     * If the search should include soft deleted models.
+     *
+     * @var boolean
+     */
+    public $withTrashed = false;
+
+    /**
+     * If the search should only include soft deleted models.
+     *
+     * @var boolean
+     */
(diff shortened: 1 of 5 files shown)


taylorotwell:
I'm not sure this is exactly the implementation I would want. Will look into it.

xxRockOnxx:
Currently use `withoutSyncingToSearch` when deleting. But having `withTrashed` and `onlyTrashed` would be great. It's been a year, did laravel implemented a different solution to this?

PR #154, opened 2017-01-10
add config options to scout.php to allow specifying connection and queue

michaelst:


taylorotwell:
This contains a bunch of other wrong and irrelevant spacing changes. Feel free to re-submit without all of those.

andreladocruz:
@taylorotwell 

The Searchable.php has this functions:
```
/**
     * Get the queue connection that should be used when syncing.
     *
     * @return string
     */
    {
    }

    /**
     * Get the queue that should be used with syncing
     *
     * @return  string
     */
    {
    }
```
but this configurations doesn't exists in scout.php configuration file.
What is the correct way to config this values?
I changed my queue names in .env file but scout is still using the old one.
I think that something is missing here

PR #155, opened 2017-01-12
add config options to scout.php to allow specifying connection and queue

michaelst:


taylorotwell:
Tweaked this a hair... you should now do the following to customize this:

```php
    'queue' => [
        'connection' => 'foo',
        'queue' => 'bar',
    ],
```

michaelst:
Doing it that way, should there also be a true/false option? As that was the name of the previous config setting to enable or disable queueing.

Issue #153, opened 2017-01-10
How to index conditionally?


sandervanhooft:
Hi,

Is there a way to conditionally index a model record?

The search database will be publicly available, I'd rather not index records marked as "unpublished" by users.

I've tried this, but I see no difference in ElasticSearch results:

```
/**
     * Get the indexable data array for the model.
     *
     * @return array
     */
    public function toSearchableArray()
    {
        if(!$this->published) { // The condition
          $this->unsearchable();
          return;
        }
        $array = $this->toArray();
        //....
        return $array;
     }
```

Kind regards,

Sander

michelecurletta:
something like another criteria after [this condition](https://github.com/laravel/scout/blob/2.0/src/Searchable.php#L51).

```
`$models = $models->filter(function ($model, $key) {

    return $model->mustBeSearchable();

});`
```

Harti:
I created a gist to sort of alleviate the issue. 
Essentially I'm overriding a few methods of the Searchable trait.

https://gist.github.com/Harti/ec1fcd28b162bbb2e7fe06ce2df22275

sandervanhooft:
@Harti, Thanks! Your trait override seems the best maintanable solution for now.

sandervanhooft:
@Harti, What do you mean by this?

> Make sure to update your index afterwards (alternatively, update/save the models that you want to be affected).

Harti:
@sandervanhooft The respective search engine's indexing happens only when you store or update an eloquent model in your database, or when you run `php artisan scout:import "App\MyModel"`. 
If you only update the code without letting your search engine re-index the records, you will still find records that are supposed to be hidden.

My trait override simply prevents the models from being added to the search engine index.

Harti:
@sandervanhooft please take note that I updated my Gist to fix an issue with it on [line 39](https://gist.github.com/Harti/ec1fcd28b162bbb2e7fe06ce2df22275#file-searchable-L39).

jorenvanhee:
You could also return an empty array in the `toSearchableArray` method.

```
    public function toSearchableArray()
    {
        if ($something === true) {
            return [];
        }

        return $this->toArray();
    }
```

michelecurletta:
@taylorotwell this seems to be a common issue, what about adding this feature out of the box?

julienbourdeau:
I think returning an empty array is the way to go. I'm not sure what you can't do with that.

royduin:
[...]

PR #161, opened 2017-02-11
Include a closure for filtering searchable items (i.e. conditionally …

Harti:
…exclude them from being indexed)

See #153

### DIFF
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -89,6 +89,16 @@ trait Searchable
     }
 
     /**
+     * Closure that decides whether or not a Searchable model should be indexed by the Scout Engine.
+     *
+     * @return bool     true if it should be included in the index, false if it should be excluded
+     */
+    protected function shouldBeSearchable()
+    {
+        return true;
+    }
+
+    /**
      * Make all instances of the model searchable.
      *
      * @return void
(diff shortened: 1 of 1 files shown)


Harti:
Oh shoot, this is no default Builder object. My bad, nevermind.

src/Searchable.php
```diff
@@ -99,6 +109,10 @@ public static function makeAllSearchable()
 
         $self->newQuery()
             ->orderBy($self->getKeyName())
+            ->filter(function($model)
```
Harti: there should probably be a `->get()` beforehand

Harti:
On another note, @taylorotwell, while I do appreciate all the work you do, documentation on how to properly test code changes before making a pull request is really poor on the contribution guide. 

I was unable to properly adapt Mockery to mock using `->get()` to return a collection which would then be `filter()`ed and lastly the `searchable()` method be applied to the results. Well, I _was_ capable of editing the test case. But I'm pretty certain that's not how it is supposed to go.

Some PHP "peasants", myself included, are probably hesitant to push changes as they don't feel professional enough to make properly tested and easily integratable pull requests. 

I do actually think that my proposed change would only have to be adapted a tad bit to work swiftly with Laravel, but contributions like mine just won't find their way into your software if rookie developers fail to run a successful CI test. Just my two cents, I don't care if this makes it into any Scout release or not, I can always override the trait like I did in #153, for me it's more about sharing the knowledge and making it easier for all the other folks to achieve the same thing.

michelecurletta:
@taylorotwell what do you think about merging this PR?

PR #163, opened 2017-02-13
Allow for customizing searchable chunk size

KKSzymanowski:
Currently the chunk size is fixed at 100 which, at least for Elasticsearch, renders the import process quite slow.

For 100 000 records import time can be lowered 3 times by increasing the chunk size to about a thousand:
import_time

This change allows for setting a custom chunk size in `config/scout.php` as `searchableChunkSize` option.
It is fully backwards compatible and doesn't affect existing applications in any way as  far as I know.

### DIFF
--- a/src/SearchableScope.php
+++ b/src/SearchableScope.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Scout;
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'll just raise the default to 500 for now.

PR #174, opened 2017-03-30
Ability to add make all searchable "with trashed"

kfuchs:
This adds the ability to include trashed items when making all searchable.

Also adds the flag `--withTrashed` to the import command.

It will fail silently if the model does not implement a withTrashed method.

### DIFF
--- a/src/Console/ImportCommand.php
+++ b/src/Console/ImportCommand.php
@@ -13,7 +13,9 @@ class ImportCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'scout:import {model}';
+    protected $signature = 'scout:import
+                            {model}
+                            {--withTrashed : Whether soft deleted models should be included}';
 
     /**
      * The console command description.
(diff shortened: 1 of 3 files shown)


taylorotwell:
I think this would require more work, particular around the delete and restore observer. Going to hold off for now.

kfuchs:
Yeah, ultimately it would be nice to have a seamless way to pick and choose which models still get synced on delete.

For now the way I solve it is by creating my own model observer that on "deleting" it disables syncing, then on "deleted" it re-enables it and runs searchable. 

The code looks like this:

```
class UserObserver
{
	/**
	 * Listen to the User deleting event.
	 *
	 * @param User $user
	 */
	public function deleting(User $user)
	{
		$user::disableSearchSyncing();
	}

	/**
	 * Listen to the User deleted event.
	 *
	 * @param User $user
	 */
	public function deleted(User $user)
	{
		$user::enableSearchSyncing();

		$user->searchable();
	}
}
```

That pretty much works. Whats missing tho is the ability to import already trashed users. To fix that, it required me having to basically do what was in the commit.

PR #176, opened 2017-04-08
Return the result from callback in withoutSyncingToSearch

jmarcher:
I think that this method would be more useful if it returns the result from the callback


jmarcher:
@taylorotwell is it possible to Tag a version with this change? Thank you in advance

PR #177, opened 2017-04-10
Default null

tabuna:
Packages using the default scout will not receive an error.


taylorotwell:
Will just leave as-is for now.

PR #183, opened 2017-05-16
Add count() support to Builder

### LINKED ISSUE #164: Support count() method
Sometimes, it is useful to show users how many search results have been found. However, the count() function, which is used in Laravel QueryBuilder, doesn't exist in Scout QueryBuilder. It will be wonderful to have it.

Thanks!

julienbourdeau:
Fixes #164 

There is already `getTotalCount` but it simply extracts the data after performing full search. With `count()` we only retrieve the `nbHits` from Algolia.

To avoid BC break, we can have a default implementation based on `search` and `getTotalCount` in the abstract Engine class (instead of an abstract declaration). 

Please let me know if want me to make any modifications.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -159,6 +159,16 @@ class Builder
         return $this->engine()->get($this);
     }
 
+    /**
+     * Retrieve the number of matching results.
+     *
+     * @return int Number of results
+     */
+    public function count()
+    {
+        return (int) $this->engine()->count($this);
+    }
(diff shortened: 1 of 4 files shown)


taylorotwell:
We can't add functions to the abstract class in a patch release. It's a breaking change for all custom drivers.

Jimmea:
So. Please tell me , what should i do  this case ?

PR #191, opened 2017-05-30
Added the ability to extend scout builder

mauricecalhoun:


taylorotwell:
Looks like the Macroable trait that is already in Illuminate/Support.

Issue #190, opened 2017-05-30
Bypass mapping to Eloquent objects and directly return search results


jorenvanhee:
Would you accept a PR that makes it possible to directly get the search results? I would like to skip the mapping to eloquent objects.

Something like this: `Order::search('Stargate Universe')->getSearchResults();`

jorenvanhee:
#192 would also allow me to do this, if you wouldn't want this as core functionality.

Issue #114, opened 2016-10-22
How to return Raw Results?


n4p4:
Currently Scout allows for getting a collection of eloquent models as search results. However, what if I want the raw results from the search engine? How would I retrieve those?

martincarlin87:
I was about to ask this but decided to look through the current issues first. Would be good to have some kind of option.

PR #194, opened 2017-06-01
Add getSearchResults on the builder

jorenvanhee:
Adds a method to get the raw search results, unmapped, from the builder.

Relates to #190 & #114 

If accepted, I will also make a pull request for the 3.0 branch.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -159,6 +159,15 @@ class Builder
         return $this->engine()->get($this);
     }
 
+    /**
+     * Get the raw results of the search, not mapped onto models.
+     *
+     * @return mixed
+     */
+    public function getSearchResults()
+    {
+        return $this->engine()->search($this);
+    }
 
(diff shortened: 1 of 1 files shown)


julienbourdeau:
What do you think of the name `raw()` instead.

```php
Model::search($query)->get(); // Collection from Laravel
Model::search($query)->raw(); // Raw results from the engine
```

jorenvanhee:
I thought about this, and I'm not quite sure. The `get` method makes a lot of sense because it returns models, just like if you would call it on the eloquent query builder. You don't even have to know that `search` returns a scout builder, it just works.

`raw` on the other hand, does not exist on the eloquent query builder. So thats why I gave it a name that clearly refers to the search.

taylorotwell:
Send this to the 3.0 branch. 2.0 is not maintained anymore.

PR #197, opened 2017-06-02
Add numeric facet filters

dorin-niscu:


morloderex:
What does excatly does this do?

dorin-niscu:
In the current version we can only apply numeric filters: $orders = App\Order::search('Star Trek')->where('user_id', 1)->get();

I created two methods for numeric / facet filters so we can also use algolia facet filters:
$orders = App\Order::search('Star Trek')->where('locale', 'en')->get();

taylorotwell:
You can use new macroable feature to add this. I plan to keep core very simple.

dorin-niscu:
I understand.
Can you please provide us an example?
Thank you.

dorin-niscu:
I found a better way to apply facet filters using the search callback.
```
$orders = App\Order::search('Star Trek', function ($algolia, $query, $options) {
            $options = array_merge($options, [
                'facetFilters' => ['locale:en']
            ]);
            return $algolia->search($query, $options);
        })->get();
```

PR #207, opened 2017-07-11
Configure/optimize bulk chunk sizes

Artistan:
This option allows you to control if the size of the data that syncs
with your search engine. This chuck (page) size allows you to 
optimize the bulk data for both searchable and unsearchable methods.

This is very useful for optimizing data imports depending on machine size and data set sizes.

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -30,6 +30,22 @@ return [
 
     'prefix' => env('SCOUT_PREFIX', ''),
 
+	/*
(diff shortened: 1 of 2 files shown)


Artistan:
Thank you for the quick review and approval, this will be a big help for us. :)

xcaptain:
This is awesome, but for some sparse data like a table with minimum id is 1 maximum id is 1000000 just this change is still not very efficient.

PR #211, opened 2017-07-12
fix for php5.6, ternary operation …

Artistan:
should be sufficient since the $chunk is defined


Artistan:
this will fix the current travis-ci build failure

PR #214, opened 2017-07-25
Add a optional parameter to the paginate method.

TwoAvocet:
Allows the ability to change the returned url's query section.
search?q=searchTerm&page=1 can now be done by using paginate(null, 'page', null, 'q');


taylorotwell:
Hmm, probably will just keep the convention for now.

PR #217, opened 2017-08-22
Added Capability to walk through Relations and index indexable data

rodelarode:
I wanted to extend the indexer to to index relationship attributes which programmer can define in an array.

Example:
static $searchableAttributes = ['id','first_name', 'last_name', 'email', 'phone_no'];
    static $searchableRelationAttributes=[
        'account'=>[
            'name',
            'account_no'
        ],
        'user'=>[
            'initial',
            'name'
        ],
        'user.profile'=>[
            'initial'
        ]
    ];

Here I defined searchableAttributes static property to define indexable fields of the model and searchableRelationAttributes static property to define relationships and their corresponding fields to load for indexing. While creating index keys the '.' is replaced with '_' character.

This works so far with only one to on relationships.

Hope this helps to improve this feature. My intention for this pull request is to share what i did in my current project may be you guys got better ideas to improve this feature.

Cheers :)

### DIFF
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -5,9 +5,12 @@ namespace Laravel\Scout;
 use Laravel\Scout\Jobs\MakeSearchable;
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Support\Collection as BaseCollection;
+use Laravel\Scout\ModelObserver;
+use Laravel\Scout\SearchableScope;
 
 trait Searchable
 {
+
     /**
      * Boot the trait.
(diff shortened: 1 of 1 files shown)


taylorotwell:
You can already return relationships in your `toSearchableArray` method.

rodelarode:
Well, I relaise that. But actually is there any solution where i can index only specific columns of a realted object of any level. i.e. while indexing customer i want to index a column from a related user -> profile. Only that column and skip all other columns.

Issue #204, opened 2017-07-02
Is there a way to get facets?


mbouclas:
Both Algolia and ES support facets, is there a way for scout to query for them?

karamqubsi:
What I did for this .I used the raw method to get data and after that you can get whatever data you want from the returned result . 
Until this moment Scout doesn't support paginating raw results , I tried to solve that in this pull request #219

driesvints:
PR got merged by now so closing this.

PR #219, opened 2017-08-25
Add paginateRaw method to Builder class

karamqubsi:
Hi , 
While working with scout using Elastic-Search engine the only way to get all the data returned from Elastic-Search is to use the raw() method from scout Builder class , but we also need the results to be paginated so I added a simple method to handle that . 

I noticed that other people may find this useful i.e. issue #204 

By implementing this (in my case ) I can now get highlights and aggregations(facets) from Elastic-Search .

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -178,19 +178,21 @@ class Builder
      * @param  int  $perPage
      * @param  string  $pageName
      * @param  int|null  $page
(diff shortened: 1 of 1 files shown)


taylorotwell:
Don't pass a new flag. Just make the new method have a little duplication. It's not a big deal.

PR #221, opened 2017-08-29
Import/Flush all

SergeyYukhnevich:
Add ability to import/flush all models by one command.
I think that it is handy command and it is useful to import all models at once.
If you have any comments or suggestions I'll be glad to get them.


taylorotwell:
No plans to add right now.

Issue #34, opened 2016-08-22
"null" driver not settable via env variable


harrygr:
When trying to set the scout engine to the "null" implementation (e.g. say as a fallback for when no env variable exists: `'driver' => env('SCOUT_DRIVER', 'null')` or for tests) the string `"null"` is parsed into a php `null` value instead resulting in an instantiation error.

Suggest either renaming the null driver so something else like nullsearcher unless there is a way in phpunit.xml and .env it can be set as a string.

morloderex:
Use double qoutes on the enviroment variable.

harrygr:
How would one do that in a phpunit.xml file where the values are already between quotes?

    <env name="SCOUT_DRIVER" value="null" />

dwightwatson:
Using double quotes on the environment variable doesn't work. I've been trying to do this to disable Laravel Scout in my test environment. Dotenv must take the string "null" to become null as well.

```
SCOUT_DRIVER="null"
```

```
>>> env('SCOUT_DRIVER');
=> null
```

My solution in the meantime has been to set the default driver to "null" in the `config/scout.php` file and then set it to "algolia" through an environment variable in production, but clearly this isn't great.

I suspect this issue will also affect the "null" Broadcast driver.

talvbansal:
@dwightwatson I've managed to set my driver to null using your method as well as the `<env name="SCOUT_DRIVER" value='"null"' />` method too.

However my unit tests are **DIRE** slow after this:

**25seconds -> 7mins** 

Are slow unit tests something you're seeing too? 

It may well be because I'm using the [TntSearch driver](https://github.com/teamtnt/laravel-scout-tntsearch-driver) however I wanted to see if this is a scout issue or their implementation issue. Although using their native implementation not the scout wrapper has no impact on testing speed.

harrygr:
@dwightwatson that's an OK suggestion but there's a problem when, for development, you have it set to "algolia" in your .env phpunit will fall back on that value if no value is set in phpunit.xml. Basically it means that if you have your local driver set to algolia there's no easy way to override that for tests. Will be a fine solution for CI though where you can just omit the driver env variable completely.

thebrubaker:
I agree,it would be good to change this "null" option to "none" or "disable" or some equivalent that won't confuse .env to typecasting null from a string to a null value.

talvbansal:
[...]

PR #224, opened 2017-09-13
Use string "null" if default driver is set to NULL

inxilpro:
If your `SCOUT_DRIVER` environmental variable is set to `"null"`, the `env()` helper will convert that to the value `NULL`. The current workaround is to set the value to `'"null"'`, which forces `env()` to load it as a string, or to add a check to your `scout.php` config file.

This updates `getDefaultDriver()` to return the string `"null"` if the value of `config('scout.driver')` returns `NULL`.

### DIFF
--- a/src/EngineManager.php
+++ b/src/EngineManager.php
@@ -52,6 +52,6 @@ class EngineManager extends Manager
      */
     public function getDefaultDriver()
     {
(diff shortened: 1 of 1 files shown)


julienbourdeau:
Relates to this _popular_ issue: https://github.com/laravel/scout/issues/34

PR #225, opened 2017-09-14
Fix PHP 5.6 compatibility for null driver

julienbourdeau:
Fix for NullDriver introduced in https://github.com/laravel/scout/pull/224 is PHP7 only.

Thanks to @royscheepens for [noticing](https://github.com/laravel/scout/commit/84762c8ed51cb57f09b5f465e09993e48baf9d55#commitcomment-24295476).


royscheepens:
Awesome!

PR #234, opened 2017-11-02
[3.0] Update dev dependencies

gabriel-caruso:
Updated `mockery/mockery` to `~1.0` and `phpunit/phpunit` to `~6.0`.

PS: I've used [this commit](https://github.com/laravel/framework/pull/17864) from `laravel/framework` to prevent failed tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,8 +28,8 @@
     },
     "require-dev": {
         "algolia/algoliasearch-client-php": "^1.10",
-        "mockery/mockery": "~0.9",
-        "phpunit/phpunit": "~5.0"
+        "mockery/mockery": "~1.0",
+        "phpunit/phpunit": "~6.0"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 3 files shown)


taylorotwell:
Tests fail.

gabriel-caruso:
@taylorotwell The tests only failed because of PHPUnit 6.0 [requires PHP 7.0](https://github.com/sebastianbergmann/phpunit/blob/master/composer.json#L25). Should we drop support to `5.6`?

browner12:
as long as we are supporting Laravel ~5.4, we need to continue supporting PHP 5.6.  If PHPUnit requires 7.0, I think you're SOL on this one.

gabriel-caruso:
@browner12 So should I tagged it as `Scout 4.0`?

browner12:
you could try, but I highly doubt they're go to bump a full Major version just for some updated dev dependencies, unless you have a legitimate reason that they **need** to be updated.

gabriel-caruso:
@browner12 The only reason is to keep everything update, including test, as `laravel/framework` did. I’ll try, thanks 😊

PR #235, opened 2017-11-02
[4.0] Updated to PHPUnit 6 and Mockery 1

gabriel-caruso:
Updated `phpunit/phpunit` to `~6.0` and `mockery/mockery` to `~1.0`.

Dropped support to `PHP 5.6`, as PHPUnit 6.0 [requires PHP 7.0](https://github.com/sebastianbergmann/phpunit/blob/master/composer.json#L25).

PS: I've used [this commit](https://github.com/laravel/framework/pull/17864) from `laravel/framework` to prevent failed tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.6
   - 7.0
   - 7.1
 
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         }
     ],
     "require": {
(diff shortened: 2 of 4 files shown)


browner12:
I would update the illuminate dependencies as well to 5.5.

but also, IMHO, don't think justifies a new major version.

gabriel-caruso:
@browner12 If @taylorotwell merge it, I'll update the `illuminate` dependencies as well.

PR #237, opened 2017-11-29
[POC] Automatically set engine manager version

tillkruss:
_This is just a POC, please don't merge._

It's kinda annoying that the Algolia user agent version in `EngineManager` is always out of date. I'd like to suggest setting the Scout version during the package install/update.

Thoughts?


src/Console/InstallCommand.php
```diff
@@ -0,0 +1,48 @@
...
+        $this->info('Setting Laravel Scout version...');
+
+        $version = collect(
+            json_decode(file_get_contents(realpath(__DIR__ . '/../../../../composer/installed.json')))
+        )->where('name', 'laravel/scout')->first()->version;
+
+        $manager = realpath(__DIR__ . '/../EngineManager.php');
+
+        $content = str_replace(
+            'const VERSION = null;',
```
julienbourdeau: I guess this will only work on the first install. You might need to use a regex instead.

src/Console/InstallCommand.php
```diff
@@ -0,0 +1,48 @@
...
+     * Execute the console command.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $this->info('Setting Laravel Scout version...');
+
+        $version = collect(
+            json_decode(file_get_contents(realpath(__DIR__ . '/../../../../composer/installed.json')))
```
julienbourdeau: I think you could use `base_path` helper for the path. It would be more readable but it might also do somehting for OS compatibility (not sure).

```php
realpath(__DIR__ . '/../../../../composer/installed.json')
```
would become
```php
base_path('vendor/composer/installed.json')
```

julienbourdeau:
Reading the composer doc, it seems like you can only invoke a static method: https://getcomposer.org/doc/articles/scripts.md#defining-scripts
I'm not sure how we could get this to work easily :/

composer.json
```diff
@@ -51,6 +51,10 @@
             ]
         }
     },
+    "scripts": {
+        "post-update-cmd": "Laravel\\Scout\\Console\\InstallCommand::handle",
+        "post-install-cmd": "Laravel\\Scout\\Console\\InstallCommand::handle",
```
julienbourdeau: This trailing comma should be removed

tillkruss:
@julienbourdeau: Thanks, great feedback. I just copied what Laravel does.

taylorotwell:
Will this create problems when installing a new version on a production server where Composer prompts you that you have made changes to vendor files and asks which version you would like to keep?

tillkruss:
I'm not entirely sure. I assume that most people use something like:

```
composer install --no-dev --prefer-dist --optimize-autoloader --no-interaction
```

Where `--no-interaction` would prevent that, but to be honestly, we either have to ask a Composer wizard or just test this.

taylorotwell:
Think I will hold off on this for now then. Worried about issues with Composer updates.

PR #240, opened 2017-12-18
Conditionally make a model instance searchable

kfirba:
Hey.

I've come across this issue where I wanted to conditionally make some models searchable. The example I had is with `Post`s. The posts in my system can be in "draft" or published state. If a post is in the "draft" state, I would not like to make it searchable.

The current state of Scout does not let us disable model syncing per instance but per class only.

To "disable" syncing for posts in the "draft" state, I've ensured that the `searchableArray()` method returns an empty array. While it does indeed ensure that the "draft" posts won't be synced, it sends a request to make the model searchable with empty data which is useless and wasting indexing operation if you're using a driver such as Algolia.

This PR adds a simple "hook" that each model can implement to determine if it should be searchable:

```php
class Post extends Model {
    use Searchable;
    
    public function shouldBeSearchable()
    {
        return $this->isPublished();
    }
}
```

By default, the `shouldBeSearchable()` method is implemented on the `Searchable` trait and returns `true` so this change does not cause any breaking changes.

### DIFF
--- a/src/ModelObserver.php
+++ b/src/ModelObserver.php
@@ -59,6 +59,10 @@ class ModelObserver
         if (static::syncingDisabledFor($model)) {
             return;
         }
+        
+        if (! $model->shouldBeSearchable()) {
+            return;
+        }
 
         $model->searchable();
     }
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -197,6 +197,16 @@ trait Searchable
     }
 
     /**
+     * Determine if the model should be searchable.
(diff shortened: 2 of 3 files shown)


kfirba:
@taylorotwell thanks for merging this. Any plans to create a new tag? Maybe 3.1?

MrHexagon:
Looking forward to shouldBeSearchable :)

rikiless:
Waiting for this too, thanks!

Issue #251, opened 2018-01-13
Indexing isn't fired on saved event


adiachenko:
ModelObserver listens to `created` and `updated` events. However, those are not fired when touching parent timestamps (see [HasRelationships::touchOwners](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php#L599)) which is a bummer if you index attributes that come from relationships. 

Monkey patching `saved` method into ModelObserver seems to work, but I wonder whether the current behavior is there by design. If so, what is the best way to address the described use case?

adiachenko:
I am now closing this as related PR has been merged into 4.0 branch.

PR #252, opened 2018-01-13
Listen to saved event instead of created/updated

adiachenko:
Related to #251.

Listening to `created` and `updated` events instead of `saved` presents a problem when index includes attributes that come from relationships (ie: product index includes dealer name from another model, post has images that are stored separately etc.), because those events aren't triggered when touching parent timestamps.


taylorotwell:
This seems generally OK.

PR #256, opened 2018-01-21
Keep Scout Builder uncoupled from the config and the framework helpers

pmatseykanets:
This PR attempts to uncouple Scout's `Builder` from the config and the framework helpers (`config()` specifically in this case) that was most likely inadvertently introduced by #236.

It also adds the `soft_delete` option to the `config/scout.php` file so that users know it's available.

I don't necessarily insist in this particular implementation but strongly believe that keeping `Builder` uncoupled is a good thing.

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -50,7 +50,7 @@ return [
     |
     | These options allow you to control the maximum chunk size when you are
     | mass importing data into the search engine. This allows you to fine
-    | tune these chunk sizes based on the capabilites of your machines.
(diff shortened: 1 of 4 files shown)


taylorotwell:
Doesn't follow framework conventions for variable casing.

pmatseykanets:
@taylorotwell Would you mind reopening the PR so I can fix the variable casing if it's the only issue? Or you'd like a separate PR? Thanks.

PR #257, opened 2018-01-22
Uncouple builder

pmatseykanets:
Same as #256 but variable casing is fixed.

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -50,7 +50,7 @@ return [
     |
     | These options allow you to control the maximum chunk size when you are
     | mass importing data into the search engine. This allows you to fine
-    | tune these chunk sizes based on the capabilites of your machines.
+    | tune these chunk sizes based on the capabilities of your machines.
     |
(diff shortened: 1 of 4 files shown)


taylorotwell:
Why do you want to uncouple it?

pmatseykanets:
I guess the best answer to the "Why" question would be your own code for the majority of the framework. May be I missing something but I could not find a single instance in whole framework where a class would reach out to the config file in its constructor. That's what service providers do. More over scout's `Builder` is a kind of DTO object that is being passed from a userland's Model to a Engine. Why should it know anything about where the configuration lives instead of getting it via constructor or in a worse case scenario some kind of method?

I most likely would've missed it but after applying #236 it broke the tests for my PostgreSQL engine complaining about the unknown `config()` function.

On a side note it'd would be great if we could resolve the scout's `Builder` from the container rather than just `new`ing it up.

In any case thanks for your time and consideration.

PR #263, opened 2018-02-10
Syncing for conditional models

arjanwestdorp:
I've come across this issue when I was using conditional searchable models. We have a `Post` model and only want to make it searchable when `$post->is_active === true`. I'm using the `shouldBeSearchable()` from  #240. I figured out this works fine when making the `Post` active, but not when deactivating the `Post` afterwards. When deactivating the `Post` you would expect it to be deleted from the search engine.

This PR implements syncing for those use cases where a searchable model turns into an unsearchable model from a `saved` event. In this case it will check if the syncing is enabled and call the `unsearchable` method.

I've made this configurable through the config file to be able to control this behaviour. I've set the default to `false` to not cause any breaking changes.

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -74,6 +74,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Sync unsearchables
+    |--------------------------------------------------------------------------
+    |
+    | This option allows you to control whether to make the model unsearchable
+    | in case a model should not be searchable anymore. This can be helpful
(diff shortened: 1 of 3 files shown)


taylorotwell:
IMO this should just be the default behavior not a configuration option. Feel free to make another PR.

arjanwestdorp:
I agree, that's what I thought as well. I'll create another PR which makes this the default behaviour.

Issue #267, opened 2018-03-01
Overwriting Delete logic


aconital:
My Algolia ObjectId is modified and not equal to primary key of my model. As a result when I call delete on my object Scout doesn't remove it from algolia.  Is there a way to overwrite the delete logic, something similar to toSearchableArray.

julienbourdeau:
Can you share what your ObjectId look like and why you modified it?

aconital:
@julienbourdeau My Laravel model's primary key is id which is an integer but because I needed my frontend to be able to filter on a unique field, I had to change the Algolia ObjectID to that unique field.
Imagine this example:
```
User{
id,
name
email
}

```
```
Algolia User 
{
objectID: email
name
}
```
I didn't want to mark that field as a facet because it has a unique value so I had to change the objectID because Algolia allows you to filter on ObjectID.

PR #268, opened 2018-03-10
Use custom Key to index a model

OliverHolz:
I had a problem similar to #44 and #267. My data is indexed with a custom ObjectID which is equal to the Eloquent ID.

I just added a new method "getScoutKey" to the SearchableTrait which returns the Eloquent key by default but can be overwritten.

This should fix both issues mentioned above.

### DIFF
--- a/src/Console/FlushCommand.php
+++ b/src/Console/FlushCommand.php
@@ -35,7 +35,7 @@ class FlushCommand extends Command
         $model = new $class;
 
         $events->listen(ModelsFlushed::class, function ($event) use ($class) {
-            $key = $event->models->last()->getKey();
+            $key = $event->models->last()->getScoutKey();
 
             $this->line('<comment>Flushed ['.$class.'] models up to ID:</comment> '.$key);
         });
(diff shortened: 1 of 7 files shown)


bb128b:
This is broken somehow. I set the value of getScoutKey to a unique index on my Eloquent Model and the results come back as 0 even though is results.

```
    /**
     * Get the value used to index the model.
     *
     * @return mixed
     */
    public function getScoutKey()
    {
        return $this->name; // This is the unique index (objectID in Algolia).
    }
```

I have 6 results but the data is an empty array.

```
{
    "current_page": 1,
    "data": [],
    "first_page_url": "http://url.test/api/search?query=qui&page=1",
    "from": null,
    "last_page": 1,
    "last_page_url": "http://url.test/api/search?query=qui&page=1",
    "next_page_url": null,
    "path": "http://url.test/api/search",
    "per_page": 15,
    "prev_page_url": null,
    "to": null,
    "total": 6
}
```

I believe it is around this code that needs refactoring.

src/Engines/AlgoliaEngine.php - Line 175

```
        $models = $builder->whereIn(
            $model->getQualifiedKeyName(),
            collect($results['hits'])->pluck('objectID')->values()->all()
        )->get()->keyBy($model->getKeyName());
```

I would do it myself but I do not know Scout well enough to make a fix. If nobody attempts it I will try on Sunday an issue a pull request.

Issue #274, opened 2018-04-05
getScoutKey() broken get() and paginate() functions


bb128b:
This issue references the merge request from #268.

This is broken somehow. I set the value of getScoutKey to a unique index on my Eloquent Model and the results come back as 0 even though is results.

```
    /**
     * Get the value used to index the model.
     *
     * @return mixed
     */
    public function getScoutKey()
    {
        return $this->name; // This is the unique index (objectID in Algolia).
    }
```

I have 6 results but the data is an empty array.

```
{
    "current_page": 1,
    "data": [],
    "first_page_url": "http://url.test/api/search?query=qui&page=1",
    "from": null,
    "last_page": 1,
    "last_page_url": "http://url.test/api/search?query=qui&page=1",
    "next_page_url": null,
    "path": "http://url.test/api/search",
    "per_page": 15,
    "prev_page_url": null,
    "to": null,
    "total": 6
}
```

I believe it is around this code that needs refactoring.

src/Engines/AlgoliaEngine.php - Line 175

```
        $models = $builder->whereIn(
            $model->getQualifiedKeyName(),
            collect($results['hits'])->pluck('objectID')->values()->all()
        )->get()->keyBy($model->getKeyName());
```

I would do it myself but I do not know Scout well enough to make a fix. If nobody attempts it I will try on Sunday an issue a pull request.

julienbourdeau:
How do you search in the index?

driesvints:
Should be fixed in above PR ^

PR #281, opened 2018-05-24
Add getScoutKeyName() and mapScoutSearchResults() functions.

jasonlfunk:
This PR fixes issue #274.

The issue is that the function `getScoutKey()` can be used to make a custom index for the model. However, the Algolia engine had no idea how to turn this custom index back into the Eloquent models. This PR provides two ways to do this, a simple, and a complex one.

The simple method is `getScoutKeyName()` which you can use to provide the Algolia engine with the name of the key to use to hydrate the Eloquent models.

```php
    public function getScoutKey()
    {
        return $this->email;
    }

   {
       return 'email';
   }
```
However, the `getScoutKey()` method does not require that a single model attribute be returned. It could be a compound key.

```php
    public function getScoutKey()
    {
        return $this->email . '-' . $this->name;
    }
```

In this case, we cannot use the `getScoutKeyName()` function. We can use the `mapScoutSearchResults()` method.

```php
public function mapScoutSearchResults(array $objectIds)
{
    $query = $this->newQuery();
    foreach($objectIds as $objectId) {
        list($email, $name) = explode('-', $objectId);
        $query->orWhere(function($query) use($email, $name) {
            $query->where('email', $email);
            $query->where('name', $name);
        });
    }
    return $query->get();
}
```

### DIFF
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -176,11 +176,11 @@ class AlgoliaEngine extends Engine
         )->keyBy($modelKey);
 
         return Collection::make($results['hits'])->map(function ($hit) use ($models, $modelKey) {
-            $key = $hit[$modelKey];
+            $key = $hit['objectID'];
 
-            if (isset($models[$key])) {
-                return $models[$key];
-            }
+            return $models->first(function ($model) use ($key) {
+                return $model->getScoutKey() === $key;
+            });
         })->filter()->values();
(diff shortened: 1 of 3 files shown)


jasonlfunk:
👍 Yes, the changes work well on my application. I didn't know that `keyBy` could accept a function as well.

Looks like you are just as good of a programmer without a beard.

taylorotwell:
Nice!

PR #286, opened 2018-06-06
Only update dirty models

### LINKED ISSUE #285: Models using `Searchable` are pushed even with no changes
Seems to be caused by #252. The `saved` event is triggered even when no changes have been made to the model. There, `$model->searchable()` is called unconditionally, regardless of whether the model is actually dirty or not. This causes any model to be imported to the search index whenever it is saved – with and without changes.

Compared to that, the `updated` event – which was used in the previous verison – is actually only triggered, when the model is dirty, see https://github.com/laravel/framework/blob/17c4c0faaecb5293febada7930a459a9efd6979b/src/Illuminate/Database/Eloquent/Model.php#L626-L637.

Observed in a Laravel project of mine, I can create a reproduction repository, if wanted.

msiemens:
Fixes #285

### DIFF
--- a/src/ModelObserver.php
+++ b/src/ModelObserver.php
@@ -66,7 +66,9 @@ class ModelObserver
             return;
         }
 
-        $model->searchable();
+        if ($model->isDirty()) {
+            $model->searchable();
+        }
     }
 
     /**
--- a/tests/ModelObserverTest.php
+++ b/tests/ModelObserverTest.php
@@ -12,6 +12,7 @@ class ModelObserverTest extends AbstractTestCase
         $observer = new ModelObserver;
         $model = Mockery::mock();
         $model->shouldReceive('shouldBeSearchable')->andReturn(true);
+        $model->shouldReceive('isDirty')->andReturn(true);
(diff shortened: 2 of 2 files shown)


taylorotwell:
No plans to change this on a point release.

msiemens:
Why?

julienbourdeau:
Beacuse `isDirty` compares the current model to what is saved in the DB, not in the search index, I'd not recommend this.

If you do any change with sync disabled like so, the next `update()` won't update the search index if the model isn't considered dirty.

```php
App\Order::withoutSyncingToSearch(function () {
    // Perform model actions...
});
```

It's a corner case, but I think the dirty check should be checking if it's dirty compared to the search index (which is _impossible_).

msiemens:
Well, as far as I can tell, this behavior was fine in version 3.0. Is there any documentation on significant changes like this one?
I understand that dirty checking requires that the Laravel application alone manages search index entries. So it's gonna be inconvenient to one group of users or another, no matter what choice is made. It would just be nice to have some sort of documentation on changes like this (if I'm not missing anything).

Issue #284, opened 2018-06-02
Add support for query Scopes


ARKKYN:
$this->customers->search($input['query'])->AmountReceivables()->withCount(['vehicles','totalInvoices','invoices'])->paginate($limit);

throws error On using  scopes

driesvints:
Should be fixed by now on 5.0, let me know if you still having troubles with this.

ARKKYN:
@driesvints  thank you

mrmonat:
This doesn't seem to work anymore as ```Model::search($query)``` returns an Instance of `\Laravel\Scout\Builder` instead of `\Illuminate\Database\Eloquent\Builder` ?

sebastiaanluca:
@driesvints Can't get it to work as in the example above, but is the following the preferred way of adding query parameters that will be applied to the search results?

```php
return Category::search($query)
    ->query(function (Builder $builder) {
        $builder->enabled();
    })
    ->take(5)
    ->get();
```

driesvints:
@sebastiaanluca hmm, can't find an example in the docs. I think so yeah but unsure. Where does the `enabled` function come from? Are you using the latest 7.x release?

sebastiaanluca:
`enabled()` is just the model scope. Can be an eager load, scope, where, etc, but still have to write tests to assert this works. Using `laravel/scout` v7.1.2 in combination with `teamtnt/laravel-scout-tntsearch-driver` v7.1.0.

Applying a scope, limit, etc after the `search()` call throws an exception it can't find the method on the builder. Adding it before complains about `search()` not being a method on the class.

driesvints:
@sebastiaanluca we don't `teamtnt/laravel-scout-tntsearch-driver`. Maybe this is an issue with this driver? Does it works with the Algolia driver?

sebastiaanluca:
Since it acts on Scout interfaces and classes I suspect it's unrelated to the used driver, but don't have a setup to test with the Algolia driver so can't compare unfortunately.

Tbh, I think it's just some incomplete or faulty documentation. Checking the Scout builder class, there's no magic method forwarding to the underlying query builder class, only a `query` method that `Set the callback that should have an opportunity to modify the database query.` You shouldn't be able to directly call scopes and relations after calling `search`.

MartinMorlog:
> @driesvints Can't get it to work as in the example above, but is the following the preferred way of adding query parameters that will be applied to the search results?
> 
> ```
> return Category::search($query)
[...]

Issue #230, opened 2017-10-05
Support relations withCount() method


yingdongzhang:
`withCount()` doesn't work with the `::search()` method, as the `search()` method doesn't return a QueryBuilder but a \Laravel\Scout\Builder instance, and it can't be chained to a QueryBuilder either. Is this expected to be supported soon?

For instance if I have Question & Answer, one Question can have more than one Answers.

Currently both of these two queries will throw a method not defined exception, depends on the order you call them:

```
$questions = Question::search('hello')->withCount('answers')->get();
$questions = Question::withCount('answers')->search('hello')->get();
```

julienbourdeau:
If you set the `withCount` property in your model, it will be taken into account.

Otherwise, if it has to be optional because you don't want to load the count all the time, you can leverage the macros.

```php
    Laravel\Scout\Builder::macro('withCount', function () {
        $this->model->withCount(['people']);

        return $this;
    });
```

You also have to implement the `withCount` method in the model:
```php
    public function withCount(array $withCount)
    {
        $this->withCount = $withCount;

        return $this;
    }
```

The you will be able to use it this way:
```php
Question::search('hello')->withCount(['answers'])->get();
```

driesvints:
https://github.com/laravel/scout/pull/290 Got merged so perhaps that solves your problem? 

Or you can try @julienbourdeau's suggestion.

Div-Man:
`Post::search('hello')->query(fn($query) => $query->withCount('answers'))->get());`

PR #288, opened 2018-06-29
allow database query customization in getScoutModelsByIds

### LINKED ISSUE #140: Custom query scope
When I use the laravel scout search() method on my model, my custom created scope query doesn't work. Is there a way how to use both?

Thanks

bastien-phi:
This pull request aims to customize query used to fetch models from database using object IDs.

It could resolve #140, #230, #284 

It can be used easily doing  : 
```php
$authors = Author::search('John')
                  ->mapUsing(function($query) {
                      // $query is Illuminate\Database\Eloquent\Builder
                      return $query->withCount('articles')->with('country');
                  })->get();
// at this point articles_count of all $authors is set with correct value
// and country relation is loaded
```
The callback is runned after engine's search so adding filter will work but can break pagination
```php
$authors = Author::search('John')
                  ->mapUsing(function($query) {
                      return $query->where('country_id', 1);
                  })->get();
```
Does it look good ?

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -162,6 +162,18 @@ class Builder
     }
 
     /**
+     * Add a callback to apply on query used to fetch models from database
+     *
+     * @param \Closure $callback
+     * @return $this
+     */
+    public function mapUsing($callback) {
+        $this->model->getScoutModelsUsing = $callback;
+
+        return $this;
+    }
+
+    /**
      * Get the raw results of the search.
      *
      * @return mixed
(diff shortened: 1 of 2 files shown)


taylorotwell:
This approach would literally break all queued actions in Laravel that use models because you can't serialize a Closure.

taylorotwell:
Will need to use some other approach.

bastien-phi:
I see. Did not think about model serialization.

I will try another approach. Does this feature look interesting ?

PR #305, opened 2018-09-12
add installation via composer

AhmedHelalAhmed:


browner12:
installation instruction are on the main Laravel documentation, as linked in the readme, so we don't need to have it here.

https://laravel.com/docs/master/scout#installation

Issue #101, opened 2016-10-12
scout:flush command doesn't flush/remove anything?


bbashy:
It seems `scout:flush` doesn't seem to do anything. Deleting a record from the database manually then doing `scout:flush` and/or `scout:import` doesn't remove the old record.

AlexCatch:
I can confirm this, doing flush with ElasticSearch doesn't work.

sebestenyb:
Neither with [TNTSearch](https://github.com/teamtnt/laravel-scout-tntsearch-driver)

bbashy:
Solution was to just do a cURL DELETE request to ES but who wants to do that when you have artisan commands.

JackWH:
I've been unable to get flushing to work with Algolia either.

LoganLehman:
@taylorotwell Any news?

aurawindsurfing:
Hi, I confirm that:
`php artisan scout:flush "App\Classified"
All [App\Classified] records have been flushed.`
does nothing to algolia index.

LoganLehman:
Ping again?

giacomodeglinnocenti:
Same issue here on fresh Laravel 5.4

eithed:
On my fresh install while flush removes entries that exist, it doesn't remove entries that do not

julienbourdeau:
I think the `scout:flush` command doest clear the index but instead, it removes items by ID.

Is it possible that the id you have in database are not the same as the one you have in your search index?

This happens when you seed your DB, import, trunctate your table and seed again. At this point, if you flush, it won't remove anything.

eithed:
Yup - that's exactly the case, as reported by OP; it "works", but you'd expect flush to empty all entries, as after truncating the DB it won't be possible to get rid of them without resorting to deleting the entire index.

IllyaMoskvin:
Seconding @julienbourdeau. Thanks for the suggestion to check ids, they were the cause of our issue as well! Our situation was a little different. Sharing the details for posterity, just in case someone is using a similar approach:

We are using a custom Solr Scout engine, and we were indexing all `Searchable` models into one core (storing a common set of fields). We couldn't use the models' primary keys to track them in the index, since that would cause collisions among different models, so we assigned them string keys namespaced with the model's classname. We forgot to update the `delete()` method in our `SolrScoutEngine` to account for this transformation. Everything fixed itself once we applied the same transformation there.

[...]

PR #310, opened 2018-10-05
[6.0] Use engine to flush records of model

driesvints:
Instead of iterating over ids from the database we use the engine itself to flush the records of an index/model.

This solves the problem where models would get out of sync with the search records and old records weren't flushed. With this implementation all records are always removed.

If there are engines which don't provide this functionality they can still iterate over the model ids if they want to.

Fixes https://github.com/laravel/scout/issues/101

@julienbourdeau would you mind also taking a look at this if this is correctly implemented? I didn't have the chance to test this with a live implementation of Algolia.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release Notes
 
-## [v6.0.0 (unreleased)](https://github.com/laravel/framework/compare/v5.0.3...v6.0.0)
+## [v6.0.0 (unreleased)](https://github.com/laravel/scout/compare/v5.0.3...v6.0.0)
 
 ### Changed
 - Flush records of a model using the engine ([#310](https://github.com/laravel/scout/pull/310))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,6 @@
+# Release Notes
+
+## [v6.0.0 (unreleased)](https://github.com/laravel/framework/compare/v5.0.3...v6.0.0)
+
+### Changed
+- Flush records of a model using the engine ([#310](https://github.com/laravel/scout/pull/310))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,4 +3,4 @@
 ## [v6.0.0 (unreleased)](https://github.com/laravel/scout/compare/v5.0.3...v6.0.0)
 
 ### Changed
-- Flush records of a model using the engine ([#310](https://github.com/laravel/scout/pull/310))
+- Flush records of a model using the engine. **This removes the emitting of the `ModelsFlushed` event.** ([#310](https://github.com/laravel/scout/pull/310))
--- a/src/Console/FlushCommand.php
+++ b/src/Console/FlushCommand.php
@@ -3,7 +3,6 @@
 namespace Laravel\Scout\Console;
 
 use Illuminate\Console\Command;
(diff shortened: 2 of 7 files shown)


src/Console/FlushCommand.php
```diff
@@ -34,12 +34,6 @@ public function handle(Dispatcher $events)
...
 
-        $events->listen(ModelsFlushed::class, function ($event) use ($class) {
-            $key = $event->models->last()->getScoutKey();
-
-            $this->line('<comment>Flushed ['.$class.'] models up to ID:</comment> '.$key);
-        });
-
         $model::removeAllFromSearch();
 
         $events->forget(ModelsFlushed::class);
```
nunomaduro: You also don't need this line.

src/Engines/AlgoliaEngine.php
```diff
@@ -194,6 +194,19 @@ public function getTotalCount($results)
...
+     * Flush all of the model's records from the engine.
+     *
+     * @param  \Illuminate\Database\Eloquent\Model  $model
+     * @return void
+     */
+    public function flush($model)
+    {
+        $index = $this->algolia->initIndex($model->searchableAs());
+
+        $index->clearIndex();
```
nunomaduro: I confirm. This is good. 👍

CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Release Notes
+
+## [v6.0.0 (unreleased)](https://github.com/laravel/framework/compare/v5.0.3...v6.0.0)
```
nunomaduro: This should be: 	## [v6.0.0 (unreleased)](https://github.com/laravel/scout/compare/v5.0.3...v6.0.0)

driesvints:
@nunomaduro thanks Nuno!

julienbourdeau:
All good 💯 

Another great new with this implementation is that it will save a lot of indexing operations for the user. This command will only **be counted (or charged) as 1 operation** instead of as many as table rows.

taylorotwell:
So this removes the ModelsFlushed event?

nunomaduro:
@taylorotwell No. With @driesvints's proposal, the method `$model::removeAllFromSearch()` no longer calls the method`$model::unsearchable()`, so the event `ModelsFlushed::class` is not fired.

But, if the developer intencionally calls `$modelsCollection::unsearchable()`, the event `ModelsFlushed::class` is fired on every batch of 500 models. This way the developer have access to the current state of the batch operation: Models already flushed, last model flushed, etc.

src/Console/FlushCommand.php
```diff
@@ -34,16 +34,8 @@ public function handle(Dispatcher $events)
```
nunomaduro: @driesvints This class no longer needs the line `use Laravel\Scout\Events\ModelsFlushed;`.

driesvints: Done!

driesvints:
@taylorotwell @nunomaduro Taylor does bring up a good point that the `ModelsFlushed` event isn't fired anymore because it accepts an array of which models were flushed. With this new implementation we just assume *all models* were flushed. But some people may rely on the event to do stuff after that. So perhaps we'll need to introduce a new event or adjust the current event to work without having to pass a collection of models.

Thoughts?

PR #311, opened 2018-10-05
[6.0] Allow people to modify query parameter name

driesvints:
With this change you can customize the query parameter name when paginating results.

Sent to 6.0 because this modifies the method signature and is a breaking change.

Fixes https://github.com/laravel/scout/issues/58

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -220,9 +220,10 @@ class Builder
      * @param  int  $perPage
      * @param  string  $pageName
      * @param  int|null  $page
+     * @param  string  $queryName
      * @return \Illuminate\Contracts\Pagination\LengthAwarePaginator
      */
-    public function paginate($perPage = null, $pageName = 'page', $page = null)
+    public function paginate($perPage = null, $pageName = 'page', $page = null, $queryName = 'query')
     {
         $engine = $this->engine();
 
@@ -239,7 +240,7 @@ class Builder
             'pageName' => $pageName,
         ]));
 
-        return $paginator->appends('query', $this->query);
+        return $paginator->appends($queryName, $this->query);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Builder.php
```diff
@@ -220,9 +220,10 @@ public function get()
      * @param  int  $perPage
      * @param  string  $pageName
      * @param  int|null  $page
+     * @param  string  $queryName
      * @return \Illuminate\Contracts\Pagination\LengthAwarePaginator
      */
-    public function paginate($perPage = null, $pageName = 'page', $page = null)
+    public function paginate($perPage = null, $pageName = 'page', $page = null, $queryName = 'query')
```
nunomaduro: On Eloquent builder `User::paginate(10)` there is a query parameter by default? Otherwise I would say to don't append nothing on the scout builder and let the user decide, e.g: `User::search($query)->paginate()->appends('q', $query);`.

driesvints: Good remark! Didn't think of that.

x1c0: why was this closed? This is an issue integrating with elastic search.. it should be possible to specify the query label

driesvints: @x1c0 Elastic search isn't supported anymore

x1c0: @driesvints I don't understand.. what that means? Laravel does not support integration with elastic search?

driesvints: Scout now only supports Algolia.

johnnyhuy:
Hey, is this change getting applied to 6 or 7? Seems like it's still hard-coded as mentioned in #58

driesvints:
@johnnyhuy please see https://github.com/laravel/scout/pull/311#discussion_r223059364

potatoru:
The `query` parameter is still annoying. It would be nice if I could use my own parameter name (`search` for example) without duplicate it with `query` (`?search=bla&query=bla`)

mhrn77:
> The `query` parameter is still annoying. It would be nice if I could use my own parameter name (`search` for example) without duplicate it with `query` (`?search=bla&query=bla`)

This is true. This is not happening with the default Laravel paginator: `?search=pencil&page=2` , but only with Laravel Scout: `?query=pencil&search=pencil&page=2`. Hope this gets fixed.

PR #314, opened 2018-10-08
[6.0] Add method count to builder

### LINKED ISSUE #164: Support count() method

nunomaduro:
This Pull Request adds the method `count` to the Builder class and it improves the developer experience while interacting with the Builder class.

Before:
```php
$rawResults = Thread::search('zonda')->raw();
$count = (new Thread)->searchableUsing()->getTotalCount($rawResults);
```

Now:
```php
$count = Thread::search('zonda')->count();
```

Fixes #164 and the original idea comes from @julienbourdeau's PR: https://github.com/laravel/scout/pull/183.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,5 +2,8 @@
 
 ## [v6.0.0 (unreleased)](https://github.com/laravel/scout/compare/v5.0.3...v6.0.0)
 
+### Added
+- Method `count` to Builder ([#314](https://github.com/laravel/scout/pull/314))
+
 ### Changed
 - Flush records of a model using the engine. **This removes the emitting of the `ModelsFlushed` event.** ([#310](https://github.com/laravel/scout/pull/310))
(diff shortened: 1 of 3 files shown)


taylorotwell:
This may not make sense for all drivers? This implementation forces a search which wouldn't make sense in some cases (MySQL driver).

driesvints:
Yeah I agree, perhaps it's best to have a count method on the builder?

julienbourdeau:
Then my first attempt would work but it adds a `count` method to the abstract engine: https://github.com/laravel/scout/pull/183/files
Since it's a major version, I guess it's ok.

driesvints:
Yeah, if we can add this on a new major version (6.0) I see no problem with it.

taylorotwell:
Yeah, I think that approach would make more sense across drivers.

Issue #316, opened 2018-11-08
Array to String conversion with Nova


m2de:
In my current implementation I get a random Array to String conversion error when performing a search. The error is thrown here https://github.com/laravel/scout/blob/6.0/src/Engines/AlgoliaEngine.php#L144 with a key of `must` and the value is an empty array.

PHP 7.211
Laravel 5.7.12
Nova 1.1.7

driesvints:
Can you provide a sample request which is being sent to the app? And some relevant code like the model the search is performed on, any controller methods, etc?

m2de:
Hi @driesvints. Thanks for your response. In an effort to do this I managed to determine exactly what was causing it and resolved the issue. I was defining a different model in the `public static $model` to what I was returning in the `public static function newModel()` method (left over from some experimenting). Apologies. Thanks for helping resolve the issue.

m2de:
Ahhhh, correction ... let me start again (I had my filter fix in place) .... I will try and provide some more info shortly ...

m2de:
Model: https://github.com/dewsign/nova-pages/blob/master/src-php/Models/Page.php

Nova Resource: https://github.com/dewsign/nova-pages/blob/master/src-php/Nova/Page.php

There is some code in there from our Elasticsearch implementation (which works fine btw), but we are trying to switch to Algolia which then returned this error.

driesvints:
Can you post the code where the searching happens, the route to it and the request that's being made that causes the error?

m2de:
Thanks @driesvints ... I simply attempt to search within Laravel Nova, it is not code I have written.

![screen recording 2018-11-08 at 11 13 37](https://user-images.githubusercontent.com/17720020/48195545-d6c1ac00-e347-11e8-95ea-06348da77f4e.gif)

Error log:

```
[2018-11-08 11:13:40] local.ERROR: Array to string conversion {"userId":1,"email":"example@email.com","exception":"[object] (ErrorException(code: 0): Array to string conversion at /Users/marcomark/Valet/leightons-maxfactor-laravel/vendor/laravel/scout/src/Engines/AlgoliaEngine.php:144)
[stacktrace]
#0 /Users/marcomark/Valet/leightons-maxfactor-laravel/vendor/laravel/scout/src/Engines/AlgoliaEngine.php(144): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(8, 'Array to string...', '/Users/marcomar...', 144, Array)
[...]

PR #317, opened 2018-11-08
[6.0] Adding filter to remove unwanted empty arrays and avoid exceptions

m2de:
This fixes issue #316 by removing empty values/arrays from the filters and the search performs as expected.

### DIFF
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -140,7 +140,7 @@ class AlgoliaEngine extends Engine
      */
     protected function filters(Builder $builder)
     {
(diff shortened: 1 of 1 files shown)


nunomaduro:
@m2de What happens if a filter contains the value `0` or `false`?

m2de:
Hey @nunomaduro . Fair point. Could I filter to remove any array values if only string values are supported? Happy to amend code if that would be an acceptable solution. I'm unable to resolve the issue outside of this package.

m2de:
Thanks for your help all. Was my bad after all!

taylorotwell:
Sounds like something that may need tests.

PR #321, opened 2018-11-18
[6.0] Don't add soft delete query filter on hard delete models

jhoff:
Currently, when the static `search` method from the `Searchable` trait is called, a new scout query builder is instantiated and the `$softDelete` flag is based only on `config('scout.soft_delete')`.

The builder should only apply soft deletes if the model that is being searched also uses them.

It seems that Algolia ignores the additional `__soft_delete` filter when it it passed in a search that doesn't use soft deletes, but other drivers have to go out of the way to fix this deficiency ( See [this section in the tnt search driver](https://github.com/teamtnt/laravel-scout-tntsearch-driver/blob/master/src/Engines/TNTSearchEngine.php#L322) as an example )

This pull request fixes the searchable trait to only apply the soft delete filter when the model being searched actually uses soft deletes.


jhoff:
@taylorotwell thanks for the merge. Is there a process to tagging / releasing new code changes?

driesvints:
@jhoff tagged 6.1.0

jhoff:
@driesvints  Thank you sir!

06chaynes:
I had already submitted a new issue ( #338 ) on this before I saw this thread, but I am having this exact issue on v6.1.3

The trait is being applied still regardless of the config setting.

PR #322, opened 2018-11-20
[6.0] Allow to change the builder implementation using the container

nunomaduro:
This Pull Request adds the possibility of changing the concrete implementation of the `Builder::class`. Behind the scenes, the developer can now specify is own implementation doing:
```php
        $this->app->bind(
            \Laravel\Scout\Builder::class,
            \My\Custom\Builder::class
        );
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release Notes
 
+## [v6.2.0 (unreleased)](https://github.com/laravel/scout/compare/v6.1.0...v6.2.0)
+### Added
+- Builder implementation can be changed using the container ([#322](https://github.com/laravel/scout/pull/322))
+
 ## [v6.1.0](https://github.com/laravel/scout/compare/v6.0.0...v6.1.0)
 
 ### Fixed
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -103,9 +103,12 @@ trait Searchable
      */
     public static function search($query = '', $callback = null)
     {
-        return new Builder(
-            new static, $query, $callback, static::usesSoftDelete() && config('scout.soft_delete', false)
-        );
+        return app(Builder::class, [
+            'model' => new static,
+            'query' => $query,
+            'callback' => $callback,
+            'softDelete'=> static::usesSoftDelete() && config('scout.soft_delete', false)
+        ]);
     }
 
     /**


driesvints:
@nunomaduro any reason why that would be a bad thing? Since you're providing a solution for a single provider it's not the most crazy idea to let people use your own trait, no? You can re-use everything from the current trait easily.

nunomaduro:
Currently, Scout Extended **only** requires a `composer require algolia/scout-extended` to work on top of Laravel Scout. I would love to keep it like that.

<p align="center">

</p>

driesvints:
But you're requiring scout already in scout extended so scout already comes with it? It doesn't really matter if you have to require the searchable trait then from scout or scout extended. The amount of steps you'll have to do are the same.

nunomaduro:
If you **already** have scout, the steps are not the same. And I also would like to redirect to the `laravel.com/scout` for the full installation process.

driesvints:
If you already have scout I'd just remove the scout dependency and require scout extended instead. I wouldn't redirect to laravel.com/docs/scout for the installation process either. I believe the installation steps at https://github.com/algolia/scout-extended#%EF%B8%8F-installation are fine atm. You can of course direct people to laravel.com/docs/scout for usage of scout.

nunomaduro:
@driesvints I do understand that we are talking about a minor thing here, and thanks for taking your time to give me your input on this.

For me, it's important that people fell that Scout Extended **just works** on top of Laravel Scout. And even small details, such as the installation process, are important.

Having developers to import our `Searchable` class instead of the scout one, feels hacky to me.

driesvints:
Ah yeah it's a minor thing, true. Just wanted to give my 2 cents :)

The current PR feels "hacky" for me 😅 I say this because it could be confusing when people are using scout that a different package binds its own Builder class which may not seem obvious at first. If you explicitly implement your own trait that it's very clear what's happening and that the search method is overwritten to provide its own Builder class.

But anyway, just wanted to voice my opinion on it. Let's see what @taylorotwell thinks about it.

boboldehampsink:
This is a breaking change for me? Now getting

```
(1/1) BindingResolutionException
Unresolvable dependency resolving [Parameter #0 [ <required> $model ]] in class Laravel\Scout\Builder
--
```

driesvints:
@boboldehampsink are you overwriting the `search` method?

boboldehampsink:
@driesvints I'm using https://github.com/worksweet/laravel-scout-mysql-driver

driesvints:
@boboldehampsink you'll have to raise an issue with them I'm afraid.

nunomaduro:
@boboldehampsink I am with @driesvints on this one. You should also take in consideration that driver is using the version `"laravel/scout": "^5.0"` - and not the latest one.

boboldehampsink:
I'm using a modified version (https://github.com/worksweet/laravel-scout-mysql-driver/pull/76) that I submitted myself - so I guess I'll have to fix this myself as well. Any recommendations?

nunomaduro:
Sorry buddy @boboldehampsink, I don't use that driver myself.  Please try to debug a little bit more.

If you can't come up with a solution I will give you a hand: Just say hi on the `scout` channel on the [Laravel Discord](https://discordapp.com/invite/KxwQuKb).

boboldehampsink:
The mysql driver does not overwrite the search method of the searchable trait...

boboldehampsink:
Found this to be a Lumen thing actually. The "app()" helper does not accept arguments there: https://github.com/laravel/lumen-framework/blob/5.7/src/helpers.php#L33-L39

Means there is no issue here. Thanks all.

Issue #188, opened 2017-05-25
Builder ignores Model custom Collection when getting data


ziabice:
In my project I use a Model that returns a custom Collection subclass (I redefined the `Model::newCollection() `method).

The problem is that Scout _Builder_ class completely ignores this new collection and instead returns a standard Eloquent\Collection: this breaks my code and seems to me the wrong behavior.

This issue affects `Builder::paginate()` and `Engine::get()`.

The solution to this problem is very trivial: let `Engine::map()` return the right Collection type (it should already, because of it's signature that tells that the return type will be a `\Illuminate\Database\Eloquent\Collection` or a subclass), then pass as-is the collection.

This means that at a certain point into your `Engine::map()` method you'll do:

    return $model->newCollection($results);

Attached there's a patch that should fix this annoying issue.

[commit-f30d7bf.txt](https://github.com/laravel/scout/files/1029879/commit-f30d7bf.txt)

driesvints:
Hmm, I can see this being useful. @nunomaduro @taylorotwell do you guys want to weigh in here? I could send in a PR for this. Just replacing `Collection::make` with `$model->newCollection` where applicable could do the trick.

PR #334, opened 2018-12-07
[6.0] Use Model collection where appropriate

driesvints:
This way when the model's collection is overwritten it still gets applied.

Fixes https://github.com/laravel/scout/issues/188

I'm not really sure if this is *breaking* or not. It's best that @nunomaduro has a look first at this. Especially the changes on the Engine's `get` call and the `AlgoliaEngine`'s `map` call need to be reviewed a bit thoroughly.

### DIFF
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -4,7 +4,6 @@ namespace Laravel\Scout\Engines;
 
 use Laravel\Scout\Builder;
 use AlgoliaSearch\Client as Algolia;
-use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Database\Eloquent\SoftDeletes;
 
 class AlgoliaEngine extends Engine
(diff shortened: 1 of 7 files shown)


nunomaduro:
Thanks @driesvints for this. I would prefer that we don't rush it, I will check it soon as I have time.

src/Engines/NullEngine.php
```diff
@@ -61,7 +62,7 @@ public function paginate(Builder $builder, $perPage, $page)
      */
     public function mapIds($results)
     {
-        return Collection::make();
+        return BaseCollection::make();
```
nunomaduro: This is not a breaking change, the contract specified on the phpdocs says:
```php
     * @return \Illuminate\Support\Collection
     */
    public function mapIds($results)
```

src/Engines/AlgoliaEngine.php
```diff
@@ -171,20 +170,16 @@ public function mapIds($results)
...
-            return Collection::make();
+            return $model->newCollection();
         }
 
-        $models = $model->getScoutModelsByIds(
-            $builder, collect($results['hits'])->pluck('objectID')->values()->all()
-        )->keyBy(function ($model) {
-            return $model->getScoutKey();
-        });
+        $objectIds = collect($results['hits'])->pluck('objectID')->values()->all();
```
nunomaduro: Your refactorization is a great job, but I would prefer that we keep this modification simple, with **only** the introduction of the `$model->newCollection()`:
```php
    public function map(Builder $builder, $results, $model)
    {
        if (count($results['hits']) === 0) {
            return $model->newCollection();
        }

        $models = $model->getScoutModelsByIds(
            $builder, collect($results['hits'])->pluck('objectID')->values()->all()
        )->keyBy(function ($model) {
            return $model->getScoutKey();
        });

        return $model->newCollection($results['hits'])->map(function ($hit) use ($models) {
            if (isset($models[$hit['objectID']])) {
                return $models[$hit['objectID']];
            }
        })->filter()->values();
    }
```

**Reason**: The current contract specified on the phpdocs says that `getScoutModelsByIds` must return mixed. Some people may return an array out of it.

driesvints: @nunomaduro at the moment there's a `keyBy` call right after the `getScoutModelsByIds` call so there's no way the current implementation can work without a collection object.

Issue #329, opened 2018-11-30
scout:import performance issues when toSearchableArray contains relationships


bkilshaw:
I recently made a model Searchable using Scout and Algolia. When running scout:import in production the whole production (Google App Engine) crashed. There was just over 15K records that needed to import. The SQL queries were going out of control (2000-2500 queries/second) and our production environment was down until the queued jobs all failed.

Someone in Discord (@fitztrev) mentioned that they ran into the same issue and tracked it down to the lack of eager loading of relationships mentioned in toSearchableArray(). Here is the method in my case:

```
public function toSearchableArray()
    {
        return [
            'provider_user_id' => $this->provider_user_id,
            'provider_name' => $this->provider->display_name,
            'receiver_user_id' => $this->receiver_user_id,
            'receiver_name' => $this->receiver->display_name,
            'receiver_hierarchy' => $this->receiver->role->hierarchy,
            'store_id' => $this->receiver->store_id,
            'category_id' => $this->category_id,
            'category_name' => $this->category->name,
            'subcategory_id' => $this->subcategory_id,
            'subcategory_name' => $this->subcategory->name,
            'comments' => $this->comments,
            'date_occurred'=> $this->date_occurred
        ];
    }
```
If the relationships aren't eager loaded it would make sense that this would generate an obscene number of queries.

I believe I could use `protected $with = ['provider', 'receiver', 'category', 'subcategory'];`  on the object, but that would mean I'm eager loading the relationships every single time I have an instance of this object, which I don't always want/need to do.

fitztrev:
Cross ref algolia/scout-extended#90

driesvints:
Did some research for this. The problem is indeed that the `makeAllSearchable` builds up a new query. Using `protected $with` would indeed solve this problem but can't be done for the reason stated above. We could solve this with an extra method on the `Searchable` trait. Something like this (or a better name):

```php
public function searchableWith(): array
{
    return ['provider', 'receiver', 'category', 'subcategory'];
}
```

Which then would be used in the `makeAllSearchable` method. I think this would be the only solution for this and still an appropriate one because like you stated yourself you'd only use this when doing an import. For all other purposes you'd do the eager loading beforehand.

driesvints:
Gonna mark this as a feature request rather than a bug because the temp solution would indeed be to use the `with` property.

terryBaz:
[...]

PR #342, opened 2019-01-05
[6.0] Add a searchableWith property to eager load some relations on batch indexing (scout:import)

terryBaz:
referring to [this issues](https://github.com/laravel/scout/issues/329).

todo :
- [ ] tests 
- [ ] update laravel scout docs (adding this property) 

> i really want to help for the tests but i did not know where to start, feedbacks will be appreciated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ composer.lock
 .DS_Store
 Thumbs.db
 /phpunit.xml
+/.idea
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -16,6 +16,13 @@ trait Searchable
     protected $scoutMetadata = [];
 
     /**
+     * The relations to eager load on batch indexing (scout:import).
+     *
+     * @var array
+     */
+    protected $searchableWith = [];
+
+    /**
      * Boot the trait.
      *
      * @return void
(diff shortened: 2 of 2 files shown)


driesvints:
Hmm, think it's best that we hold off on this implementation. @nunomaduro can you jump in on the issue to figure out a solution first before we accept a PR? Thanks!

fitztrev:
@nunomaduro @driesvints - I don't understand the problem with this PR. What you describe is the behavior I would expect to happen.

```php
class Post extends Model
{
    protected $with = ['user'];
}

class Post extends Model
{
    protected $searchableWith = ['user'];
}
```

The property `$searchableWith` behaves the same as `$with`, but allows you to load those relationships only in the context of indexing records. So if your `toSearchableArray` contains data from a relationship, you can eager load and make the `scout:import` command much faster and more efficient.

This PR fixes #329 and algolia/scout-extended#90 for me.

ultrono:
@fitztrev I think @nunomaduro takes issue with the fact that unless the `toSearchableArrray()` method is defined on the model, the relationship is also indexed.

Using the code from this PR, to index 16,500 records, with a single relationships (a booking mode, with a 'customer' relationship massively speeds up indexing. With eager loading it takes 2 mins 24 seconds. WIthout eager loading the time was 5 mins 09 seconds. 

I'm using the TNTSearch driver.

Issue #348, opened 2019-01-23
[7.0] Algolia API client v2


mvdnbrk:
Algolia rewrote their API client entirely and released v2 of their client two months ago.

Any plans on upgrading scout to use the latest version of the Algolia API client?

The upgrade doesn't seem to be very complicated, let me know if a PR for this is appreciated.

lucasmichot:
It's a very good reason to postpone 👍 
Congrats @nunomaduro !!

mvdnbrk:
If I have the time I will take a look into it and submit a PR.

We can release this by the end of february 😄

Congrats and have a great time @nunomaduro :tada:

PR #349, opened 2019-01-24
[7.0] Upgrade Algolia API client to v2

mvdnbrk:
This PR aims to upgrade the Algolia API client to v2.

See issue #348.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "illuminate/support": "~5.4"
     },
     "require-dev": {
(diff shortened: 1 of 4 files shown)


nunomaduro:
@taylorotwell The upgrade it's easy indeed, but I would prefer to **delay** this until mid February. If it's fine for you.

**Reason**: I would like to be close if support is needed. I and will be in honeymoon until mid February.

PR #355, opened 2019-02-13
[7.0] Allow a query scope when importing via `scout:import`

fitztrev:
I have a table with a few hundred thousand records, but only want to index a subset of them.

The `shouldBeSearchable` is too slow for this purpose because it iterates over each record individually after it's already fetched it. This lets you add a database query scope in your model so it's much faster.


taylorotwell:
It's unclear from your PR what scopeOnlySearchable would look like or do.

fitztrev:
Resubmitted with example in #356.

PR #358, opened 2019-02-25
[7.0] DRY queueMakeSearchable. Remove return statement in queueMakeSearchable

matchish:
We have duplicated code in queueMakeSearchable and MakeSearchable. Also, queueMakeSearchable declared 'void' so it must not return a value

### DIFF
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -60,7 +60,7 @@ trait Searchable
         }
 
         if (! config('scout.queue')) {
-            return $models->first()->searchableUsing()->update($models);
+            dispatch_now(new MakeSearchable($models));
         }
 
         dispatch((new MakeSearchable($models))


src/Searchable.php
```diff
@@ -60,7 +60,7 @@ public function queueMakeSearchable($models)
         }
 
         if (! config('scout.queue')) {
-            return $models->first()->searchableUsing()->update($models);
+            dispatch_now(new MakeSearchable($models));
```
driesvints: The return is to prevent it from being fired twice. With this change the command will be fired again with the dispatch below it.

matchish: We can dispatch in 'else' section instead of immediately return
```
        if (! config('scout.queue')) {
            dispatch_now(new MakeSearchable($models));
        } else {
            dispatch((new MakeSearchable($models))
                    ->onQueue($models->first()->syncWithSearchUsingQueue())
                    ->onConnection($models->first()->syncWithSearchUsing()));
    }
```

matchish: I've added new commit in the local repo but it didn't appear in the PR. Can I open a new one?

driesvints: yes

PR #359, opened 2019-02-26
[8.0] Rename and change queueMakeSearchable and queueRemoveFromSearch. Add RemoveFromSearch job

matchish:
queueMakeSearchable has code duplication with MakeSearchable job
queueMakeSearchable must not return value as it declared as void
queueRemoveFromSearch and queueMakeSearchable renamed because it dispatches job but doesn't push it to queue if scout.queue is false
I can't write tests for the methods because package don't have 'config' 'dispatch' and 'dispatch_now' helpers so I should change signatures of  queueMakeSearchable and queueRemoveFromSearch methods and pass dependencies as arguments
Added RemoveFromSearch job
ifEmpty check on collection removed because of it duplicated in MakeSearchable and RemoveFromSearch jobs

### DIFF
--- a/src/Jobs/RemoveFromSearch.php
+++ b/src/Jobs/RemoveFromSearch.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Laravel\Scout\Jobs;
+
+use Illuminate\Bus\Queueable;
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Contracts\Queue\ShouldQueue;
+
+class RemoveFromSearch implements ShouldQueue
+{
+    use Queueable, SerializesModels;
+
+    /**
+     * The models to be made searchable.
+     *
+     * @var \Illuminate\Database\Eloquent\Collection
+     */
+    public $models;
+
+    /**
+     * Create a new job instance.
+     *
+     * @param  \Illuminate\Database\Eloquent\Collection  $models
+     * @return void
+     */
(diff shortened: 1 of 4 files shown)


driesvints:
Eh, I'm not so sure if all these changes are necessary...

matchish:
I agree that it works without these changes)
But it's hard to test without these changes(dirty tricks requires)
If you open SearchableTest you can see that we don't have tests wich cover dispatching MakeSearchable job
But we have a dirty trick with config function(overided in the test)
Also queueFromSearch delete records from search immediately, not in queue

matchish:
Now have more tests. And these test less fragile because don't rely on implementation details

taylorotwell:

matchish:
@taylorotwell Thanks for reply. No, problem, I understand.
What about this line? https://github.com/laravel/scout/blob/484ff556fe2def2eab9f8eb2e60becba36c439d9/src/Searchable.php#L83
Should we introduce RemoveFromSearch job?

jshayes:
It used to be like that, it seems like it was intentionally deleted at some point.
https://github.com/laravel/scout/commit/f46b2b8a343162cf227afeb2dcd3709cb4bc626b

Also, it seems like there would be an issue when queueing up deletes, when the object is deleted by the time the job runs. The serializes models trait won't be able to unserialize the models since they no longer exist.

Issue #182, opened 2017-05-10
How to import a larg table? Query scanning too much rows


182050940:
scout:import use a query like:

```
select * from `articles` where `articles`.`deleted_at` is null order by `id` asc limit 500 offset 1000000
```
select query with offset limit is too much slow

driesvints:
This should now be chunked, can you check again in the latest version and report if this is still a problem? Thanks!

yuloh:
It is chunking but it looks like it's [using `chunk`](https://github.com/laravel/scout/blob/8177460f69cb1eb1d4ce857fdcd39a52c3c7c575/src/SearchableScope.php#L34) which [uses limit/offset](https://github.com/laravel/framework/blob/d6fc4cfa470b2807c50f8371739627aea32a34fc/src/Illuminate/Database/Concerns/BuildsQueries.php#L28).

To prevent using limit/offset it would need to use [chunkById](https://github.com/laravel/framework/blob/5.8/src/Illuminate/Database/Query/Builder.php#L2225) instead.

driesvints:
@yuloh can you send in a PR with a test?

PR #360, opened 2019-03-05
[8.0] Use chunkById instead of chunk

yuloh:
Using chunkById is dramatically faster when importing millions of rows.  This issue was originally reported in #182.

On a table with 1,632,576 rows, using the null scout driver so we don't have any noise from HTTP requests, an import goes from 29 minutes and 57 seconds to 28 seconds.

Before:

```
$ time php artisan scout:import "App\User"
All [App\User] records have been imported.
php artisan scout:import "App\User"  14.80s user 3.69s system 1% cpu 29:57.14 total
```

After:

```
$ time php artisan scout:import "App\User"
php artisan scout:import "App\User"  17.18s user 4.06s system 73% cpu 28.809 total
```

Concerns:

`chunkById` currently removes all ORDER BY clauses.  If someone had called `searchable` like this it wouldn't work as expected anymore:

```php
App\Order::where('price', '>', 100)->orderBy('price')->searchable();
```

I doubt that anyone actually needs that behavior?  If they do there are a few options:

- Make this opt in with a config setting, i.e. `config('scout.chunk.by_id')`
- Make `chunkById` preserve order by clauses.  This is [explained here](https://use-the-index-luke.com/sql/partial-results/fetch-next-page) but it's a little confusing.  I've implemented this before for a proprietary ElasticSearch indexer.  It shouldn't be too difficult to add to Eloquent.

### DIFF
--- a/src/SearchableScope.php
+++ b/src/SearchableScope.php
@@ -31,7 +31,7 @@ class SearchableScope implements Scope
     public function extend(EloquentBuilder $builder)
     {
         $builder->macro('searchable', function (EloquentBuilder $builder, $chunk = null) {
-            $builder->chunk($chunk ?: config('scout.chunk.searchable', 500), function ($models) {
+            $builder->chunkById($chunk ?: config('scout.chunk.searchable', 500), function ($models) {
                 $models->filter->shouldBeSearchable()->searchable();
 
                 event(new ModelsImported($models));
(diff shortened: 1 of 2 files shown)


driesvints:
Hmm, I agree that the trad-off is definitely worth it. Wondering what @nunomaduro thinks.

PR #372, opened 2019-04-18
[7.0] Calling values() on sorted collection to reset the array keys.

Redshift-Web-Design:
This prevents un-expected sorting when working with the json data. The tests still passed because plucking the id from the sorted collection will give you the expected sort order. Working with the data in JS, for example, will not. Not confident enough with testing to devise a test for this I'm sorry :-(

### DIFF
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -191,7 +191,7 @@ class AlgoliaEngine extends Engine
                 return in_array($model->getScoutKey(), $objectIds);
             })->sortBy(function($model) use ($objectIdPositions) {
                 return $objectIdPositions[$model->getScoutKey()];
-            });
+            })->values();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


driesvints:
I think an extra assertion in the test added [here](https://github.com/laravel/scout/pull/369/commits/dafbfa74031c660f34f93a64b566e6164f9e5baa) which checks the full array of the results would help and verify the behavior.

Redshift-Web-Design:
Have added an extra line or so to that test. Doesn't feel very elegant, but it does produce a failed test without the values() method call. I might leave the testing to someone more experienced in that area :-)

driesvints:
Hmm, wasn't exactly what I had in mind. I was more thinking of verifying the exact structure of the results. I'll try to whip up something for this soon.

Redshift-Web-Design:
Thanks for the feedback Dries. I have committed something much simpler in-case it's of help! Hope you are killing it mate!

PR #376, opened 2019-05-09
Fixing PHPdoc referencing session driver

ChrisThompsonTLDR:
I think this is a copy/paste from https://github.com/laravel/framework/blob/e6c8aa0e39d8f91068ad1c299546536e9f25ef63/src/Illuminate/Session/SessionManager.php#L204 and is not meant to say `session` but rather `Scout`.


driesvints:
Thanks!

PR #384, opened 2019-09-10
Add __call() method to AlgoliaEngine

tillkruss:
This allows calling methods like `multipleQueries()` on the Algolia search client easily.

I'm not sure if tests are needed for this. Please advise.

This PR allows calls like this:

```php
User::searchableUsing()->multipleQueries();

// or

app(EngineManager::class)->engine()->multipleQueries($indices);
```

### DIFF
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -228,4 +228,16 @@ class AlgoliaEngine extends Engine
     {
         return in_array(SoftDeletes::class, class_uses_recursive($model));
     }
+    
+    /**
+     * Dynamically call the Algolia client instance.
+     *
+     * @param  string  $method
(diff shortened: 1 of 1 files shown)


nunomaduro:
The usage would be something like this, correct?

```php
User::searchableUsing()->multipleQueries();
```

tillkruss:
Correct, I updated the description.

nunomaduro:
Alright, I do think that's a useful feature. @taylorotwell @driesvints Willing to add this?

@tillkruss If they agree, do you mind of adding some tests?

tillkruss:
@nunomaduro: I wouldn't know what to test TBH without linking the tests to Algolia's `SearchClient` itself, which might change and I'd say it's up to the developer to test their call themselves, not this package. Do you have a suggestion?

PR #386, opened 2019-09-22
[7.0] Removed legacy phpunit flag

joaorobertopb:
This **PR** will:

- Remove legacy phpunit flag,  **syntaxCheck="false"**:
    - That configuration setting has not had an effect in years. It was removed long ago.
- Ignore **.phpunit.result.cache**

**See:**
phpunit-scout


driesvints:
Thanks

PR #405, opened 2020-06-10
update name of api key field

wilfra:
"Admin API Key" is the term Algolia uses in their dashboard. They don't use the term 'Secret' and using it here is somewhat confusing.


driesvints:
Not worth the breaking change tbh. This could be better clarified in the docs.

PR #406, opened 2020-06-21
[8.x] Custom ‘with’ eager loading definitions for bulk searchable indexing

serpentblade:
Adds support for `$searchableWith` as an override to the built-in `$with` member on models.  This allows a custom eager loading strategy specific to indexing requirements

Related Feature Request: #329

### DIFF
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -16,6 +16,15 @@ trait Searchable
     protected $scoutMetadata = [];
 
     /**
+     * The relations to eager load loading models via Scout
+     * Uses model's default $with value when set to `null`.
+     * @see \Illuminate\Database\Eloquent\Model::$with
+     *
+     * @var array | null
+     */
+    protected $searchableWith = null;
+
+    /**
      * Boot the trait.
      *
      * @return void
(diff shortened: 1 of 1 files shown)


taylorotwell:

serpentblade:
Hi @taylorotwell,

I completely understand and respect the desire to keep the framework minimal due to maintenance requirements.  While this does not correct a bug in the framework in the traditional sense, I feel it corrects a performance limitation with regard to how data is loaded by the `scout:import` command.

As noted in #329, the requirements for eager loading when indexing can often be very different from the requirements for general use of model in the framework.   For example, I may want to include a number of additional fields from relations to be used for faceted search, but I would not necessarily want those same relations eagerly loaded in my actual application.

It should also be noted that some of the solutions suggested, such as overriding `makeAllSearchable()` feels like a generally undesirable solution as it always runs the risk of the original implementation changing in future versions.  I would deem that a workaround rather than a solution.

I did note another PR that was rejected with a similar solution (#342), but made sure to avoid the same breaking changes it introduced. My implementation is meant to provide an _optional_ set of eager loading rules, meaning it retains full backwards compatibility with existing usage of `$with`, and that only when a  dev explicitly defines the new `$searchableWith` parameter will it affect the output from the default implementation of `toSearchableArray()` and override the existing eager loading rules defined in `$with`.

Ultimately, I will respect your decision on this either way.  But I would be remiss to not attempt to argue for the performance benefits this introduces to the indexing process.  Coupled with my other PR that was recently merged (thank you), it provides the other half of the missing pieces to making the scout:import feel feature complete.

Issue #385, opened 2019-09-10
Add user identification by default?


tillkruss:
From the Algolia docs:

>For analytics query aggregation to work properly, the analytics engine relies on a user id.
>
>Many front-end solutions tag every search with the end-user’s HTTP request IP. However, it is important to not rely on this behavior. You should explicitly create a unique user identifier that you maintain in your own application. For example, you could use a unique id you store in a database.
>
>When using a back-end API client, you want to:
>
> - set the X-Forwarded-For header to forward the user IP to Algolia
> - set the userToken search parameter to tell Algolia which user is querying

Should we add this by default? Happy to open a PR.

 - https://www.algolia.com/doc/api-reference/api-methods/set-extra-header/
 - https://www.algolia.com/doc/guides/getting-insights-and-analytics/search-analytics/out-of-the-box-analytics/how-to/specify-which-user-is-doing-the-search/

nunomaduro:
Overall I think it's a good idea, you may also want to keep in mind this issue https://github.com/laravel/scout/issues/379 - as they are related and both want to modify the `SearchConfig` instance.

How would you see this concerning the developer experience?

tillkruss:
I assume @moosapour-arash won't be sending a PR, so this could be part of it. Great tip!

My first thoughts about solving this were automatically adding `X-Algolia-UserToken` and `X-Forwarded-For` as default headers. Probably a config toggle to turn it off for the Europeans.

tillkruss:
I guess this could be done fairly easily on a per-app basis:

```php
<?php

namespace App\Providers;

use Laravel\Scout\EngineManager;
use Laravel\Scout\Engines\AlgoliaEngine;

use Algolia\AlgoliaSearch\Support\UserAgent;
use Algolia\AlgoliaSearch\Config\SearchConfig;

use Illuminate\Support\ServiceProvider;

class SearchServiceProvider extends ServiceProvider
{
    public function boot()
    {
        UserAgent::addCustomUserAgent(
            config('app.name'),
            env('HEROKU_RELEASE_VERSION', 'testing')
        );
        
        resolve(EngineManager::class)->extend('algolia', function () {
            $config = SearchConfig::create(
                config('scout.algolia.id'),
                config('scout.algolia.secret')
            );

            $config->setDefaultHeaders([
                'X-Forwarded-For' => '...',
                'X-Algolia-UserToken' => '...',
            ]);

            return new AlgoliaEngine(
                SearchClient::createWithConfig($config),
[...]

PR #411, opened 2020-07-30
[8.x] Auto identify user

driesvints:
This PR will auto set the user to be identified through Algolia through a config option. Other engines can make use of this new option as well if they want. Turned off by default to avoid breaking changes but I'd recommend to turn this on by default in the next major version.

However: I'm currently not seeing user stats appearing in the dashboard when I test:

Screenshot 2020-07-30 at 21 45 05

Headers are being set properly and all so I don't know what I'm missing here. Need to ping @nunomaduro about it. In draft until this is resolved.

Closes https://github.com/laravel/scout/issues/385

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -77,11 +77,11 @@ return [
     | Auto Identify User
     |--------------------------------------------------------------------------
     |
-    | This option allows to control whether to automatically
-    | identify the currently authenticated user with your
-    | search engine of choice. Disabled by default.
+    | This option allows you to control whether to automatically
+    | identify the currently authenticated user with a search
+    | search engine of choice. It is turned off by default.
     |
-    | Support engines: "algolia"
+    | Supported engines: "algolia"
     |
     */
 
--- a/config/scout.php
+++ b/config/scout.php
@@ -74,6 +74,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Auto Identify User
+    |--------------------------------------------------------------------------
+    |
+    | This option allows to control whether to automatically
+    | identify the currently authenticated user with your
+    | search engine of choice. Disabled by default.
+    |
+    | Support engines: "algolia"
+    |
+    */
+
+    'user' => env('SCOUT_USER', false),
+
+    /*
+    |--------------------------------------------------------------------------
     | Algolia Configuration
     |--------------------------------------------------------------------------
     |
--- a/config/scout.php
+++ b/config/scout.php
@@ -79,13 +79,13 @@ return [
     |
     | This option allows you to control whether to automatically
     | identify the currently authenticated user with a search
-    | search engine of choice. It is turned off by default.
+    | engine that is supported. By default it's turned off.
     |
     | Supported engines: "algolia"
     |
     */
 
-    'user' => env('SCOUT_USER', false),
+    'identify_user' => env('SCOUT_IDENTIFY_USER', false),
 
     /*
     |--------------------------------------------------------------------------
--- a/src/EngineManager.php
+++ b/src/EngineManager.php
@@ -6,7 +6,6 @@ use Algolia\AlgoliaSearch\Config\SearchConfig;
 use Algolia\AlgoliaSearch\SearchClient as Algolia;
 use Algolia\AlgoliaSearch\Support\UserAgent;
 use Exception;
-use Illuminate\Http\Request;
 use Illuminate\Support\Manager;
 use Laravel\Scout\Engines\AlgoliaEngine;
 use Laravel\Scout\Engines\NullEngine;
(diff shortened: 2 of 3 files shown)


config/scout.php
```diff
@@ -72,6 +72,21 @@
...
+    /*
+    |--------------------------------------------------------------------------
+    | Auto Identify User
+    |--------------------------------------------------------------------------
+    |
+    | This option allows to control whether to automatically
+    | identify the currently authenticated user with your
+    | search engine of choice. Disabled by default.
+    |
+    | Support engines: "algolia"
```
jbrooksuk, line 84: Supported*

config/scout.php
```diff
@@ -72,6 +72,21 @@
 
     'soft_delete' => false,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Auto Identify User
+    |--------------------------------------------------------------------------
+    |
+    | This option allows to control whether to automatically
```
jbrooksuk, line 80: This option allows you*

src/EngineManager.php
```diff
@@ -59,6 +65,31 @@ protected function ensureAlgoliaClientIsInstalled()
         throw new Exception('Please install the Algolia client: algolia/algoliasearch-client-php.');
     }
 
+    /**
+     * Set the default an Algolia config headers.
```
jbrooksuk, line 69: Drop the "an"

config/scout.php
```diff
@@ -72,6 +72,21 @@
 
     'soft_delete' => false,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Auto Identify User
+    |--------------------------------------------------------------------------
+    |
+    | This option allows you to control whether to automatically
+    | identify the currently authenticated user with a search
```
nunomaduro, line 81: The `search` word is duplicated here.

src/EngineManager.php
```diff
@@ -59,6 +65,31 @@ protected function ensureAlgoliaClientIsInstalled()
...
+     * @return array
+     */
+    protected function defaultAlgoliaHeaders()
+    {
+        if (! config('scout.user')) {
+            return [];
+        }
+
+        $headers = [];
+        $request = $this->container->make(Request::class);
```
nunomaduro, line 80: Because we are using `config()` on the 75, maybe we can use `request()` here too?

src/EngineManager.php
```diff
@@ -76,10 +107,10 @@ public function createNullDriver()
      */
     public function getDefaultDriver()
     {
-        if (is_null($this->app['config']['scout.driver'])) {
+        if (is_null($this->container['config']['scout.driver'])) {
```
nunomaduro, line 110: Because we are using `config()` on the 75, maybe we can use  `config()` here too?

src/EngineManager.php
```diff
@@ -76,10 +107,10 @@ public function createNullDriver()
...
             return 'null';
         }
 
-        return $this->app['config']['scout.driver'];
+        return $this->container['config']['scout.driver'];
```
nunomaduro, line 114: Because we are using `config()` on the 75, maybe we can use  `config()` here too?

config/scout.php
```diff
@@ -72,6 +72,21 @@
...
+    | search engine of choice. It is turned off by default.
+    |
+    | Supported engines: "algolia"
+    |
+    */
+
+    'user' => env('SCOUT_USER', false),
```
nunomaduro, line 88: I would probably change this for:

- `'user_analytics' => env('SCOUT_USER_ANALYTICS', false),`
- or `'identify_user' => env('SCOUT_IDENTIFY_USER', false),`

I like the first better I think.

src/EngineManager.php
```diff
@@ -59,6 +65,31 @@ protected function ensureAlgoliaClientIsInstalled()
...
+    protected function defaultAlgoliaHeaders()
+    {
+        if (! config('scout.user')) {
+            return [];
+        }
+
+        $headers = [];
+        $request = $this->container->make(Request::class);
+
+        if (filter_var($ip = $request->ip(), FILTER_VALIDATE_IP, FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE)) {
```
nunomaduro, line 82: If there is no request at all ( testing, cli ), do we still want to modify the headers with his information?

src/EngineManager.php
```diff
@@ -59,6 +65,31 @@ protected function ensureAlgoliaClientIsInstalled()
...
+            $headers['X-Forwarded-For'] = $ip;
```
nunomaduro, line 83: I think its cool if you run some tests on Algolia dashboard. Like the same userToken, but different IPs shouldn't be consider a different user. You need to test this.

src/EngineManager.php
driesvints: It should already strip things out for ips like 127.0.0.1 etc because of the `filter_var` so I think we're good here.

driesvints: I've added a `! config('app.debug')` check.

config/scout.php
driesvints: I went with the latter 😁

driesvints:
Everything looking good now in Algolia dashboard:

Screenshot 2020-07-31 at 10 23 27

mstaack:
oh brilliant!

PR #413, opened 2020-08-19
Update builder class pagination methods to resolve LengthAwarePaginator using container

### LINKED ISSUE #371: Creating the Paginator using the container
It would be great, if the `LengthAwarePaginator` in `Laravel\Scout\Builder` would be created using the `Container` - like in `Illuminate\Database\Eloquent\Builder` - so that we can customize it or automatically append `query`parameters.

stevenmaguire:
This PR attempts to address to fix #371. The LengthAwarePaginator was being instantiated directly in the pagination methods. This solution intends to use the container to resolve that class instead. This will give other container-based configurations a chance to hook into the paginators emitted by Scout in the same way that they can hook into Paginators emitted by Eloquent.

The specific use case that leads me down this path was an observation that the following Laravel-specific code was not affecting LengthAwarePaginators from Scout.

```php
        $app->resolving(LengthAwarePaginator::class, static function (LengthAwarePaginator $paginator) {
            return $paginator->appends(request()->query());
        });

        $app->resolving(Paginator::class, static function (Paginator $paginator) {
            return $paginator->appends(request()->query());
        });
```

The test coverage for this class was pretty sparse. No existing tests break. I am happy to write some more specific tests to exercise the container work if needed.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Scout;
 
+use Illuminate\Container\Container;
 use Illuminate\Pagination\LengthAwarePaginator;
 use Illuminate\Pagination\Paginator;
 use Illuminate\Support\Traits\Macroable;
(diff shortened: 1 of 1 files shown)


stevenmaguire:
Thanks for merging this - hopefully, it helps others out! Is it trivial to cut a new release?

driesvints:
@stevenmaguire release day is tomorrow

stevenmaguire:
Great! Thanks for the clarification.

PR #422, opened 2020-10-14
Add ability to eager load relations when calling scout:import

jwohlfert23:
When loading data into the search index using the `scout:import` command while having a `toSearchableArray()` method that includes data from relationships like so...
```php
public function toSearchableArray() {
        $data = Arr::only($this->toArray(), [
            'id', 'first_name', 'last_name', 'email', 'department'
        ]);
        $data['customer'] = optional($this->customer)->name;
        return $data;
}
```

It introduces an n+1 problem and if you have a lot of data to sync to an index can be burdensome on the database and very slow. 

This PR introduces a new method on the Searchable trait named `searchableEagerLoads()` which is called in the `makeAllSearchable` method, which is what `scout:import` uses under the hood.

Therefore, to solve the above problem, one could add this method to their model and eliminate the n+1 problem:

```php
public function searchableEagerLoads() {
     return ['customer'];
}
```

This is my first PR to a Laravel repo, so I'm open to any feedback/thoughts.

### DIFF
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -106,7 +106,7 @@ trait Searchable
             'model' => new static,
             'query' => $query,
             'callback' => $callback,
-            'softDelete'=> static::usesSoftDelete() && config('scout.soft_delete', false),
+            'softDelete' => static::usesSoftDelete() && config('scout.soft_delete', false),
         ]);
     }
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Just pushed a commit to add a `makeAllSearchableUsing` method which receives the query and can modify it however you wish, including calling `with`:

image

jwohlfert23:
And...that's a much better solution than mine :) Appreciate it, Taylor!

PR #424, opened 2020-10-22
Allow ordering of import command results.

Sladewill:
Allow ordering of import command results for certain search engines such as elastic search where the more recent data is more critical to import first to make a smoother transition.


driesvints:
@Sladewill can you fix the tests?

Sladewill:
Yeah just looking at that now.

Sladewill:
Tests are passing now.

PR #425, opened 2020-11-03
[8.x] PHP 8 Support

driesvints:
Currently blocked by https://github.com/algolia/algoliasearch-client-php/pull/644


driesvints:
We've decided to remove the blocking dependency so the base package can already be patched for other drivers. Anyone who's using Algolia will need to wait until Algolia has patched their package (see the PR above).

PR #432, opened 2020-12-08
Removes Algolia Dev Dependency

freshleafmedia:
This PR removes the Aloglia dev dependency.

I believe this is a good idea for a number of reasons:

- While many developers will be using Algolia with this package it isn't a requirement
- If the Algolia package is missing and you attempt to use the Algolia engine an exception is thrown which explains what to do. See [src/EngineManager.php#L54](https://github.com/laravel/scout/blob/8.x/src/EngineManager.php#L54)
- Having it as a dev dependency means it's possible for an application to work in a local environment where dev packages are installed but when deployed to a production environment, where dev dependencies aren't installed, it will break.
- It allows #425 to be merged

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,6 @@
         "illuminate/support": "^6.0|^7.0|^8.0"
     },
     "require-dev": {
-        "algolia/algoliasearch-client-php": "^2.2",
         "mockery/mockery": "^1.0",
         "phpunit/phpunit": "^8.0|^9.3"
     },


driesvints:
Algolia is the main search engine for this package and without shipping it Scout would basically be an empty shell. We have no intention on removing it as the default. Also, this PR is a breaking change on a current stable release.

I know everyone's waiting for PHP 8 support in Scout but there's nothing we can do but wait until Algolia ships PHP 8 support on their package.

driesvints:
I'm sorry, I'm incorrect. This indeed wouldn't be a breaking change. And this indeed doesn't seems to have such a big impact, strange. 

I'll consider this PR a bit today before I reopen this.

freshleafmedia:
Thanks. In my use case I'm using Scout without Algolia so the dependency doesn't make sense and is also limiting my ability to use PHP 8.

driesvints:
I'm sorry I didn't get to this again. I'll come back to this later this week.

driesvints:
@freshleafmedia thanks for this 👍

Issue #152, opened 2017-01-09
Aborted save with transaction rollback should not add to index


adamthehutt:
When a model save is wrapped in a transaction that is rolled back, the model is still added to the searchable index, even though it doesn't actually exist in the database. Unfortunately I don't know enough about the innards of Scout to suggest a fix, but presumably it needs to delay queueing the job until it's certain that the save has been committed. (All of this also applies to updates.)

An added benefit of fixing this, for those using DatabaseTransactions in unit tests, is that it would solve issue #34 as well.

jitendra-1217:
@adamthehutt : Did you find any solutions (or workaround) it?

adamthehutt:
@jitendra-1217 No. It's a complicated issue to solve. It's possible that the best solution would be to link the Queue actions to any surrounding DB transaction. If the transaction rolls back, the queued jobs are dropped. If it is committed, the jobs are passed to the queue runner. Personally I can't imagine a situation where that wouldn't be the desired behavior, but it would break BC.

adamthehutt:
There are other references to this problem (not having to do with Scout). No one seems to have a good workaround or solution. I think it's a general Laravel issue that needs to be addressed.

https://laracasts.com/discuss/channels/general-discussion/queues-within-db-transactions
http://stackoverflow.com/questions/35764150/combining-transactions-with-queues-in-laravel

There may be a solution involving [job events](https://laravel.com/docs/5.4/queues#job-events) but the docs aren't very clear.

adamthehutt:
Here's another reference to the same issue, with a different proposed solution:
https://laravel.io/forum/03-03-2016-combining-transactions-with-queues-in-laravel

joshforbes:
I think I have a related problem:

In an app I am working on there is a fairly long company registration process (persists about 8 records).  This process is wrapped in a transaction. The first record that has to be created is searchable. When this object is created it fires off the scout queue job.  When the scout queue job serializes models, the transaction sometimes has not yet been committed. Meaning that $this->models will be empty and no record will get indexed.

I agree that it is more of a Laravel issue. I would think the ideal solution would be that create/save events should not be fired off until a transaction is committed. But I assume this is either difficult or there is a reason it is this way.

mhd21:
https://github.com/fntneves/laravel-transactional-events

driesvints:
Going to close this as this is more of a Laravel issue then it is a Scout issue as mentioned above.

PR #436, opened 2020-12-21
[8.x] Run observer callbacks after database transactions have committed

bakerkretzmar:
This PR sets `public $afterCommit = true` on Scout's `ModelObserver` class, so that the `saved`, `deleted`, `forceDeleted`, and `restored` model event callbacks, if they are triggered inside a database transaction, are not executed until that transaction has committed.

**Background**

This fixes an existing issue where incorrect data is synced to a search index if a model event Scout listens for, such as `saved`, is fired inside a database transaction.

If Scout is not set up to queue search index updates, this issue affects all code that fires the above model events inside database transactions, because Scout's observer method is executed immediately—before the transaction has committed or been rolled back—syncing stale or invalid data.

If Scout _is_ set up to use the queue this issue will appear intermittently, when the queue workers pick up and execute the syncing job before the transaction has committed or rolled back. In my experience this happens almost all the time.

See #152 and laravel/nova-issues#1906.

It wasn't really possible to address this at all up until now, but transaction-aware code execution was just added to Laravel core (see laravel/framework#35373 and laravel/framework#35434). In Laravel 8.17.0 and later, it's trivial to delay any listeners and observers until open database transactions have committed.

**Backwards compatibility**

Theoretically this could change some behaviour in existing Scout installations, but I would argue that that behaviour is a bug and should be fixed. I can't imagine a situation where an application would be relying on stale or invalid data being persisted to a search index, so I doubt this change would break anything.

**Tests**

Since model observers are basically just larger listener classes, in my opinion this PR is covered by Laravel's tests for delaying listener execution until transactions have committed: https://github.com/laravel/framework/blob/8.x/tests/Integration/Events/ListenerTest.php.

### DIFF
--- a/src/ModelObserver.php
+++ b/src/ModelObserver.php
@@ -7,6 +7,13 @@ use Illuminate\Database\Eloquent\SoftDeletes;
 class ModelObserver
 {
     /**
+     * Only dispatch the observer's events after all database transactions have committed.
+     *
+     * @var bool
+     */
+    public $afterCommit = true;
+
+    /**
      * The class names that syncing is disabled for.
      *
      * @var array


driesvints:
I've sent in a PR to revert this because this causes test suite that have the `DatabaseTransactions` trait applied to fail. We can make this change in the next major release. Feel free to send in a PR to master.

PR #441, opened 2021-01-03
[8.x] Run observer callbacks after transactions by default

bakerkretzmar:
Following up on #436, #437, and #440, this PR changes the default value of the new `after_commit` config option to `true`. The current behaviour, with this option set to `false`, causes stale data to be synced to search indexes in applications using database transactions (see #436).

I'm targeting the `8.x` branch again because it's very unlikely that this change will affect consuming application code, and the config option now makes it easy to disable when necessary, like in test code that uses database transactions.

Personally, I think it's preferable to make this behaviour the default as soon as possible. If you folks are more comfortable making that change in the next major version of Scout I'm happy to update this PR to target the `master` branch instead, let me know. I'll PR an update to the docs soon too. Thanks for all your attention to this!


taylorotwell:
Why is it necessary to make this change as soon as possible? Hasn't the current code been the default behavior of Scout since its initial release?

bakerkretzmar:
@taylorotwell yes, the current behaviour has been in place since the beginning. I just think it's a bug and we should fix it now since we're able to, rather than waiting. To be honest I'm surprised it hasn't caused more issues for people, especially since Nova is affected by it.

taylorotwell:
I sympathize with your point of view but think I will leave the default as-is for now just to avoid a breaking change. It matches the default behavior of the queues in Laravel. We could re-evaluate this in a future release.

PR #442, opened 2021-01-08
[8.x] Use the Config facade instead

gocanto:
## Description

This PR replaces the use of the `config()` helper to retrieve the `after_commit` value when creating the `ModelObserver`.

The reason why we would want to have this change is, we would be able to mock the configuration object when `not` extending the application within unit tests.

As these changes stand by now, they break all tests that call/create `models` using the `searchable` trait when extending the PHPUnit tests case class `[PHPUnit\Framework\TestCase]`.

Even though it is known we could use these helpers and we should be extending the framework test case class to work on apps tests, they are not resolved if the Laravel app is not present (in the unit context).

Reference: https://github.com/laravel/scout/pull/440

Thank you.

### DIFF
--- a/src/ModelObserver.php
+++ b/src/ModelObserver.php
@@ -3,6 +3,7 @@
 namespace Laravel\Scout;
 
 use Illuminate\Database\Eloquent\SoftDeletes;
+use Illuminate\Support\Facades\Config;
 
 class ModelObserver
 {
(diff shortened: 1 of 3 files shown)


gocanto:
any idea why the facade is not present there since the config helper resolved those fields out of the app too? - unless I am missing something.

gocanto:
@taylorotwell @driesvints thank you guys. Would we be able to have these changes within the next release? If so, when will it be?

Cheers!

driesvints:
Releases are done on Tuesdays.

PR #449, opened 2021-02-08
[8.x] Add option to disable global query scopes

Kovah:
Initial problem: while using Scout's import command, Scout received only a fraction of model entries from the database because I have a global query scope defined.

This pull requests includes two changes:
- The `makeAllSearchable()` function in the `Searchable` trait now accepts a second option `$disableGlobalScopes`, that disables global query scopes using the withoutGlobalScopes() function.
- The import command accepts a new `--disable-scopes` option, which passes the option to the makeAllSearchable() function mentioned above.

I tested this feature with both the Algolia and the TNT Search driver. In both cases all model entries were successfully imported.

### DIFF
--- a/src/Console/ImportCommand.php
+++ b/src/Console/ImportCommand.php
@@ -15,7 +15,8 @@ class ImportCommand extends Command
      */
     protected $signature = 'scout:import
             {model : Class name of model to bulk import}
-            {--c|chunk= : The number of records to import at a time (Defaults to configuration value: `scout.chunk.searchable`)}';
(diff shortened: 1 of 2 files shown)


Kovah:
I think the CI checks must be re-run, P7.3 L8.0 failed because of a Github 500 error.

driesvints:
Re-ran them

taylorotwell:
This is a breaking change since you change a method signature.

Note you can extend the `makeAllSearchableUsing` method for your model and run $query->withoutGlobalScopes() there in your own code.

image

PR #457, opened 2021-04-08
[9.x] Meilisearch index command

driesvints:
Ports over the `IndexCommand` from Meilisearch which is pretty handy for managing your indexes. I've also implemented a generic way so it works for all engines.

### DIFF
--- a/src/Console/IndexCommand.php
+++ b/src/Console/IndexCommand.php
@@ -39,7 +39,7 @@ class IndexCommand extends Command
             if ($this->option('delete')) {
                 $engine->deleteIndex($this->argument('name'));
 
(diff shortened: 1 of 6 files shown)


driesvints:
I'll make this generic so it can be added into any engine really.

mmachatschek:
We have unit tests that could be added too. https://github.com/meilisearch/meilisearch-laravel-scout/blob/main/tests/Unit/MeilisearchConsoleCommandTest.php

taylorotwell:
I split this into `scout:index` and `scout:delete-index`.

driesvints:
@mmachatschek can you PR that?

PR #460, opened 2021-04-08
Rename Index command

chrysanthos:
Doesn't really matter but the class only creates an index now since the [delete part](https://github.com/laravel/scout/commit/c30f194c9005da3cff4868e3a228fbf71cda2a3a) was moved to a new command so I thought I'd give it a more descriptive name.


driesvints:
@chrysanthos I've reverted this PR. In general we keep our command class names the same as the command itself.

chrysanthos:
No worries @driesvints! 👍

Issue #474, opened 2021-05-31
Option to set `MakeSearchable`'s retries and backoff


MatanYadaev:
Every day a few exceptions are thrown in my application because of connection issues with Algolia, reported about it [here](https://github.com/laravel/scout/issues/427).
I think the option to increase the number of retries and backoff could eliminate these connection issues with Algolia, and make Scout more configurable and flexible for its users.

driesvints:
From what I read, this should be done automatically by the client so I don't think there's anything we can do in Scout? 

https://www.algolia.com/doc/api-client/methods/advanced/#retry-logic

/cc @julienbourdeau

MatanYadaev:
@driesvints I'm talking about being able to add `$retries` and `$backoff` properties to `MakeSearchable` job.

driesvints:
@MatanYadaev talked this over with the team. Feel free to attempt a PR, thanks.

PR #476, opened 2021-06-01
[9.x] Overridable jobs

MatanYadaev:
This PR makes the package's jobs open for extension, for users that want to adjust the `$tries`, `$backoff`, `$maxExceptions` properties.

I implemented the same approach that implemented on Laravel Sanctum, providing the user the freedom the choose which job class will be used in the config file.

Discussed in #474.


config/scout.php
```diff
@@ -100,6 +100,20 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | @TODO add description
+    |
+    */
+
+    'jobs' => [
+        'make_searchable' => Laravel\Scout\Jobs\MakeSearchable::class,
+        'remove_from_search' => Laravel\Scout\Jobs\RemoveFromSearch::class,
+    ],
```
driesvints, line 115: Please avoid using the config file for this. All our first party libraries use a static property on a library specific class. We'd need to create a new one for Scout like: https://github.com/laravel/cashier-stripe/blob/master/src/Cashier.php#L69

driesvints:
Putting this in draft for now. Can you make the changes I requested?

MatanYadaev:
@driesvints Sure, thanks for the review and comment. It looks good now?
I wasn't sure if the tests are good or redundant, WDYT?

src/Scout.php
```diff
@@ -0,0 +1,23 @@
...
+     * @var string
+     */
+    public static $makeSearchableJob = MakeSearchable::class;
+
+    /**
+     * The remove from search job name.
+     *
+     * @var string
+     */
+    public static $removeFromSearchJob = RemoveFromSearch::class;
```
driesvints, line 22: Can you also add the static setter methods like in Cashier?

MatanYadaev: Done, thanks.

tests/Unit/MakeSearchableTest.php
```diff
@@ -24,4 +25,15 @@ public function test_handle_passes_the_collection_to_engine()
...
+    public function test_overridden_make_searchable_handle_passes_the_collection_to_engine()
+    {
+        $job = new OverriddenMakeSearchable($collection = Collection::make([
+            $model = m::mock(),
+        ]));
+
+        $model->shouldReceive('searchableUsing->update')->with($collection);
+
+        $job->handle();
+    }
```
driesvints, line 38: This test probably isn't necessary.

tests/Unit/RemoveFromSearchTest.php
```diff
@@ -32,6 +33,17 @@ public function test_handle_passes_the_collection_to_engine()
...
+    {
+        $job = new OverriddenRemoveFromSearch($collection = Collection::make([
+            $model = m::mock(),
+        ]));
+
+        $model->shouldReceive('searchableUsing->delete')->with($collection);
+
+        $job->handle();
+    }
+
```
driesvints, line 46: This test probably isn't necessary either.

driesvints:
If you fix the StyleCI warning and remove the two tests I highlighted I think we're good.

MatanYadaev:
@driesvints I didn't use StyleCI before. I need to register and configure an API key, right? Currently, the StyleCI website doesn't work for me so I can't do it locally.
There isn't a CI job for this?

driesvints:
You can't see https://github.styleci.io/analyses/wjbvPE?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link ?

MatanYadaev:
@driesvints Yes, the page doesn't work for me. Reported it to Graham on [Twitter](https://twitter.com/MatanYadaev/status/1399743034187784193).

driesvints (approved):

PR #471, opened 2021-05-11
Use queued job for "unsearching" when Scout queue is enabled

stevebauman:
**Description**:

Closes https://github.com/laravel/scout/issues/331

This PR adds a `RemoveFromSearch` queuable job which will be dispatched when `scout.queue` is enabled.

This specific queuable job has an overwritten `restoreCollection()` method that instantiates an Eloquent collection of the deleted models, by force-filling the model primary keys supplied by the `ModelIdentifier` object. This allows extended Scout engines to retrieve the ID used for indexing, so they can purge records properly.

All credit goes to @matt-allan, as this is a slightly modified copy of his work on his forked repo 🙌 

---

**References**:

Here are references to the most popular Scout Elasticsearch packages (they only use `getScoutKey()` for retrieving a unique identifier for the model):

https://github.com/babenkoivan/elastic-scout-driver/blob/2e45a933dcc8b051b25cbf8010623e51ac457a35/src/Factories/DocumentFactory.php#L24

https://github.com/babenkoivan/scout-elasticsearch-driver/blob/15d349a0401e2bab6f774255dc3668fd28b6d4a0/src/Indexers/BulkIndexer.php#L67

https://github.com/babenkoivan/scout-elasticsearch-driver/blob/15d349a0401e2bab6f774255dc3668fd28b6d4a0/src/Payloads/DocumentPayload.php#L19-L29

Issue #331, opened 2018-12-04
queueRemoveFromSearch() should optionally use a queue


chamby:
Engine delete operations, much like update operations, invoke the network and add latency. I would expect `Searchable::queueRemoveFromSearch()` to optionally use a queue just like `Searchable::queueMakeSearchable()` does. It even has queue in the name.

A new Job called `MakeUnsearchable` would also need to be created.

nunomaduro:
I kind of agree with you on this one. Possible code for a Pull Request:

```php
    /**
     * Dispatch the job to make the given models unsearchable.
     *
     * @param  \Illuminate\Database\Eloquent\Collection  $models
     * @return void
     */
    public function queueRemoveFromSearch($models)
    {
        if ($models->isEmpty()) {
            return;
        }

        if (! config('scout.queue')) {
            return $models->first()->searchableUsing()->delete($models);
        }

        dispatch((new RemoveFromSearch($models))
                ->onQueue($models->first()->syncWithSearchUsingQueue())
                ->onConnection($models->first()->syncWithSearchUsing()));
    }
```

@driesvints What do you think about this?

driesvints:
@nunomaduro looks good.

nunomaduro:
Thanks @driesvints. I gonna work on a Pull Request on the next couple days.

michaelnguyen547:
I just found out today that unsearchable() does not use queue. This PR will fix my issue. Thanks

mvdnbrk:
I am trying to implement this with the code suggested by @nunomaduro.

We will dispatch a job `RemoveFromSearch` passing `$models` which is an instance of  `\Illuminate\Database\Eloquent\Collection`, but by the time the job will be executed the models are already deleted from the database.

Any suggestions how to get around this?

jshayes:
I think that it why queueing deletes was initially removed in this commit https://github.com/laravel/scout/commit/f46b2b8a343162cf227afeb2dcd3709cb4bc626b

chamby:
Yes. This limitation makes sense. It might be possible to obtain the relevant keys for deletion using `$model->getScoutKey()` and queue the RemoveFromSearch Job with that data instead of attempting to de-serialize the model itself in the Job. The Scout keys are all that is actually needed issue a deletion request.

nathanblogs:
Any update on this, It was really confusing looking at the code to find out it doesn't actually queue.

jasonlfunk:
[...]

PR #479, opened 2021-06-10
Remove custom unserialize from RemoveFromSearch

Magnesium38:
The custom unserialization process in the `RemoveFromSearch` job does not take a customized scout key into account correctly.

Given a model with an overriden key like below that relies on other data on the model, the engines won't be able to process the models since they only have their primary key to work with.
```php
public function getScoutKey()
{
    return $this->custom_id;
}
```

In the case of the MeiliSearch engine, it will queue a request to delete a model with the key of `null`. This request continues on to MeiliSearch proper and it processes the request just fine, resulting in nothing being deleted. I'm not sure what the result would be on any other engine, but I would expect it to be similar.

The original PR that added this functionality added a test `test_models_are_deserialized_without_the_database`. In the process of creating this PR, I tried to think of ways to preserve that functionality by possibly customizing the serialization as well, or altering what the constructor does. I failed to come up with a clean and simple way to preserve that functionality. Every attempt I tried felt brittle.

Instead, I feel that this job being able to deserialize a model without the database is beyond the scope of what this package is aiming to provide, especially since the same feature is not available on the `MakeSearchable` job. If this avoiding the database for deserialization is required by an application, a custom version of the job could be used instead thanks to #476.

### DIFF
--- a/src/Jobs/RemoveFromSearch.php
+++ b/src/Jobs/RemoveFromSearch.php
@@ -4,7 +4,6 @@ namespace Laravel\Scout\Jobs;
 
 use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
-use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Illuminate\Queue\SerializesModels;
 
 class RemoveFromSearch implements ShouldQueue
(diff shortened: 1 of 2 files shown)


driesvints:
This will be a breaking change. This class was introduced in https://github.com/laravel/scout/pull/471. @stevebauman did you test this with both meilisearch and algolia when you sent this in?

stevebauman:
Hi @driesvints! I tested this feature on Algolia --- however, I did not test using a custom scout key. My apologies!

I'm going to see if I can add a test and resolve this in some way (cleanly). Give me a couple hours. I'll reply back whether or not I've come up with anything 👍

stevebauman:
Hi @Magnesium38 -- removing the custom unserialization from the job will break the job, as the models will already be deleted by the time the job executes (if the model isn't using soft deletes), effectively leaving your index with data that doesn't actually exist in your database.

See: https://github.com/laravel/scout/issues/331#issuecomment-457847513

Magnesium38:
Oooo, that's a solid reasoning behind needing to avoid the database. My typical use case doesn't delete data from the database, it's toggling keys that'll get used by `shouldBeSearchable`, which triggers the job that way.

Closing this since it is superseded by #480

PR #480, opened 2021-06-10
[9.x] Fix removing queued models with custom Scout keys

stevebauman:
## Purpose

This PR patches the issue described in PR https://github.com/laravel/scout/pull/479.

## Description

This PR fixes the issue described above. However, if developers are customizing their Scout key via `getScoutKey()` and are utilizing queuing, they must also modify the `getScoutKeyName()` to return the name of this custom key, so that the models can be hydrated and be properly removed from indexes. I can also submit a PR to the docs repo if this gets merged in so it can instruct developers to perform this step when queuing using custom scout keys.

The `RemoveableScoutCollection` class extends the Eloquent collection class, and pulls the queueable ID's from the model's `getScoutKey()` method. This is the cleanest solution I was able to come up with the least possible intrusion and work on the developers side.

In regards to the issue posted in the above PR, they would simply have to add this into their model after this patch:

```php
public function getScoutKey()
{
    return $this->custom_id;
}

{
    return $this->qualifyColumn('custom_id');
}
```

Let me know if there's any issues/comments/concerns with this approach. Thanks for your time! ❤️

### DIFF
--- a/src/Jobs/RemoveFromSearch.php
+++ b/src/Jobs/RemoveFromSearch.php
@@ -2,11 +2,11 @@
 
 namespace Laravel\Scout\Jobs;
 
-use Illuminate\Support\Str;
 use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Illuminate\Queue\SerializesModels;
+use Illuminate\Support\Str;
 
 class RemoveFromSearch implements ShouldQueue
 {
--- a/src/Jobs/RemoveFromSearch.php
+++ b/src/Jobs/RemoveFromSearch.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Scout\Jobs;
 
+use Illuminate\Support\Str;
 use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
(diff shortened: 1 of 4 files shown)


driesvints:
@stevebauman seems to break the test suite.

stevebauman:
@driesvints Okay we're all set! Fixed the test and style issues 👍

src/Jobs/RemoveFromSearch.php
```diff
@@ -14,7 +15,7 @@ class RemoveFromSearch implements ShouldQueue
     /**
      * The models to be removed from the search index.
      *
-     * @var \Illuminate\Database\Eloquent\Collection
+     * @var RemoveableScoutCollection
```
driesvints, line 18: Please revert this.

src/Jobs/RemoveFromSearch.php
```diff
@@ -56,9 +57,25 @@ protected function restoreCollection($value)
...
                 });
             })
         );
     }
+
+    /**
+     * Get the unqualified Scout key name.
+     *
+     * @param string $keyName
+     *
```
driesvints, line 74:
```suggestion
```

src/Jobs/RemoveFromSearch.php
stevebauman: Done!

taylorotwell:
So does this replace #479?

stevebauman:
> So does this replace #479?

Yup it does!

PR #483, opened 2021-06-14
[9.x] Re-query scout engine when paginate results contains insufficient keys to generate proper pagination count query

### LINKED ISSUE #481: [9.x] Pagination total messed up with query callback while eager loading
- Scout Version: 9.1.1
- Laravel Version: 8.42.1
- PHP Version: 8.0.5
- Database Driver & Version: MySQL 8.0.25

### Description:
After update from 8.x to 9.x, looks like #469 broke pagination total for me when using code like this:

```php
$qb = Article::search('*')->query(function ($query) {
    $query->with(['authors']);
})->paginate(10, 'page', 1);
```

It always returns `total` number same as `per_page` I passed as argument to paginate function.

I understand other users reported (#450 ) invalid total numbers when using query callback to filter results with `where`. Any suggestions how can I force to use `total` number from engine while still using eager loading in query callback?

EDIT:

Since for us it's asap and we don't use where`s in query callback, I created my own scout builder to fix this issue for now.

```php
<?php

namespace App\Support\Scout;

class Builder extends \Laravel\Scout\Builder
{
    /**
     * Get the total number of results from the Scout engine, or fallback to query builder.
     *
     * @param  mixed  $results
     * @return int
     */
    protected function getTotalCount($results)
    {
        $engine = $this->engine();

        return $engine->getTotalCount($results);
    }
}
```

and in `AppServiceProvider.php`

```php
$this->app->bind(\Laravel\Scout\Builder::class, \App\Support\Scout\Builder::class);
```

crynobone:
fixes #481

Signed-off-by: Mior Muhammad Zaki <crynobone@gmail.com>

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -372,12 +372,24 @@ class Builder
     {
         $engine = $this->engine();
 
+        $totalCount = $engine->getTotalCount($results);
+
         if (is_null($this->queryCallback)) {
-            return $engine->getTotalCount($results);
+            return $totalCount;
+        }
+
+        $ids = $engine->mapIds($results)->all();
+
+        if (count($ids) < $totalCount) {
+            $ids = $engine->keys(tap(clone $this, function ($builder) use ($totalCount) {
+                $builder->take(
+                    is_null($this->limit) ? $totalCount : min($this->limit, $totalCount)
+                );
+            }))->all();
         }
 
         return $this->model->queryScoutModelsByIds(
-            $this, $engine->mapIds($results)->all()
+            $this, $ids
         )->toBase()->getCountForPagination();
     }
 


src/Builder.php
```diff
@@ -372,12 +372,24 @@ protected function getTotalCount($results)
...
-            return $engine->getTotalCount($results);
+            return $totalCount;
+        }
+
+        $ids = $engine->mapIds($results)->all();
+
+        if (count($ids) < $totalCount) {
+            $ids = $engine->keys(tap(clone $this, function ($builder) use ($totalCount) {
+                $builder->take(
+                    is_null($this->limit) ? $totalCount : min($this->limit, $totalCount)
```
crynobone, line 386: setting the limit will prevent Scout Driver/Engine from setting a default limit. e.g: Meilisearch has 20 records limit by default.

PR #484, opened 2021-06-16
meilisearch engine AddDocuments with getScoutKeyName

mojtaba-p:
Scout has "getScoutKeyName" method that must be used for getting model key name instead of "getKeyName" because a model can has different primary key in meilisearch. for example in mysql user primary key can be "id" and in meilisearch can be "code"

### DIFF
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -65,7 +65,7 @@ class MeiliSearchEngine extends Engine
         })->filter()->values()->all();
 
         if (! empty($objects)) {
-            $index->addDocuments($objects, $models->first()->getKeyName());
+            $index->addDocuments($objects, $models->first()->getScoutKeyName());
         }
     }
 


mmachatschek:
@taylorotwell actually this is not that easy to fix and will break code of people that do not explicitly set the `->toSearchableArray()`

`->getScoutKeyName()` will return for e.g. a `Movie` model => `movies.id` instead of `id`.

`->toSearchableArray()` will not have `movies.id` in the array representation of the model. Therefore updates to models will not have any effect on the meilisearch objects.

The Algolia driver does`t have these kind of issues around `getScoutKeyName()` and `getScoutKey()` because they explicitly make use of the `objectID` that gets sent with every request.

So in my opinion this needs to be reverted

driesvints:
We reverted this. Thanks @mmachatschek

driesvints:
@mmachatschek I actually wonder why `getScoutKeyName` makes use of `getQualifiedKeyName` and not `getKeyname` 🤔

mmachatschek:
> @mmachatschek I actually wonder why `getScoutKeyName` makes use of `getQualifiedKeyName` and not `getKeyname`

me too. We already pointed this out to some contributors over in https://github.com/meilisearch/meilisearch-laravel-scout/pull/129 and https://github.com/meilisearch/meilisearch-laravel-scout/issues/126

I already had something in mind to make a change so that the MeiliSearch engine would replicate something like the Algolia engine where there is a specific identifier that will always pull the same information.

I haven't looked much around at other packages that extend scout, but I think they all have similar issue like this.

driesvints:
Guessing it could be if it could conflict with other relationship id's when you return a custom payload but unsure.

furmannodesol:
how to make a custom primary key for meilisearch indexes which is not a database field?

markhowell-works:
I hate to resurrect an old thread, but I still have not found a way to use a custom value for search index key that is different from the DB primary key. I'm trying to add multiple models to the same index and the lack of this feature is the only thing keeping it from working for me.

PR #488, opened 2021-06-28
Collection Engine

taylorotwell:
Over the weekend, I was hacking on something where I wanted to quickly use Scout without setting up any other external dependencies like Algolia or MeiliSearch.

I cooked up this quick driver that is intended for usage in **local development and small staging environments only**. It is not performant on large datasets and does not have robust fuzzy matching or anything. However, it is perfectly adequate in local development environments where you often don't have more than a couple hundred models.

However, this driver uses Eloquent's cursor method to retrieve filtered results and then uses `Str::contains` against the `LazyCollection` to match against all of the searchable columns. Note it is not possible to do those checks in the database because we do not know the data types of the columns and some RDBMS like Postgres will throw an error if you try to execute a `where like` clause against an integer column like an ID.

**So, in summary, this driver gives you a quick way to use Scout locally and get some generally sensible results from it just so you can continue building your application's search UI and functionality without worrying about the robustness of the search results themselves.**

I decided against calling this a `database` engine because it does not perform any sort of FULLTEXT column searching or use database search features. It iterates through a `LazyCollection` of all models of a given type and filters their column's values against the query within PHP.

### DIFF
--- a/src/EngineManager.php
+++ b/src/EngineManager.php
@@ -8,6 +8,7 @@ use Algolia\AlgoliaSearch\Support\UserAgent;
 use Exception;
 use Illuminate\Support\Manager;
 use Laravel\Scout\Engines\AlgoliaEngine;
+use Laravel\Scout\Engines\CollectionEngine;
 use Laravel\Scout\Engines\MeiliSearchEngine;
 use Laravel\Scout\Engines\NullEngine;
 use MeiliSearch\Client as MeiliSearch;
@@ -124,7 +125,17 @@ class EngineManager extends Manager
     }
 
     /**
-     * Create a Null engine instance.
+     * Create a database engine instance.
+     *
+     * @return \Laravel\Scout\Engines\CollectionEngine
+     */
+    public function createCollectionDriver()
+    {
+        return new CollectionEngine;
+    }
+
+    /**
+     * Create a null engine instance.
      *
      * @return \Laravel\Scout\Engines\NullEngine
      */
--- a/src/EngineManager.php
+++ b/src/EngineManager.php
@@ -125,7 +125,7 @@ class EngineManager extends Manager
     }
 
     /**
-     * Create a database engine instance.
+     * Create a collection engine instance.
      *
      * @return \Laravel\Scout\Engines\CollectionEngine
      */
--- a/src/Engines/CollectionEngine.php
+++ b/src/Engines/CollectionEngine.php
@@ -172,7 +172,7 @@ class CollectionEngine extends Engine
         $results = $results['results'];
 
         if (count($results) === 0) {
-            return LazyCollection::make($model->newCollection());
+            return LazyCollection::empty();
         }
 
         $objectIds = collect($results)
--- a/src/Engines/CollectionEngine.php
+++ b/src/Engines/CollectionEngine.php
@@ -83,12 +83,13 @@ class CollectionEngine extends Engine
     protected function searchModels(Builder $builder)
     {
         $models = $builder->model->query()
+                        ->when(count($builder->wheres) > 0, function ($query) use ($builder) {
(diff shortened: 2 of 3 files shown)


src/Engines/CollectionEngine.php
```diff
@@ -0,0 +1,239 @@
+<?php
+
+namespace Laravel\Scout\Engines;
+
+use Illuminate\Support\LazyCollection;
+use Illuminate\Support\Str;
+use Laravel\Scout\Builder;
+
+class CollectionEngine extends Engine
```
nunomaduro, line 9: What do you think about calling this `DatabaseCollectionEngine`? 💅🏻

src/EngineManager.php
```diff
@@ -124,7 +125,17 @@ protected function ensureMeiliSearchClientIsInstalled()
     }
 
     /**
-     * Create a Null engine instance.
+     * Create a database engine instance.
```
nunomaduro, line 128:
```suggestion
     * Create a collection engine instance.
```

src/Engines/CollectionEngine.php
```diff
@@ -0,0 +1,239 @@
...
+     * @param  mixed  $results
+     * @param  \Illuminate\Database\Eloquent\Model  $model
+     * @return \Illuminate\Support\LazyCollection
+     */
+    public function lazyMap(Builder $builder, $results, $model)
+    {
+        $results = $results['results'];
+
+        if (count($results) === 0) {
+            return LazyCollection::make($model->newCollection());
```
JosephSilber, line 175:
```suggestion
            return LazyCollection::empty();
```

src/Engines/CollectionEngine.php
```diff
@@ -0,0 +1,239 @@
...
+                ->pluck($model->getKeyName())
+                ->values()->all();
+
+        $objectIdPositions = array_flip($objectIds);
+
+        return $model->queryScoutModelsByIds(
+                $builder, $objectIds
+            )->cursor()->filter(function ($model) use ($objectIds) {
+                return in_array($model->getScoutKey(), $objectIds);
+            })->sortBy(function ($model) use ($objectIdPositions) {
```
JosephSilber, line 188: Sort, by definition, cannot be lazy.

This means that as soon as a single item is pulled out at the end of the chain, it'll have to pull all records into memory.

taylorotwell: So should I just switch back to `get` instead of `cursor`? I don't have big memory concerns here. This driver is only for local development.

JosephSilber: In this scenario, there isn't much practical difference between the two.

src/Engines/CollectionEngine.php
```diff
@@ -0,0 +1,239 @@
...
+        $models = $builder->model->query()
+                        ->when(count($builder->wheres) > 0, function ($query) use ($builder) {
+                            foreach ($builder->wheres as $key => $value) {
+                                $query->where($key, $value);
+                            }
+                        })
+                        ->orderBy($builder->model->getKeyName(), 'desc')
+                        ->get();
+
+        $models = $models->values();
```
MatanYadaev, line 94: Why `->values()` is necessary?

src/Engines/CollectionEngine.php
```diff
@@ -0,0 +1,239 @@
...
+
+        if (count($models) === 0) {
+            return $models;
+        }
+
+        $columns = array_keys($models->first()->toSearchableArray());
```
MatanYadaev, line 100: Not sure if it's relevant, but `toSearchableArray` might be dynamic and depend on the model record. It might be better to execute it per record, and not only based on the first record.

akulmehta:
Sorry for my naivety, but how does one go about using this? Is there documentation for this?

PR #489, opened 2021-06-30
[9.x] Update scout.php

ErickZH:
Change default driver to collection and sort alphabetically drivers


driesvints:
This is a major breaking change for anyone who hasn't set this env variable.

PR #490, opened 2021-07-01
[9.x] Improve observer strategy

ksntcrq:
# Problem

The current `ModelObserver` implementation calls `unsearchable` every time a model gets deleted. Practically, it means that the underlying engine (in our case, at StuDocu, Algolia) would be asked to delete the model from its index regardless of it actually being in the index or not.

# Solution

Two methods named `wasSearchableBeforeUpdate` and `wasSearchableBeforeDelete` are implemented to respectively check if a model was searchable before being updated or deleted. `ModelObserver` gets smarter and first checks if a model was actually searchable before calling `unsearchable`, thanks to one or the other method.

Now, the flow is as follow:

* When a model gets updated and is not searchable, we first check if it was searchable using `wasSearchableBeforeUpdate`. If it was, it is made unsearchable. If it wasn't, there is no need to make it unsearchable.
* When a model gets deleted, we first check if it was searchable using `wasSearchableBeforeDelete`. If it was, it is made unsearchable. If it wasn't, there is no need to make it unsearchable.

# Additional fix

The `ModelObserverTest` unit tests were not asserting anything and were actually broken after calling `disableSyncingFor` in `test_saved_handler_doesnt_make_model_searchable_when_disabled`. They're now fixed and extended to cover the use cases described above.

### DIFF
--- a/src/ModelObserver.php
+++ b/src/ModelObserver.php
@@ -79,7 +79,9 @@ class ModelObserver
         }
 
         if (! $model->shouldBeSearchable()) {
-            $model->unsearchable();
+            if ($model->wasSearchableBeforeUpdate()) {
+                $model->unsearchable();
+            }
 
             return;
         }
@@ -96,6 +98,10 @@ class ModelObserver
     public function deleted($model)
     {
         if (static::syncingDisabledFor($model)) {
+            return;
+        }
+
+        if (! $model->wasSearchableBeforeDelete()) {
             return;
         }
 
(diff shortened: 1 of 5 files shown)


taylorotwell:
I changed the `wasSearchable...` methods to always return `true`. You can override them to provide your own opinions on how they should work.

ksntcrq:
Thank you, @taylorotwell.

I was wondering - will you guys include the suggested `wasSearchable...` implementations in the documentation as an example, so users won't have to figure it out themselves?

PR #492, opened 2021-07-02
[9.x] Add Searchable contract to complement trait

wimski:
In a recent project using Scout I wanted to check if a class uses the `Searchable` trait. Currently this can only be achieved by one of the following options:

```php
// A. check trait usage
if (! in_array(Searchable::class, class_uses($model)) {
    throw new Exception();
}

// B. check method existence
if (! method_exists($model, 'searchableAs')) {
    throw new Exception();
}
```

* Option A will work, but static analysis will not understand that `$model` has a specific set of methods.
* Option B does support static analysis, but is messy, because it doesn't really check for the trait (or specific contract).

Therefore I've added a contract to complement the trait.

```php
use Laravel\Scout\Contracts\Searchable as SearchableContract;

if (! $model instanceof SearchableContract::class) {
    throw new Exception();
}
```

This contract is not required for general use of Scout and is not a breaking change.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -14,7 +14,7 @@ class Builder
     /**
      * The model instance.
      *
-     * @var \Illuminate\Database\Eloquent\Model
+     * @var \Illuminate\Database\Eloquent\Model|\Laravel\Scout\Contracts\Searchable
      */
     public $model;
 
--- a/src/Contracts/Searchable.php
+++ b/src/Contracts/Searchable.php
@@ -1,0 +1,193 @@
+<?php
+
+namespace Laravel\Scout\Contracts;
+
+interface Searchable
+{
+    /**
+     * Boot the trait.
+     *
+     * @return void
+     */
+    public static function bootSearchable();
+
+    /**
+     * Register the searchable macros.
+     *
+     * @return void
+     */
+    public function registerSearchableMacros();
+
+    /**
+     * Dispatch the job to make the given models searchable.
+     *
+     * @param  \Illuminate\Database\Eloquent\Collection  $models
+     * @return void
+     */
+    public function queueMakeSearchable($models);
+
+    /**
+     * Dispatch the job to make the given models unsearchable.
+     *
+     * @param  \Illuminate\Database\Eloquent\Collection  $models
+     * @return void
+     */
+    public function queueRemoveFromSearch($models);
+
+    /**
+     * Determine if the model should be searchable.
+     *
+     * @return bool
+     */
+    public function shouldBeSearchable();
+
+    /**
+     * Perform a search against the model's indexed data.
+     *
+     * @param  string  $query
+     * @param  \Closure  $callback
+     * @return \Laravel\Scout\Builder
+     */
+    public static function search($query = '', $callback = null);
+
+    /**
+     * Make all instances of the model searchable.
+     *
+     * @param  int  $chunk
+     * @return void
+     */
+    public static function makeAllSearchable($chunk = null);
+
+    /**
+     * Make the given model instance searchable.
+     *
+     * @return void
+     */
+    public function searchable();
+
+    /**
+     * Remove all instances of the model from the search index.
+     *
+     * @return void
+     */
+    public static function removeAllFromSearch();
+
+    /**
+     * Remove the given model instance from the search index.
+     *
+     * @return void
+     */
+    public function unsearchable();
+
+    /**
+     * Get the requested models from an array of object IDs.
+     *
+     * @param  \Laravel\Scout\Builder  $builder
+     * @param  array  $ids
+     * @return mixed
+     */
+    public function getScoutModelsByIds(\Laravel\Scout\Builder $builder, array $ids);
+
+    /**
+     * Get a query builder for retrieving the requested models from an array of object IDs.
+     *
+     * @param  \Laravel\Scout\Builder  $builder
+     * @param  array  $ids
+     * @return mixed
+     */
+    public function queryScoutModelsByIds(\Laravel\Scout\Builder $builder, array $ids);
+
+    /**
+     * Enable search syncing for this model.
+     *
+     * @return void
+     */
+    public static function enableSearchSyncing();
+
+    /**
+     * Disable search syncing for this model.
+     *
+     * @return void
+     */
+    public static function disableSearchSyncing();
+
+    /**
+     * Temporarily disable search syncing for the given callback.
+     *
+     * @param  callable  $callback
+     * @return mixed
+     */
+    public static function withoutSyncingToSearch($callback);
+
+    /**
+     * Get the index name for the model.
+     *
+     * @return string
+     */
+    public function searchableAs();
+
+    /**
+     * Get the indexable data array for the model.
+     *
+     * @return array
+     */
+    public function toSearchableArray();
+
+    /**
+     * Get the Scout engine for the model.
+     *
+     * @return mixed
+     */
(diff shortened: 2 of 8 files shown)


driesvints:
I don't really see the use of a contract when it's never used anywhere.

denniskoster:
> I don't really see the use of a contract when it's never used anywhere.

The use would be to be able to easily signal searchable models in code outside of the package itself. It allows for stricter typing.

taylorotwell:
Can you elaborate on the static analysis part? Can it not see that your model uses the `Searchable` trait?

wimski:
> Can you elaborate on the static analysis part? Can it not see that your model uses the `Searchable` trait?

Sure, when you use a specific model there is no issue, but when you have an unspecified model (which you know should be searchable), there's know way to for the code to know. And now way to tell the code. An interface fixes that.

```php
/**
 * @param Model $model
 */
public function doesSomeSearchThing($model)
{
   // We know the $model should have the Searchable trait
   // but this can't be inferred because this is a generic model
    if ($model->shouldBeSearchable()) {
        // do something
    }
}

/**
 * @param Model|Searchable $model
 */
public function anotherSearchMethod($model)
{
    // Here the search method can be inferred
    // because of the interface
    if ($model->shouldBeSearchable()) {
        // do something
    }
}
```

crynobone:
That's can be done via following:

```php
use Laravel\Scout\Searchable;

/**
 * @param Model $model
 */
public function doesSomeSearchThing($model)
{
   // We know the $model should have the Searchable trait
   // but this can't be inferred because this is a generic model
    if (in_array(Searchable::class, class_uses_recursive($model))) {
        // do something
    }
}
```

Adding an interface would mean that in order to use `$model instanceof SearchableContract`, each model has to implements `SearchableContract` and uses `Searchable`.

wimski:
@crynobone That's the option A in my original post. Even-though this checks if the trait is present to prevent missing method call errors at runtime, static analysis will be none the wiser. This blocks auto-completion/click-through in you IDE and things like PHPStan will (rightly so) report issues.

The point isn't to check if a class uses a certain trait, it's that you can reliable know that the class has a specific set of public methods whether they come from a trait or not. An interface helps to define that 'contract'.

driesvints:
@wimski you can still add this interface to your own app. I don't see the use case of defining it in Scout when it's not used anywhere.

wimski:
@driesvints That's what I initially did, but it also means I have to include that interface in everything project that uses Scout. And I have to keep it up-to-date with trait changes in every project. It makes a lot more sense to have it maintained and shipped alongside the trait.

I've added a commit that uses the interface in annotations which helps static analysis. Other pieces of code could benefit from the interface as well, but that would introduce breaking changes which is a path I'm not willing to go down right now seeing as introducing a simple interface is already met with so much resistance.

## Example of breaking changes omitted from the PR

### `Laravel\Scout\Searchable`
```php
/**
 * Dispatch the job to make the given models searchable.
 *
 * @param  \Illuminate\Database\Eloquent\Collection  $models
 * @return void
 */
public function queueMakeSearchable($models)
{
    if ($models->isEmpty()) {
        return;
    }

    $model = $models->first();

    // This interface check ensures all the following
    // searchable methods are available,
    // also to static analysis and your IDE
    if (! $model instanceof SearchableContract) {
        return;
    }

    if (! config('scout.queue')) {
        return $model->searchableUsing()->update($models);
    }

    dispatch((new Scout::$makeSearchableJob($models))
            ->onQueue($model->syncWithSearchUsingQueue())
            ->onConnection($model->syncWithSearchUsing()));
}

/**
 * Dispatch the job to make the given models unsearchable.
 *
 * @param  \Illuminate\Database\Eloquent\Collection  $models
 * @return void
 */
public function queueRemoveFromSearch($models)
{
    if ($models->isEmpty()) {
        return;
    }

    $model = $models->first();

    // This interface check ensures all the following
    // searchable methods are available,
    // also to static analysis and your IDE
    if (! $model instanceof SearchableContract) {
        return;
    }

    if (! config('scout.queue')) {
        return $model->searchableUsing()->delete($models);
    }

    dispatch(new Scout::$removeFromSearchJob($models))
        ->onQueue($model->syncWithSearchUsingQueue())
        ->onConnection($model->syncWithSearchUsing());
}
```

taylorotwell:
Do we actually need all of these methods on the interface? Would it be better to include the most common methods only?

denniskoster:
@taylorotwell I think it's a good idea to have all the available public methods in that interface. Considering that the point is to help the IDE and static analysis tools understand what an instance of a class that implement this interface can do, I don't think it would be wise to leave some methods out. That would result in a situation where calling public methods that are *not* included in the contract would still cause the IDE and static analysis tools to complain.

nunomaduro:
I agree with @driesvints. I don't see the need for this.

denniskoster:
You're the guys who have to maintain it in the end so fair play in that regard. I just wonder if someone can elaborate why there seems to be such an aversion to introducing an interface, that doesn't break anything and you don't _have_ to use if you don't want to. It adds value for those that care about strict typing and adding it has no downsides as far as I can tell.

Sure, everyone can just create this interface themselves if they want to but having it be provided by the package itself is something I always appreciate personally. One example of this is [spatie/eloquent-sortable](https://github.com/spatie/eloquent-sortable/tree/master/src) though, granted, they do not include *all* of the public methods in the provided interface.

driesvints:
@denniskoster I'm sorry but I already told you. The interface wouldn't be implemented anywhere in the library. It would effectively be unused. It only makes sense if you're ever going to swap out Scout for a different search library, something you'll never do.

denniskoster:
@driesvints Thanks Dries. 

I think there are more valid reasons for wanting an interface than just "wanting to swap out an entire search library", as perfectly demonstrated by @wimski in code examples above. Besides the benefits for the static analysis tools as mentioned by Wim, an interface will also greatly help your IDE and therefor the user. 

I realize that I will probably not convince you of the added benefit of having this interface but I would still like to propose the following example, hoping that it might explain better what I am trying to say.

```php
<?php

namespace App\Processors;

use App\Contracts\SomeTraitInterface;
use App\Traits\SomeTrait;

class SomeProcessor
{
    public function processWithoutInterface($model): int
    {
        if (! in_array(SomeTrait::class, class_uses($model))) {
            throw new \InvalidArgumentException('Provided class should use ' . SomeTrait::class);
        }

        if (! method_exists($model, 'reduce')) {
            throw new \InvalidArgumentException('Provided class should have a reduce method');
        }

        // Even though we have checked the provided model has the provided trait
        // and even explicitly check that the model has a 'reduce' method, the IDE
        // will still not understand where to find this reduce method or what its
        // parameters are.
        return $model->reduce(15, 2);
    }

    public function processWithInterface(SomeTraitInterface $model): int
    {
        // The 'reduce' method is defined in the interface, therefore the IDE understands
        // what the reduce method accepts as parameters as well as where to find this
        // method. Additionally, PHP will not let me pass any object that does not implement
        // this interface to this method, removing the need for custom checks and custom
        // exceptions, resulting in a much cleaner and more readable method body.
        return $model->reduce(15, 2);
    }
}
```

Thanks for your time and patience!

nunomaduro:
I think we all understand that use case, @denniskoster. Thank you for explaining. Yet, the question here not benefit, but rather if the trait should be at package level or at application level (created by all users who have this need).

taylorotwell:
I think for now I will just let developers maintain their own interface. I don't want to take on the maintenance burden.

PR #493, opened 2021-07-05
Downcase attribute and query for case-insensitive search

dwightwatson:
A minor change to the collection engine that allows case-insensitive search by simply downcasing the attribute and query arguments before calling `contains`. Using the test data at the moment a search for a lowercase `otwell` wouldn't match either of you because your names are capitalised.


driesvints:
This isn't always wanted I think. A real-life situation of a search engine involves case sensitivity. For example: https://www.algolia.com/doc/guides/managing-results/refine-results/faceting/#case-sensitivity

We could indeed make an exception for the collection engine since it's more meant as a testing playground.

Issue #496, opened 2021-07-06
Add an ability to retrieve hit metadata for MeiliSearch like for Algolia for model instances


siarheipashkevich:
Hi all.

What do you think about an ability for retrieving hit metadata for MeiliSearch from response like for Algolia [Retrieve hit metadata](https://www.algolia.com/doc/framework-integration/laravel/searching/server-side-search/?client=php#retrieve-hit-metadata)?

It is needed for retrieving the result of the search from MeiliSearch as [Attributes to highlight](https://docs.meilisearch.com/reference/features/search_parameters.html#attributes-to-highlight) or something else for instances of the searchable models.

Currently we need to manually write this logic:
1. get raw search results from MeiliSearch
2. get models by ids from search result from db
3. map needed properties from search result (for highlight attributes we have `_formatted` property in raw data) to our instances

Thanks.

driesvints:
Yes, that sounds reasonable. Feel free to attempt a PR.

PR #497, opened 2021-07-06
[9.x] Added an ability to retrieve hit metadata for MeiliSearch engine

siarheipashkevich:
Hi all.

What do you think about an ability for retrieving hit metadata for MeiliSearch from response like for Algolia Retrieve hit metadata?

It is needed for retrieving the result of the search from MeiliSearch as Attributes to highlight or something else for instances of the searchable models.

Currently we need to manually write this logic:

get raw search results from MeiliSearch
get models by ids from search result from db
map needed properties from search result (for highlight attributes we have _formatted property in raw data) to our instances
Thanks.

https://github.com/laravel/scout/issues/496

### DIFF
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -241,7 +241,7 @@ class MeiliSearchEngine extends Engine
 
         $hits = collect($results['hits'])->keyBy($model->getKeyName());
 
-        $objectIds = collect($results['hits'])->pluck($model->getKeyName())->values()->all();
+        $objectIds = $hits->pluck($model->getKeyName())->values()->all();
         $objectIdPositions = array_flip($objectIds);
 
         return $model->queryScoutModelsByIds(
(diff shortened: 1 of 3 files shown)


src/Engines/MeiliSearchEngine.php
```diff
@@ -201,17 +209,22 @@ public function map(Builder $builder, $results, $model)
...
+            ->filter(function ($model) use ($objectIds) {
+                return in_array($model->getScoutKey(), $objectIds);
+            })
+            ->each(function ($model) use ($hits) {
+                $this->setHitMetadata($model, $hits);
+            })
+            ->sortBy(function ($model) use ($objectIdPositions) {
+                return $objectIdPositions[$model->getScoutKey()];
+            })
+            ->values();
```
driesvints, line 227: Can you undo the style changes you made here so it's more easier to parse the diff?

taylorotwell:

PR #503, opened 2021-07-30
Leverage Laravel Unique Job

ngunyimacharia:
# Added

- ShouldBeUnique interface and uniqueId() method to MakeSearchable.php


taylorotwell:
The PR doesn't include a description of why this would be necessary.

ngunyimacharia:
The purpose of the pull request was to prevent duplicate model updates. I will re-do using a better approach and send another PR this coming week. Thank you for the feedback, will also add a comprehensive description.

PR #509, opened 2021-08-23
[10.x] Refactor the use of getScoutKeyName

driesvints:
This PR changes the use of getScoutKeyName. The most significant change is the removal of the use of the qualified column name in the `getScoutKeyName` method. Other changes are using the `getScoutKeyName` instead of `getKeyName` so people can modify what column is used to index search records. This was already possible with Algolia but not with Meilisearch.

The reason why the qualified column was removed was that if someone were to override `getScoutKeyName` (with the other changes in mind from this PR) the `toSearchableArray` wouldn't be compatible anymore as it won't append the correct column name and thus search records in Meilisearch won't index properly anymore. If we want to make `getScoutKeyName` adjustable for Meilisearch as well (as currently stated in the docs) then this change is necessary. As pointed out by @mmachatschek in https://github.com/laravel/scout/pull/484#issuecomment-863189618, Algolia doesn't suffer from this as it's using its own internal objectID mapping. 

The upgrade path for this unfortunately is that people will have to re-index their search records. If that's not wanted then we should re-consider removing the ability to override the key name. 

I'm also not entirely sure what the original reason for the use of the qualified column name was. I do not know if there will be a breaking change because if this.

I'd appreciate a review from you @mmachatschek before I mark this as ready to review for Taylor.

### DIFF
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -144,7 +144,7 @@ trait Searchable
             ->when($softDelete, function ($query) {
                 $query->withTrashed();
             })
-            ->orderBy($self->getKeyName())
+            ->orderBy($self->getScoutKeyName())
             ->searchable($chunk);
     }
 
@@ -382,7 +382,7 @@ trait Searchable
      */
     public function getScoutKeyName()
     {
-        return $this->getQualifiedKeyName();
+        return $this->getKeyName();
     }
 
     /**
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -144,7 +144,9 @@ trait Searchable
             ->when($softDelete, function ($query) {
                 $query->withTrashed();
             })
-            ->orderBy($self->getScoutKeyName())
+            ->orderBy(
+                $self->qualifyColumn($self->getScoutKeyName())
+            )
             ->searchable($chunk);
     }
 
@@ -240,7 +242,7 @@ trait Searchable
         }
 
         return $query->whereIn(
-            $this->getScoutKeyName(), $ids
+            $this->qualifyColumn($this->getScoutKeyName()), $ids
         );
     }
 
--- a/src/Engines/CollectionEngine.php
+++ b/src/Engines/CollectionEngine.php
@@ -100,7 +100,9 @@ class CollectionEngine extends Engine
                                 $query->whereIn($key, $values);
                             }
                         })
-                        ->orderBy($builder->model->getScoutKeyName(), 'desc');
+                        ->orderBy(
+                            $builder->model->qualifyColumn($builder->model->getScoutKeyName()), 'desc'
+                        );
 
         $models = $this->ensureSoftDeletesAreHandled($builder, $query)
                         ->get()
--- a/src/Engines/CollectionEngine.php
+++ b/src/Engines/CollectionEngine.php
@@ -100,7 +100,7 @@ class CollectionEngine extends Engine
                                 $query->whereIn($key, $values);
                             }
                         })
-                        ->orderBy($builder->model->getKeyName(), 'desc');
+                        ->orderBy($builder->model->getScoutKeyName(), 'desc');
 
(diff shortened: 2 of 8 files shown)


mmachatschek (changes requested):
@driesvints 

please update the `\Laravel\Scout\Jobs\RemoveFromSearch::restoreCollection()` method and remove the "fix" to get the unqualified key name.

also update the `\Laravel\Scout\Tests\Fixtures\SearchableModelWithCustomKey::getScoutKeyName()` method.

mmachatschek (changes requested):
@driesvints 

When changing `Searchable::getScoutKeyName()` the method `Searchable::queryScoutModelsByIds()` should be updated so that the key is uses a qualified column name? What do you think?

```diff
    public function queryScoutModelsByIds(Builder $builder, array $ids)
    {
        $query = static::usesSoftDelete()
            ? $this->withTrashed() : $this->newQuery();

        if ($builder->queryCallback) {
            call_user_func($builder->queryCallback, $query);
        }

        return $query->whereIn(
-            $this->getScoutKeyName(), $ids
+            $this->qualifyColumn($this->getScoutKeyName()), $ids
        );
    }
```

driesvints:
@mmachatschek did some of those changes.

> When changing Searchable::getScoutKeyName() the method Searchable::queryScoutModelsByIds() should be updated so that the key is uses a qualified column name? What do you think?

Should we also do this for the `makeAllSearchable`'s `orderBy` statement?

```diff
    public static function makeAllSearchable($chunk = null)
    {
        $self = new static;

        $softDelete = static::usesSoftDelete() && config('scout.soft_delete', false);

        $self->newQuery()
            ->when(true, function ($query) use ($self) {
                $self->makeAllSearchableUsing($query);
            })
            ->when($softDelete, function ($query) {
                $query->withTrashed();
            })
-            ->orderBy($self->getScoutKeyName())
+            ->orderBy($this->qualifyColumn($self->getScoutKeyName()))
            ->searchable($chunk);
    }
```

src/Searchable.php
```diff
@@ -144,7 +144,7 @@ public static function makeAllSearchable($chunk = null)
             ->when($softDelete, function ($query) {
                 $query->withTrashed();
             })
-            ->orderBy($self->getKeyName())
+            ->orderBy($self->getScoutKeyName())
```
mmachatschek, line 147:
```suggestion
            ->orderBy($self->qualifyColumn($self->getScoutKeyName()))
```

mmachatschek (changes requested):
@driesvints Yes.

Both `Searchable::queryScoutModelsByIds()` and `Searchable::makeAllSearchable()` should have the qualifiedColumn call IMO.

Additionally after the changes with the `qualifiedColumn` name the Algolia driver should have the same behaviour as before the changes.

driesvints:
> Both Searchable::queryScoutModelsByIds() and Searchable::makeAllSearchable() should have the qualifiedColumn call IMO.

@mmachatschek done, can you check again?

> Additionally after the changes with the qualifiedColumn call the Algolia driver should have the same behaviour as before the changes.

How do we practically do this?

mmachatschek (approved):
> How do we practically do this?

Nothing to do here. Just wanted to point out that for algolia everything should be working as before now that you made the changes.

LGTM

driesvints:
Okay, this is ready for review now. I tried out the changes on both the Laravel.io and blade-ui-kit.com codebases without problems. @taylorotwell feel free to give this a review.

mmachatschek:
@driesvints MeiliSearch v0.21.0 was released yesterday. It brings breaking changes on the library level.
I'm not sure if there will be the 10.x release after merging this PR, but maybe lets wait for the update of the meilisearch/meilisearch-php library. The release of it should be out this week maybe even today.

rosszt:
Fantastic, this is a great improvement for me. I had this exact issue this week 👍

PR #510, opened 2021-08-24
[9.x] Support rename of filters to filter in meilisearch 0.21.x

georgeboot:
Meilisearch 0.21.0 [contains a breaking change](https://github.com/meilisearch/MeiliSearch/releases/tag/v0.21.0) regarding the `filters` attribute. It was renamed to `filter` instead.
This PR addresses that by checking what version of the SDK was installed, and handling the attribute accordingly.

Not that user would need to manually add the attributes they want to filter on to the `filterableAttributes` setting, they can not be used on any field any more as one could before version 0.21.

### DIFF
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -127,9 +127,9 @@ class MeiliSearchEngine extends Engine
     {
         $meilisearch = $this->meilisearch->index($builder->index ?: $builder->model->searchableAs());
 
-        // from 0.21.0 onwards, `filters` is renamed to `filter`
         if (version_compare(MeiliSearch::VERSION, '0.21.0') >= 0) {
             $searchParams['filter'] = $searchParams['filters'];
+
             unset($searchParams['filters']);
         }
 
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Scout\Engines;
 
+use Composer\InstalledVersions;
 use Illuminate\Support\LazyCollection;
 use Laravel\Scout\Builder;
 use MeiliSearch\Client as MeiliSearch;
@@ -125,6 +126,12 @@ class MeiliSearchEngine extends Engine
     protected function performSearch(Builder $builder, array $searchParams = [])
     {
         $meilisearch = $this->meilisearch->index($builder->index ?: $builder->model->searchableAs());
+
+        // from 0.21.0 onwards, `filters` is renamed to `filter`
+        if (version_compare($this->getInstalledMeilisearchVersion(), '0.21.0') >= 0) {
+            $searchParams['filter'] = $searchParams['filters'];
+            unset($searchParams['filters']);
+        }
 
         if ($builder->callback) {
             $result = call_user_func(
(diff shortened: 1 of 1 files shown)


src/Engines/MeiliSearchEngine.php
```diff
@@ -309,4 +316,18 @@ public function __call($method, $parameters)
...
+
+    /**
+     * Get the version of the meilisearch/meilisearch-php package intalled by Composer.
+     *
+     * @return string|null
+     */
+    protected function getInstalledMeilisearchVersion()
+    {
+        try {
+            return InstalledVersions::getVersion('meilisearch/meilisearch-php');
```
driesvints, line 328: This won't work for users still on Composer v1.0

georgeboot: Any recommendations on how to support 1.x as well, other than reading raw `composer.lock`?

driesvints: None I'm afraid...

crynobone: You can to read it from https://github.com/meilisearch/meilisearch-php/blob/main/src/MeiliSearch.php#L9

src/Engines/MeiliSearchEngine.php
```diff
@@ -126,6 +127,12 @@ protected function performSearch(Builder $builder, array $searchParams = [])
     {
         $meilisearch = $this->meilisearch->index($builder->index ?: $builder->model->searchableAs());
 
+        // from 0.21.0 onwards, `filters` is renamed to `filter`
+        if (version_compare($this->getInstalledMeilisearchVersion(), '0.21.0') >= 0) {
+            $searchParams['filter'] = $searchParams['filters'];
+            unset($searchParams['filters']);
+        }
```
driesvints, line 134: Instead of doing this can't we just to the following?

```suggestion
        $searchParams['filter'] = $searchParams['filters'] ?? $searchParams['filter'];
        
        unset($searchParams['filters']);
```

src/Engines/MeiliSearchEngine.php
georgeboot: Oh wow completely missed that. Will update.

src/Engines/MeiliSearchEngine.php
driesvints: I'm not sure if it's a bad thing that the `filters` key stays in place when using an older Meilisearch version?

georgeboot: This is the inverse, right? This will now always send `filter` but you might still be running meilisearch >=0.20.0 and that version still expects `filters`.

driesvints: But the `filters` key will stay. Only the `filter` key will be set as well. Does it matter if a `filter` key is set on newer versions?

georgeboot: Unfortunately they do a strict scheme check, so sending both `filter` and `filters` to a 0.21 will error because `filters` is not recognised.

driesvints: Right, gotcha. Thanks for clarifying that 👍

src/Engines/MeiliSearchEngine.php
```diff
@@ -126,6 +126,12 @@ protected function performSearch(Builder $builder, array $searchParams = [])
     {
         $meilisearch = $this->meilisearch->index($builder->index ?: $builder->model->searchableAs());
 
+        // from 0.21.0 onwards, `filters` is renamed to `filter`
+        if (version_compare($this->getInstalledMeilisearchVersion(), '0.21.0') >= 0) {
```
driesvints, line 130:
```suggestion
        if (version_compare(MeiliSearch::VERSION, '0.21.0') >= 0) {
```

And rename the current `MeiliSearch` alias to `MeiliSearchClient`.

georgeboot:
Only issue I can still see, is that we now check for the installed version of the SDK, not for the actual server version (which is technically more important). It is thus up to the user to match (major) versions.

The only way to check actual server version is by doing an API call, which isn't really ideal.

driesvints:
@georgeboot yeah indeed. But I think that's up to the user to match or up to the SDK, not for a third party library like Scout.

PR #524, opened 2021-09-18
[meilisearch] support boolean filters

G3z:


G3z:
if a value is indexed as boolean it's now possible to use `->where('active', true)`:
this produces a meili query as follows `--data '{"filter": ["active = true"]}'`

please note that `--data '{"filter": ["active = 1"]}' does not work

limitations: `->where('active', 1)` would still not work (maybe we should take a look at `$casts` parameter)

mgralikowski:
Hah, I faced the same problem today.

It is so strange for me to write sth like:
```
$query->where('is_active', 'true');
```
because only this works, even is_active is a boolean value.

PR #525, opened 2021-09-20
Update MeiliSearchEngine.php

mgralikowski:
For now, there is no option to write a query with multiple conditions on the same attribute. Of course for simple attributes like "gender" it does not make sense to make a query like `gender = M and gender = W` but for array, it does though.

Let's take an example: we want to find only entries with tag id = 1 **and** id = 2. By using whereIn we do not get correct results because conditions in whereIn are connected by **or** so we get entries with id 1 **or** 2, not with both. The second idea is to use the standard where() multiple times.
```
$query->where('tag', 1);
$query->where('tag', 2);
```
It looks logical. Wheres are joined by **and** but it doesn't work. Why? Because where method uses an associative array so we are limited to one where() condition per attribute.

So my PR introduces a new possibility:
```
$query->where('tag', [1, 2]); // find entries with tag id 1 and with tag id 2
$query->whereIn('tag', [1, 2]); // find entries with tag 1 or 2 (notching changed).
```

Also, my PR contains a small refactor to avoid redundancy (DRY).

### DIFF
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -149,6 +149,20 @@ class MeiliSearchEngine extends Engine
     }
 
     /**
+     * Convert attribute name and value to a string representation of condition.
+     *
+     * @param string $key
+     * @param mixed  $value
+     * @return string
+     */
+    private function filterExpression(string $key, $value): string
+    {
+        return filter_var($value, FILTER_VALIDATE_INT) !== false
+            ? sprintf('%s=%s', $key, $value)
+            : sprintf('%s="%s"', $key, $value);
+    }
+
+    /**
      * Get the filter array for the query.
      *
      * @param  \Laravel\Scout\Builder  $builder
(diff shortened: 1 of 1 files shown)


taylorotwell:
I guess I'm confused - why does an entity have a tag ID of 1 AND 2?

taylorotwell:
To be honest I'm not totally sure how this fixes your problem. Your `tags` field is an array, right? How does checking tags = foo AND tags = bar help you. Don't you want to check that thats CONTAINS foo and bar at the same time?

mgralikowski:
Yes, `tags` is an array.  `['foo', 'bar', 'baz']`. Yes, I want to find resources that contain `foo` and `bar` at the same time. 
```
$query->where('is_public', true); // standard condition
$query->whereIn('language', ['en', 'pl']); // in en or in pl
$query->where('tags', ['foo', 'bar']); // both
```
The final string query to the Meilisearch API should looks like this:
```
'filter' => 'is_public="true" AND (language="pl" OR language="en") AND tags="foo" AND tags="bar"'
```

taylorotwell:

PR #528, opened 2021-09-24
[9.x] - Collection Engine: add support for non-scalar values

macbookandrew:
The Algolia and MeiliSearch engines both allow searchable data to be non-scalar values (e.g., an [array of ingredients](https://www.algolia.com/doc/guides/sending-and-managing-data/prepare-your-data/how-to/setting-searchable-attributes/) or [genres](https://docs.meilisearch.com/reference/features/filtering_and_faceted_search.html#example-2)).

The Collection engine, however, expects all values to be scalar, and during search, throws a `TypeError` if a model’s `toSearchableArray()` returns any non-scalar key.

```
TypeError: mb_strtolower(): Argument #1 ($string) must be of type string, array given in /path/to/project/vendor/laravel/framework/src/Illuminate/Support/Str.php:361
Stack trace:
#0 /path/to/project/vendor/laravel/framework/src/Illuminate/Support/Str.php(361): mb_strtolower(Array, 'UTF-8')
// etc.
```

This PR is a stupid-simple fix that just JSON-encodes non-scalar values on the theory that it will result in a string that `Str::contains()` will be able to inspect for the search term.

### DIFF
--- a/src/Engines/CollectionEngine.php
+++ b/src/Engines/CollectionEngine.php
@@ -122,6 +122,9 @@ class CollectionEngine extends Engine
             $searchables = $model->toSearchableArray();
 
             foreach ($searchables as $value) {
+                if (! is_scalar($value)) {
+                    $value = json_encode($value);
+                }
                 if (Str::contains(Str::lower($value), Str::lower($builder->query))) {
                     return true;
                 }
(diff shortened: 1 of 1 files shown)


macbookandrew:
The biggest benefit I see is being able to use the Collection engine in tests that depend on Scout search results (i.e., testing the content of a page rendered by a search controller).

Issue #529, opened 2021-09-27
support operators in filter


G3z:
Both algolia and meilisearch supports operatators other than `=` when using filters

` <, <=, =, !=, >=, > ` are available for both engine

[Algolia docs](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/how-to/filter-by-numeric-value/#applying-a-numeric-filter)
[MeiliSearch docs](https://docs.meilisearch.com/reference/features/filtering_and_faceted_search.html#using-filters)

is it possible to support this ?

from what i understand:
`Laravel\Scout\Builder::where` has to accept 3 parameters instead of 2
and
`Laravel\Scout\Engines\AlgoliaEngine::filters` and `Laravel\Scout\Engines\MeiliSearchEngine::filters` to apply the operator

if you agree I can prepare a PR

driesvints:
Yeah that sounds interesting, definitely. Probably best to send to master for breaking changes. If you can and preserve the current behavior as much as possible to minimize the breaking changes that'd be great. Thanks!

PR #531, opened 2021-09-28
support operators in filters

G3z:
TL;DR;
Support for query operators like `->where('foo','>','1')` short (old) version still works `->where('foo','1')`

Both Algolia and MeiliSearch supports different operators when filtering:
` <, <=, =, !=, >=, > ` are available for both engine

[Algolia docs](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/how-to/filter-by-numeric-value/#applying-a-numeric-filter)
[MeiliSearch docs](https://docs.meilisearch.com/reference/features/filtering_and_faceted_search.html#using-filters)

this PR enables the user to choose a different operator from the default (`=`) while maintaining backward compatibility

solves #529

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -111,12 +111,17 @@ class Builder
      * Add a constraint to the search query.
      *
      * @param  string  $field
+     * @param  string  $operator
      * @param  mixed  $value
      * @return $this
      */
-    public function where($field, $value)
+    public function where($field, $operator, $value = null)
     {
-        $this->wheres[$field] = $value;
+        if(func_num_args() == 2){
+            $value = $operator;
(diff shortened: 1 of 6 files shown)


driesvints:
@G3z please add a much more thorough explanation to the pr for Taylor to review or it will be closed.

G3z:
i hope it's clearer now

taylorotwell:
This would have to be part of a new major release because any third-party Scout drivers would not be expecting where clauses with this format.

G3z:
should I re-target this on master to target the next major ?

driesvints:
@G3z yeah best to target master

PR #532, opened 2021-09-28
support operators in where function

G3z:
Support for query operators like `->where('foo','>','1')` short (old) version still works `->where('foo','1')`

Both Algolia and MeiliSearch supports different operators when filtering:
` <, <=, =, !=, >=, > ` are available for both engine

[Algolia docs](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/how-to/filter-by-numeric-value/#applying-a-numeric-filter)
[MeiliSearch docs](https://docs.meilisearch.com/reference/features/filtering_and_faceted_search.html#using-filters)

this PR enables the user to choose a different operator from the default (`=`) while maintaining backward compatibility

solves #529

PR is now targeted on master

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -111,12 +111,17 @@ class Builder
      * Add a constraint to the search query.
      *
      * @param  string  $field
+     * @param  string  $operator
      * @param  mixed  $value
      * @return $this
      */
-    public function where($field, $value)
+    public function where($field, $operator, $value = null)
     {
-        $this->wheres[$field] = $value;
+        if (func_num_args() == 2) {
+            $value = $operator;
+            $operator = '=';
+        }
+        $this->wheres[$field] = [$operator, $value];
 
         return $this;
     }
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -155,8 +155,9 @@ class AlgoliaEngine extends Engine
      */
     protected function filters(Builder $builder)
     {
-        $wheres = collect($builder->wheres)->map(function ($value, $key) {
-            return $key.'='.$value;
+        $wheres = collect($builder->wheres)->map(function ($operation, $key) {
+            [$operator,  $value] = $operation;
+            return $key.$operator.$value;
         })->values();
 
         return $wheres->merge(collect($builder->whereIns)->map(function ($values, $key) {
--- a/src/Engines/CollectionEngine.php
+++ b/src/Engines/CollectionEngine.php
@@ -89,9 +89,10 @@ class CollectionEngine extends Engine
                             call_user_func($builder->callback, $query, $builder, $builder->query);
                         })
                         ->when(! $builder->callback && count($builder->wheres) > 0, function ($query) use ($builder) {
-                            foreach ($builder->wheres as $key => $value) {
+                            foreach ($builder->wheres as $key => $operation) {
+                                [$operator, $value] = $operation;
                                 if ($key !== '__soft_deleted') {
-                                    $query->where($key, $value);
+                                    $query->where($key, $operator, $value);
                                 }
                             }
                         })
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -156,14 +156,16 @@ class MeiliSearchEngine extends Engine
      */
     protected function filters(Builder $builder)
     {
-        $filters = collect($builder->wheres)->map(function ($value, $key) {
+        $filters = collect($builder->wheres)->map(function ($operation, $key) {
+            [$operator,  $value] = $operation;
+
             if (is_bool($value)) {
-                return sprintf('%s=%s', $key, $value ? 'true' : 'false');
+                return sprintf('%s%s%s', $key, $operator, $value ? 'true' : 'false');
+            } elseif (is_numeric($value)) {
+                return sprintf('%s%s%s', $key, $operator, $value);
             }
 
-            return is_numeric($value)
-                            ? sprintf('%s=%s', $key, $value)
-                            : sprintf('%s="%s"', $key, $value);
+            return sprintf('%s%s"%s"', $key, $operator, $value);
         });
 
         foreach ($builder->whereIns as $key => $values) {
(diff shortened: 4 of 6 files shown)


G3z:
that test fail is wired: I just tested on 7.3.31 and it works 🤔

src/Builder.php
```diff
@@ -111,12 +111,17 @@ public function within($index)
...
      */
-    public function where($field, $value)
+    public function where($field, $operator, $value = null)
     {
-        $this->wheres[$field] = $value;
+        if (func_num_args() == 2) {
+            $value = $operator;
+            $operator = '=';
+        }
+        $this->wheres[$field] = [$operator, $value];
```
driesvints, line 124:
```suggestion
        if (func_num_args() == 2) {
            $value = $operator;
            $operator = '=';
        }

        $this->wheres[$field] = [$operator, $value];
```

src/Engines/CollectionEngine.php
```diff
@@ -89,9 +89,10 @@ protected function searchModels(Builder $builder)
                             call_user_func($builder->callback, $query, $builder, $builder->query);
                         })
                         ->when(! $builder->callback && count($builder->wheres) > 0, function ($query) use ($builder) {
-                            foreach ($builder->wheres as $key => $value) {
+                            foreach ($builder->wheres as $key => $operation) {
+                                [$operator, $value] = $operation;
```
driesvints, line 93:
```suggestion
                                [$operator, $value] = $operation;

```

src/Engines/MeiliSearchEngine.php
```diff
@@ -156,14 +156,16 @@ protected function performSearch(Builder $builder, array $searchParams = [])
      */
     protected function filters(Builder $builder)
     {
-        $filters = collect($builder->wheres)->map(function ($value, $key) {
+        $filters = collect($builder->wheres)->map(function ($operation, $key) {
+            [$operator,  $value] = $operation;
```
driesvints, line 160:
```suggestion
            [$operator, $value] = $operation;
```

src/Engines/AlgoliaEngine.php
```diff
@@ -155,8 +155,9 @@ protected function performSearch(Builder $builder, array $options = [])
      */
     protected function filters(Builder $builder)
     {
-        $wheres = collect($builder->wheres)->map(function ($value, $key) {
-            return $key.'='.$value;
+        $wheres = collect($builder->wheres)->map(function ($operation, $key) {
+            [$operator,  $value] = $operation;
+            return $key.$operator.$value;
```
driesvints, line 160:
```suggestion
            [$operator, $value] = $operation;

            return $key.$operator.$value;
```

driesvints (changes requested):

driesvints (approved):

taylorotwell:
Can you think of a way to support this without breaking every 3rd party driver?

G3z:
I believe incompatibility lives in the `protected function filters(Builder $builder)` method of the engine
since `$builder->wheres` is now a touples array.
right ?

maybe i can use a different property than `$wheres` (`$whereops`, `$whereConditions` ?) if an operator is present
this way _old_ drivers relying on $wheres being an array of string will still work but new o uptated driver can use the new property to get the operator

```php
$query->where('key', '>', 0);
// $builder->whereConditions['key'] = ['>',0]; 
// $builder->wheres['key'] = 0;

$query->where('key', 0)
// $builder->whereConditions['key'] = ['=',0]; 
// $builder->wheres['key'] = 0;
```

new driver can process `$whereConditions` while old can rely on `$wheres`
what do you think? 
should i deprecate `$wheres` ?

taylorotwell:
I do think a new property would prevent breaking changes until drivers have time to update.

src/Builder.php
```diff
@@ -90,7 +97,7 @@ public function __construct($model, $query, $callback = null, $softDelete = fals
         $this->callback = $callback;
 
         if ($softDelete) {
-            $this->wheres['__soft_deleted'] = 0;
+            $this->where('__soft_deleted', 0);
```
G3z, line 100: I used function to support both new and old property

src/Builder.php
```diff
@@ -53,6 +53,13 @@ class Builder
      */
     public $wheres = [];
```
G3z, line 54: should i deprecate this in 10.0 ?

taylorotwell: Yes it could probably be deprecated.

mgralikowski:
Will it cover filtering by range? 
```php
->where('foo', '>', 1)
->where('foo', '<', 10)
```
I am asking, because I mentioned about problem with multiple where conditions in my MR:
https://github.com/laravel/scout/pull/525#issuecomment-924062234

G3z:
no, it does not, the second `where` will obscure the first.
I can fix that by using multidimensional array for `$whereConditions`:
```php
// from
$whereconditions['foo'] = ['>', 1];
// to
$whereconditions['foo'] = [['>', 1], ['<', 10]];
```

i will add this test case

driesvints:
@G3z there seems to be some conflicts as well. Can you fix them?

G3z:
@driesvints what do you mean by conflicts?  i see all checks are green

mgralikowski:
So.. The final result is very similar to what I proposed here: https://github.com/laravel/scout/pull/525
As I understand, we would be able to make:
`$whereconditions['tag'] = [['=', 'foo'], ['=', 'bar']];` // show entries with tag foo and with tag bar
?

G3z:
i see no conflicts in that panel
Schermata 2021-10-05 alle 11 37 58

@mgralikowski you should be able to do
`Model::search('something')->where('tag', 'foo')->where('tag', 'bar');`

taylorotwell:
Honestly this is just turning into more than I want to take on right now. Scout allows you to tap into the underlying Algolia / Meili query if you need to. I've done it before in my own applications.

mgralikowski:
Yeah, everything can be done directly in the engine driver, but then what is the purpose of scout? I thought Scout for search engines is like the Eloquent for a database.

FrittenKeeZ:
I also can't see why this shouldn't be implemented in 10.x, when it won't break backwards compatibility with 3rd party drivers.

PR #537, opened 2021-10-05
Handle sorting (Meilisearch)

mgralikowski:
Sorting in the meilisearch is for a long time and works very good so there is no reason to not implement this functionality.

https://docs.meilisearch.com/reference/features/search_parameters.html#sort

```php
$query->orderBy('date', 'desc');
$query->orderBy('price', 'asc');
```

By this query, we sort results first by date (newest) and next by price (cheapest). Sorting works like in MySQL.

### DIFF
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -97,7 +97,7 @@ class MeiliSearchEngine extends Engine
         return $this->performSearch($builder, array_filter([
             'filters' => $this->filters($builder),
             'limit' => $builder->limit,
-            'sort' => $this->sorters($builder),
+            'sort' => $this->buildSortFromOrderByClauses($builder),
         ]));
     }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:

mgralikowski:
@taylorotwell come on.

It does not add any complexity, it is just a one, protected method... 

There is a whole article about sorting:
https://docs.meilisearch.com/reference/features/sorting.html#configuring-meilisearch-for-sorting-at-search-time

crynobone:
> If you search resources by text, you are not able to sort them anymore for example by date.

Adding this feature will have zero effect on Laravel Nova, the original limitation still persists.

mgralikowski:
The issue with Nova is exactly what my PR resolves - a combination of full-text searching/filtering with sorting at the same moment. In the Nova, we want to find rows by phrase and in the second step - sort them by date - for example.

I have not analyzed the code base of Nova, so I am sure it will not resolve the issue automatically, this is only the first step and basically for Meilisearch only.

francoism90:
Can we add this at least using a macro?

francoism90:
@taylorotwell Thanks for the merge. :)

Issue #535, opened 2021-10-01
Pagination with query callback broke pagination in last meilisearch version


eliasjtg:
- Scout Version: 9.2.9
- Scout Driver: MeiliSearch 0.22
- Laravel Version:8.61.0
- PHP Version: 8.0.9
- Database Driver & Version: mysql 8.0.26-0ubuntu0.20.04.2

### Description:

After upgrade meilisearch to 0.22 and upgrade meilisearch-php 0.19.1 the pagination is broken when queryCallback is provided.

![image](https://user-images.githubusercontent.com/22435698/135650629-0e53bfa3-1cc6-49e2-b69a-d031df6a225f.png)
Builder.php:398 return $totalCount that is always 0

When ->query() callback don't passed, pagination work perfectly
```
return Post::search($search)->paginate(5);
```
![image](https://user-images.githubusercontent.com/22435698/135651388-3a83b34e-4810-404d-b3e7-6dd80b3d9a7c.png)

### Steps To Reproduce:

Using query() reproduce the bug
```
return Post::search($search)->query(function($query) {
    $query->with('image');
})->paginate(5);
```
![image](https://user-images.githubusercontent.com/22435698/135651331-c7ed666f-3629-46eb-ad50-7a97b976b118.png)

driesvints:
Did you restart Meilisearch after upgrading?

eliasjtg:
I'm using Homestead, i do `vagrant reload --provision`


This happening in my team and in production vps

mmachatschek:
@eliasjtg can you post the result of `return Post::search($search)->paginateRaw(5);` so we can verify if this is an issue with scout or with meilisearch?

eliasjtg:
A workaround was remove all ->query() callbacks and do the query in the model overriding `queryScoutModelsByIds()` function

eliasjtg:
> @eliasjtg can you post the result of `return Post::search($search)->paginateRaw(5);` so we can verify if this is an issue with scout or with meilisearch?

Working on it

eliasjtg:
@mmachatschek 

The output:

```
{
  "current_page": 1,
  "data": {
    "hits": [
      {
        "title": "My first post",
        "description": "It's my first post",
        "id": "1633101447865131000",
        "amount": "1",
        "city_id": 1633100293859345000,
        "category_id": 1633101141456362000,
        "sale": 1,
        "created_at": 1633101447,
        "user_deleted": 0,
        "parent_category": null,
[...]

PR #538, opened 2021-10-06
[10.x] Return correct output of mapIds method

mmachatschek:
This fixes issue #535 

The problem with the meilisearch engine in this context is that when calling the `mapIds()` method on the engine, it tries to pluck all primary keys of the results. For MeiliSearch this key is customizable so we can **not** guess the primary key.

The inital implementation just returned the values of the first entry of the result array.

Algolia doesn't have this problem as it always uses `objectID` as the primary key.

This is a breaking change because we need to pass the model to the `mapIds()` method in order to correctly pluck the primary key from the results.


mmachatschek:
@taylorotwell so if the result of the meilisearch payload is not ordered with the primary field first, then the ids are selected from the first element of each item. See this:

grafik

taylorotwell:
But that's not what the existing code is depending on - I don't see anywhere in the current code where we expect the `id` field or any other field to be "first" in the list.

taylorotwell:
Nevermind I see where you are talking about.

mmachatschek:
When doing a search with `Model::search('something')->query(function(){//dosomething})->paginate()` the Builder instance on scout needs to get the primary ids from the underlying engine `mapIds()` , to apply the query added to the builder.
The resulting pagination is broken because there are no entries in the database. because mapIds spits out the wrong primaryKeys to filter in the database query.

mmachatschek:
I'll update the PR tomorrow 👍

Issue #541, opened 2021-10-25
Eager load relationships when re-indexing models using `searchable()`


zaalbarxx:
Currently when dispatching `searchable()` on model or database collection the models will get passed to search driver's `update` method and then the `toSearchableArray()` will get called on each of the items. The issue here is that if we are using relationships in the models we will get n + 1 queries, because each `toSearchableArray()` call will lazy load the relationships for single model.

My proposal is to re-use the currently existing `makeAllSearchableUsing` method and provide additional `loadSearchableRelations()` in `Searchable` trait which will then get called before `toSearchableArray()` on items and eager load the relationships.

PR #542, opened 2021-10-25
[9.x] Eager load relationships when re-indexing models using `searchable` (#541)

zaalbarxx:
Currently when dispatching searchable() on model or database collection the models will get passed to search driver's update method and then the toSearchableArray() will get called on each of the items. The issue here is that if we are using relationships in the models we will get n + 1 queries, because each toSearchableArray() call will lazy load the relationships for single model. This is not a problem when updating single index, but having a situation when we want to update 100 related indexes it will cause unnecessary DB queries.

My proposal is to re-use the currently existing makeAllSearchableUsing method and provide additional loadSearchableRelations() in Searchable trait which will then get called before toSearchableArray() on items and eager load the relationships.

* added `loadSearchableRelations` to `Searchable` trait
* eager loading relationships using `loadSearchableRelations` in drivers before the models will get casted to arrays using `toSearchableArray`

### DIFF
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -160,7 +160,7 @@ trait Searchable
     }
 
     /**
-     * Loads all searchable relations for all models using eager loading
+     * Loads all searchable relations for all models using eager loading.
      */
     public function loadSearchableRelations($models)
     {


liamjoc:
@taylorotwell I do believe this is a bug. n+1 still occurs when SCOUT_QUEUE is set to true.

rickgoemans:
I'm having the same issue and what @liamjoc mentions seems to be correct.

For me the simple temp solution was adding a `$this->loadMissing('...')` at the beginning of the `toSearchableArray` function on the model, this ensures it won't throw any errors but this still has the n+1 problem.

negreanucalin:
This PR seems vital imo because it behaves at it should with queue off which can lead to performance issues

JamesHemery:
> @taylorotwell I do believe this is a bug. n+1 still occurs when SCOUT_QUEUE is set to true.

Same issue

PR #546, opened 2021-11-10
[10.x] Provide searchable data array with primary key and value for MeiliSearch

mmachatschek:
Since `getScoutKeyName()` and `getScoutKey()` return the expected data (https://github.com/laravel/scout/pull/509) we can now build the searchable data ourselves.

This adds consistancy to the already implemented collection engine and Algolia engine that make use of a "predefined" primary key.

Furthermore, this reduces bugs on the implementation side for users that override the `toSearchableArray()` method and forget to add the primary key.

### DIFF
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -62,7 +62,11 @@ class MeiliSearchEngine extends Engine
                 return;
             }
 
-            return array_merge($searchableData, $model->scoutMetadata());
+            return array_merge(
+                [$model->getScoutKeyName() => $model->getScoutKey()],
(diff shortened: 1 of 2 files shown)


taylorotwell:
Any breaking changes for applications already using MeiliSearch and Scout 9?

mmachatschek:
I don't consider this as breaking, because people customizing the `toSearchableArray()` already need to provide the primary key and value with which the data gets fetched from the database later. The `array_merge()` will handle duplicate data in the payload that gets sent to MeiliSearch.

I created #547 that fixes it on the 9.x branch with the "incorrect" use of scout key name handling.

PR #547, opened 2021-11-11
[9.x] Fix issues for users providing searchable array without primary key

mmachatschek:
This is a backport to the changes in #546 for the 9.x branch

### DIFF
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -62,7 +62,11 @@ class MeiliSearchEngine extends Engine
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


src/Engines/MeiliSearchEngine.php
```diff
@@ -62,7 +62,11 @@ public function update($models)
                 return;
             }
 
-            return array_merge($searchableData, $model->scoutMetadata());
+            return array_merge(
+                [$model->getKeyName() => $model->getScoutKey()],
```
mmachatschek, line 66: `$model->getKeyName()` because in line 73 we set the primary key to `$models->first()->getKeyName()`

PR #550, opened 2021-11-21
Add syncToSearchWithoutQueue($callback) to skip using the queue.

dannydinges:
At times it's nice to skip using the queue. This gives you instant indexing in certain situations while still having the ability to use the queue for all other Scout syncs.


taylorotwell:
I would maybe just build your own helper for this that sets `scout.queue` to `false` then sets it back to its previous value after your callback is executed.

PR #553, opened 2021-11-29
[9.x] Pass builder instance to custom search callback

mmachatschek:
This adds the possibility to access data configured on the builder instance in a custom search query. As pointed out here https://github.com/laravel/scout/pull/537#issue-1017081540 there is already a "orderBy" method on the builder instance that is currently only useful to the collection engine but not to the MeiliSearch engine or algolia.

This PR makes the following example possible and leaves the implementation of the orderBy query on the user side if needed and doesn't add additional methods to the builder or engine:

```php
Product::search('query', function($meilisearch, $query, $options, $builder) {
    $options['sort'] = collect($builder->orders)->map(function ($order) {
        return $order['column'] . ':' . $order['direction'];
    })->toArray();
    return $meilisearch->rawSearch($query, $options);
})->where('price', '>', 50)->orderBy('price', 'asc')->paginate(10);
```

### DIFF
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -140,7 +140,8 @@ class AlgoliaEngine extends Engine
                 $builder->callback,
                 $algolia,
                 $builder->query,
-                $options
+                $options,
+                $builder,
             );
         }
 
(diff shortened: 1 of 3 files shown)


taylorotwell:
I'm a little confused why you would use `orderBy` at all. Why not just put the `sort` directly in the search callback instead of retrieving it from the builder. You already know what you want to sort by.

taylorotwell:
We probably just need to revisit the MeiliSearch search PR or have it resubmitted.

PR #556, opened 2021-12-10
Correct issue pagination

oulfr:
When we have data with tow pages the key zero in $result[0] doesn't existe, the list $result index begin from the next page in page size

### DIFF
--- a/src/Engines/CollectionEngine.php
+++ b/src/Engines/CollectionEngine.php
@@ -167,7 +167,7 @@ class CollectionEngine extends Engine
         $results = $results['results'];
 
         return count($results) > 0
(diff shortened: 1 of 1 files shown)


taylorotwell:
I am closing this pull request because it lacks sufficient explanation, tests, or both. It is difficult for us to merge pull requests without these things because the change may introduce breaking changes to the framework.

Feel free to re-submit your change with a **thorough** **explanation** of the feature and **tests** - integration tests are preferred over unit tests. Please include it's benefit to end users; the reasons it does not break any existing features; how it makes building web applications easier, etc.

Thanks!

oulfr:
there is no $results[0] whene we are in $page 2

PR #564, opened 2021-12-23
[9.x] Add a DatabaseEngine

driesvints:
> This PR depends on https://github.com/laravel/framework/pull/40129

This PR adds a new database engine to Scout that's powered by the upcoming `whereFulltext` addition to Laravel's query builder. Using this, we'll be able to offer natural language search in Scout on `fulltext` indexes for Scout.

I managed to quickly set this up thanks to the existing Collection engine. We'll probably need to abstract those so functionality is shared.

The way this works is that you define the columns of your index through the `toSearchableArray` method:

```php
/**
 * Get the indexable data array for the model.
 *
 * @return array
 */
{
    return [
        'name' => $this->name,
        'bio' => $this->bio,
    ];
}
```

Then use it as usual:

```php
$users = User::search('Hello world!')->get();
```

`where` statements to limit results are also supported:

```
$invoices = Invoice::search('Forge')->where('status', 'open')->get();
```

Later on when `whereFulltext` gets supported by more database engines in Laravel, the upgrade will be seamless.

### Todo
- [ ] Figure out a better way to configure columns when models have multiple `fulltext` indexes.
- [ ] Figure out a way to pass options to the `whereFulltext` method. Maybe through the `scoutMetadata` method?
- [ ] Abstract similar functionality from the collection driver.
- [ ] Add tests

### DIFF
--- a/src/Engines/DatabaseEngine.php
+++ b/src/Engines/DatabaseEngine.php
@@ -8,9 +8,10 @@ use Illuminate\Support\LazyCollection;
 use Laravel\Scout\Attributes\SearchUsingFullText;
 use Laravel\Scout\Attributes\SearchUsingPrefix;
 use Laravel\Scout\Builder;
+use Laravel\Scout\Contracts\PaginatesEloquentModels;
 use ReflectionMethod;
 
-class DatabaseEngine extends Engine
+class DatabaseEngine extends Engine implements PaginatesEloquentModels
 {
     /**
      * Create a new engine instance.
@@ -61,19 +62,33 @@ class DatabaseEngine extends Engine
     }
 
     /**
-     * Perform the given search on the engine.
+     * Paginate the given search on the engine.
      *
      * @param  \Laravel\Scout\Builder  $builder
      * @param  int  $perPage
      * @param  int  $page
-     * @return mixed
+     * @return \Illuminate\Contracts\Pagination\LengthAwarePaginator
      */
     public function paginate(Builder $builder, $perPage, $page)
     {
-        return [
-            'results' => $this->searchModels($builder, $page, $perPage),
-            'total' => $this->buildSearchQuery($builder)->toBase()->getCountForPagination(),
-        ];
+        return $this->buildSearchQuery($builder)
+                ->orderBy($builder->model->getKeyName(), 'desc')
+                ->paginate($perPage, ['*'], 'page', $page);
+    }
+
+    /**
+     * Paginate the given search on the engine using simple pagination.
+     *
+     * @param  \Laravel\Scout\Builder  $builder
+     * @param  int  $perPage
+     * @param  int  $page
+     * @return \Illuminate\Contracts\Pagination\Paginator
+     */
+    public function simplePaginate(Builder $builder, $perPage, $page)
+    {
+        return $this->buildSearchQuery($builder)
+                ->orderBy($builder->model->getKeyName(), 'desc')
+                ->simplePaginate($perPage, ['*'], 'page', $page);
     }
 
     /**
@@ -86,15 +101,11 @@ class DatabaseEngine extends Engine
      */
     protected function searchModels(Builder $builder, $page = null, $perPage = null)
     {
-        $models = $this->buildSearchQuery($builder)->when(
+        return $this->buildSearchQuery($builder)->when(
             ! is_null($page) && ! is_null($perPage),
             function ($query) use ($page, $perPage) {
                 return $query->forPage($page, $perPage);
             })->orderBy($builder->model->getKeyName(), 'desc')->get();
-
-        return count($models) > 0
-                ? $models->filter->shouldBeSearchable()->values()
-                : $models;
     }
 
     /**
--- a/src/Engines/DatabaseEngine.php
+++ b/src/Engines/DatabaseEngine.php
@@ -1,0 +1,274 @@
+<?php
+
+namespace Laravel\Scout\Engines;
+
+use Illuminate\Database\Eloquent\SoftDeletes;
+use Illuminate\Support\Arr;
+use Illuminate\Support\LazyCollection;
+use Illuminate\Support\Str;
+use Laravel\Scout\Builder;
+
+class DatabaseEngine extends Engine
+{
+    /**
+     * Create a new engine instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        //
+    }
+
+    /**
+     * Update the given model in the index.
+     *
+     * @param  \Illuminate\Database\Eloquent\Collection  $models
+     * @return void
+     */
+    public function update($models)
+    {
+        //
+    }
+
+    /**
+     * Remove the given model from the index.
+     *
+     * @param  \Illuminate\Database\Eloquent\Collection  $models
+     * @return void
+     */
+    public function delete($models)
+    {
+        //
+    }
+
+    /**
+     * Perform the given search on the engine.
+     *
+     * @param  \Laravel\Scout\Builder  $builder
+     * @return mixed
+     */
+    public function search(Builder $builder)
+    {
+        $models = $this->searchModels($builder);
+
+        return [
+            'results' => $models->all(),
+            'total' => count($models),
+        ];
+    }
+
+    /**
+     * Perform the given search on the engine.
+     *
+     * @param  \Laravel\Scout\Builder  $builder
+     * @param  int  $perPage
+     * @param  int  $page
+     * @return mixed
+     */
+    public function paginate(Builder $builder, $perPage, $page)
+    {
+        $models = $this->searchModels($builder);
+
+        return [
+            'results' => $models->forPage($page, $perPage)->all(),
+            'total' => count($models),
+        ];
+    }
+
+    /**
+     * Get the Eloquent models for the given builder.
+     *
+     * @param  \Laravel\Scout\Builder  $builder
+     * @return \Illuminate\Database\Eloquent\Collection
+     */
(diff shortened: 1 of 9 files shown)


taylorotwell:
We will want a way to also search on normal columns using LIKE or just normal value check like against IDs, etc.

Inspiration for this can be taken from Nova's database search functionality which builds the `or where` clauses for you, etc.

tpetry:
> We will want a way to also search on normal columns using LIKE or just normal value check like against IDs, etc.
How to decide which one to use? Fulltext search and checking against IDs can be index supported, wildcard search with a leading wildcard can not use an index (except on PostgreSQL with pg_trgm). But wildcard search with a trailing wildcard can use an index, but again only if it is indexed and only for an index which will support the operation (hash indexes won't work).

And even for fulltext search scout has to know which columns are fulltext indexed or the query will be stopped with an error on missing index (MySQL) or will be slow (PostgreSQL).

driesvints:
> We will want a way to also search on normal columns using LIKE or just normal value check like against IDs, etc.

@taylorotwell normal value checks are already supported:

```php
$invoices = Invoice::search('Forge')->where('id', 5)->get();
```

For `LIKE` they can already use the callback with the `search` method:

```php
use Laravel\Scout\Builder;

$invoices = Invoice::search('Forge', function ($query, Builder $builder string $search) {
    $query->where('title', 'LIKE', $search);
})->get();
```

That'll invoke the fulltext search and an additional `where like` clause.

We could also add a new `whereLike` method to the Scout builder that only works on the `DatabaseEngine`:

```php
$invoices = Invoice::search('Forge')->whereLike('title', 'Forge')->get();
```

But that might be a bit too much and confusion for the other engines?

@taylorotwell I actually want to know more why you want to support LIKE and additional column clauses. It seems to me we need to focus this all towards fulltext indexes? Like @tpetry says, I don't think it's ideal to mix things up with LIKE, etc.

taylorotwell:
@driesvints it's extremely common to search by primary ID in business applications (for example, order ID) *in addition* to fields that may be full text indexed like a shipping address or order notes field.

The type of search functionality I'm proposing is exactly the kind of search functionality offered by the search field in Laravel Nova.

I will look into this PR more tomorrow.

src/Engines/DatabaseEngine.php
```diff
@@ -0,0 +1,273 @@
...
+        return $models->filter(function ($model) use ($builder) {
+            if (! $model->shouldBeSearchable()) {
+                return false;
+            }
+
+            if (! $builder->query) {
+                return true;
+            }
+
+            return false;
```
taylorotwell, line 124: I feel like I must be missing something really obvious here. This `filter` operation will basically always remove every model from the result set unless the search query is empty? How does this ever actually return any results in normal circumstances?

driesvints:
> So, for example, if I type "1" in the search box I may want to find the user with an ID of 1, while I can also type "Arkansas" in the search box to find orders with "Arkansas" in the full-text indexed notes. Without that functionality I frankly would not find a database driver for Scout useful at all.

How will you know which field(s) to search? Sounds more like the ID is part of a separate filter than the search box?

> The type of search functionality I'm proposing is exactly the kind of search functionality offered by the search field in Laravel Nova, which I think I noted that code may provide some good inspiration here as it works really well across multiple database drivers.

I'll investigate that.

src/Engines/DatabaseEngine.php
driesvints: Odd, it worked when I tried it but your remarks seem to make sense yeah.

driesvints: I pushed a change for this. I think I must have reverted something the last time I pushed. Anyway, it's fixed now.

driesvints:
> A possible solution for the database driver could be a specification for every column. I like the approach of pmatseykanets/laravel-scout-postgres. Additionally to the normal scout configuration, another array is defined with options for the driver. The search method would be defined for every column (wildcard/fulltext) and be used by the driver to build the query.

Does that mean that we add "or" clauses for every mapped column?

tpetry:
> Does that mean that we add "or" clauses for every mapped column?

That was my first idea. But it does not work as defining multiple columns for fulltext searching would try to do a fulltext search on each by it's own and then `or` the results, which will not work.

Maybe something like this (i don't like the structure, just sharing the idea):

```php
public function searchableOptions()
{
    return [
        'columns' => [
            'like_search' => ['id', 'email'], // these can be "or"ed
            'fulltext_search' => [ // each group needs to use one whereFullText and then can be "or"ed
                ['body', 'title],
                ['subject', 'description],
            ],
        ],
    ];
}
```

taylorotwell:
Hey all,

I've pushed my work for today which is a pretty big rewrite of this. It now behaves more like Laravel Nova search with the added ability for real full-text searching with the new `whereFullText` methods as well as a combination of LIKE and also LIKE with no leading %.

I built a small search interface for myself to test it with:

CleanShot 2022-01-06 at 15 18 58@2x

### Overview

So, given the following `toSearchableArray` method:

CleanShot 2022-01-06 at 15 18 21@2x

The following naive, LIKE based query would be executed by default, allowing very basic searching across all columns:

CleanShot 2022-01-06 at 15 19 59@2x

This is exactly like Nova.

If you try to search by an integer and the primary key type is an integer for the model, it will attempt an exact search against that value instead of a LIKE - again, just like Nova.

#### Full Text Support

You may easily specify which columns actually have full text indexes using Attributes:

image

This now generates a query like so:

CleanShot 2022-01-06 at 15 21 30@2x

#### LIKE Without Wildcard Prefix

Some columns you may want to be searched using `where like 'foo%'` - with no leading `%` on the like for better performance. This is also supported:

CleanShot 2022-01-06 at 15 22 18@2x

#### Other Improvements

I have made quite a few other improvements. Most notably, the original implementation of this driver executed 2 database queries... one to perform the search and then an extra, fairly useless query on all the matching models by ID using a `where in` statement. That has been removed and now the models returned from the initial search query are used as the actual result set so only 1 query is executed to perform the search.

robsontenorio:
Does fulltext approach , instead MeiliSeach or ElasticSearch, implies in too much heavy load on database? Locks ? Performance?

Does anybody has any concerns about that ?

I really would love to get rid off that extra piece of infrastructure (Melisearch/Elastic)

fhferreira:
> 

+1

> Does fulltext approach , instead MeiliSeach or ElasticSearch, implies in too much heavy load on database? Locks ? Performance?
> 
> Does anybody has any concerns about that ?
> 
> I really would love to get rid off that extra piece of infrastructure (Melisearch/Elastic)

+1

driesvints:
@taylorotwell this looks amazing, thanks! I do believe the `SearchUsingFullText` attribute needs an options array for extra options to pass to the `whereFullText` method for engines underneath to use. 

/cc @tpetry ☝️

src/Engines/DatabaseEngine.php
```diff
@@ -0,0 +1,342 @@
...
+                        $builder->model->qualifyColumn($column),
+                        $builder->query
+                    );
+                } else {
+                    if ($canSearchPrimaryKey && $column === $builder->model->getKeyName()) {
+                        continue;
+                    }
+
+                    $query->orWhere(
+                        $builder->model->qualifyColumn($column),
```
crynobone, line 157:
```suggestion
                        $builder->qualifyColumn($column),
```

src/Engines/DatabaseEngine.php
```diff
@@ -0,0 +1,342 @@
...
+            if ($canSearchPrimaryKey) {
+                $query->orWhere($builder->model->getQualifiedKeyName(), $builder->query);
+            }
+
+            $likeOperator = $connectionType == 'pgsql' ? 'ilike' : 'like';
+
+            foreach ($columns as $column) {
+                if (in_array($column, $fullTextColumns)) {
+                    $query->orWhereFullText(
+                        $builder->model->qualifyColumn($column),
```
crynobone, line 148:
```suggestion
                        $builder->qualifyColumn($column),
```

Available since 5.5 https://github.com/laravel/framework/pull/22577

taylorotwell:
> Does fulltext approach , instead MeiliSeach or ElasticSearch, implies in too much heavy load on database? Locks ? Performance?
> 
> Does anybody has any concerns about that ?
> 
> I really would love to get rid off that extra piece of infrastructure (Melisearch/Elastic)

I have no idea on the performance. This driver is intended for usage on smaller data sets and workloads where such performance isn't a concern.

tpetry:
> Does fulltext approach , instead MeiliSeach or ElasticSearch, implies in too much heavy load on database? Locks ? Performance?
> 
> Does anybody has any concerns about that ?
> 
> I really would love to get rid off that extra piece of infrastructure (Melisearch/Elastic)

It's highly dependent on your database, your data model and your queries. In the past MySQL was pretty bad when you did a fulltext search with a condition to filter rows because MySQL was not able to use multiple indexes for a single query. But that has changed with 8.0, whether it will use multiple indexes or not is still depending on a lot of things and some magic of MySQL's query planner.
I only have recent experience with PostgreSQL's fulltext search which is working really nice with a few million entries. And as PostgreSQL will use multiple indexes since 8.1 (2005) the query planner is really good at it and you can blindly throw queries at it and it will execute the fulltext queries very fast. 

As said, there is no general statement to performance, it depends on a lot of things. It's best you try your workload and if it's working you can be happy to remove Melisearch/Elastic from your stack.

taylorotwell:
> Instead of researching performance problems of solution x just implement it and test it.

✔️

PR #566, opened 2022-01-06
[9.x] use whereIntegerInRaw instead of whereIn

it-can:
```whereIn``` can become slow with a lot of id's passed, ```whereIntegerInRaw``` should be used here I guess

### DIFF
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -239,7 +239,7 @@ trait Searchable
             call_user_func($builder->queryCallback, $query);
         }
 
-        return $query->whereIn(
+        return $query->whereIntegerInRaw(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by taylorotwell** [comment]
What if they are UUIDs?

**Comment by it-can** [comment]
Aah you are right, didn't think about this... would an extra check be worth doing? Something like

```

public static function isNumericArray($array)
    {
        foreach ($array as $a) {
            if (! is_numeric($a)) {
                return false;
            }
        }

        return true;
    }

 $whereIn = static::isNumericArray($ids) ?
            'whereIntegerInRaw' :
            'whereIn';

        return $query->{$whereIn}(
            $this->getScoutKeyName(), $ids
        );
```

PR #574, opened 2022-01-14
Return paginator for paginateRaw (fixes #573)

beko1997:
This fixes #573 as paginateRaw is currently always treated as a void method in 9.4.0.

Issue #573, opened 2022-01-14
Returned null with paginateRaw after upgrading to 9.4.0


beko1997:
- Scout Version: 9.4.0
- Scout Driver: Algolia
- Laravel Version: 8.79
- PHP Version: 8.0
- Database Driver & Version: SQL Server

### Description:
`paginateRaw` returns null after upgrading to version 9.4.0. It is returning a valid paginator with version 9.3.4.

### Steps To Reproduce:

````php
$result = SpecificModel::search('')
  ->within('valid_index_name')
 ->with([
        'facets' => ['facet1', 'facet2'],
        'filters' => ''
    ])
   ->paginateRaw(50, 'page', 1);

$result === null; // true
````

Result in 9.3.4 is a valid pagination response.

beko1997:
It probably has something to do with not returning the paginatior in de `Builder`:
https://github.com/laravel/scout/blob/7811f6d025cd0024da19a065b0eefb43e6e7eb69/src/Builder.php#L417

I will work on a PR for a fix

PR #575, opened 2022-01-14
Added tests for paginateRaw

beko1997:
This PR adds tests for the `paginateRaw` method in `Builder.php`. See #574 and #573.


driesvints:
Thanks @beko1997

PR #582, opened 2022-02-08
Skip adding search constraints with empty search on `DatabaseEngine`

rodrigopedra:
Right now if we search for an empty string using the `DatabaseEngine`, for example in a endpoint that lists all users:

```php
Route::get('/api/users', fn () => User::search(request()->query('q'))->paginate());
```

Even when the search string is empty, the where clauses are added:

```sql
select *
from `users`
where (`users`.`name` like '%%' or `users`.`email` like '%%' or match (`users`.`notes`) against ('' in natural language mode))
order by `id` desc
limit 30 offset 0
```

This PR skips adding the search where clauses if the an empty search string.

### DIFF
--- a/src/Engines/DatabaseEngine.php
+++ b/src/Engines/DatabaseEngine.php
@@ -139,6 +139,10 @@ class DatabaseEngine extends Engine implements PaginatesEloquentModels
      */
     protected function initializeSearchQuery(Builder $builder, array $columns, array $prefixColumns = [], array $fullTextColumns = [])
     {
+        if (blank($builder->query)) {
+            return $builder->model->query();
+        }
+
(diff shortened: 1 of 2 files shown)


src/Engines/DatabaseEngine.php
```diff
@@ -139,6 +139,10 @@ protected function buildSearchQuery(Builder $builder)
      */
     protected function initializeSearchQuery(Builder $builder, array $columns, array $prefixColumns = [], array $fullTextColumns = [])
     {
+        if (blank($builder->query)) {
```
driesvints, line 142: Did you mean `empty`?

taylorotwell: `blank` is a Laravel support helper

Issue #586, opened 2022-02-15
MeilisearchEngine paginate() incorporate sort


mrABR:
- Scout Version: ^9.4
- Scout Driver: MeiliSearch
- Laravel Version: 8.83.1
- PHP Version: 8.1
- Database Driver & Version: MySql 8
- Meilisearch Version: v0.25.0 (behaves the same with 0.22.0)
- Meilisearch-php Version: ^0.22.0

Allow me to start by thanking all the work you guys have put into Laravel and its community with the constant evolution.

### Description:

Recently I've implemented scout on a project to serve a faster search. After implementing it, I've imported the records required and created filtered attributes and sortable attributes, so that it would eventually come to replace the mysql search implementation that already exists on the project.

One of the conditions on the project is that: if a paginated variable is sent on the query parameter, it should paginate, otherwise it should return all (well, not really all but this is to simplify the situation).

This is when I've noticed that the MeilisearchEngine `paginated` function doesn't apply the sort, while the `search` (that is used by `get` in inheritance by the abstract Engine) applies the sort.

### Steps To Reproduce:

`http://localhost/users?paginate=20&sort=name+asc` selects the first 20 records, doesn't sort by name ascending
`http://localhost/users?sort=name+asc` selects the first 20 records (default by meilisearch) sorted by name ascending

```
//This snippet is simplified to test out the behaviour
public function index(Request $request)
    {
       [$column, $direction] = explode(' ', $request->sort);
       $builder = User::search($request->search)->orderBy($column, $direction);

        $records = $request->has('paginate')
            ? $builder->paginate($request->paginate)
            : $builder->get();
        return response(new UserCollection($records), 200);
    }
```

In MeilisearchEngine, we've got this `paginate` function which is very similar to the `search` function used by `get`.

```
 /**
     * Perform the given search on the engine.
     *
     * @param  int  $perPage
     * @param  int  $page
     * @return mixed
     */
    public function paginate(Builder $builder, $perPage, $page)
    {
        return $this->performSearch($builder, array_filter([
            'filters' => $this->filters($builder),
            'limit' => (int) $perPage,
            'offset' => ($page - 1) * $perPage,
        ]));
    }
```

I've tried playing around it and found that we can provide a paginated AND sorted result if we add the following line to the `array_filter` :  `'sort' => $this->buildSortFromOrderByClauses($builder)`

I believe this use case is interesting, as it provides a way to return results very similarly to an order by query on a regular database.

mrABR:
[...]

PR #587, opened 2022-02-15
Implements Meilisearch sort on paginate

### LINKED ISSUE #572: Sorting feature is missing in MeiliSearch Engine
- Scout Version: 9.3.4
- Scout Driver:  MeiliSearch
- Laravel Version: 8.61
- PHP Version: 8.0.14
- Database Driver & Version: Mysql 8

### Description:
MeiliSearch supports sorting feature out of the box after defining sortable parameters while indexing. The given `orderBy` method is supposed to work for sorting. Consider the following code. 

```php
return User::search('Admin')
        ->orderBy("id", "desc")
        ->get();
```

The above code should generate a query like in https://github.com/laravel/scout/blob/9.x/src/Engines/MeiliSearchEngine.php#L114

```php
public function paginate(Builder $builder, $perPage, $page) 
 { 
     return $this->performSearch($builder, array_filter([ 
         'filters' => $this->filters($builder), 
         'limit' => (int) $perPage, 
         'sort'=> ['id:desc'],
         'offset' => ($page - 1) * $perPage, 
     ])); 
 } 
```
But the `sort` options is missing while performing the search. 

https://github.com/laravel/scout/blob/7811f6d025cd0024da19a065b0eefb43e6e7eb69/src/Engines/MeiliSearchEngine.php#L114-L121

https://github.com/laravel/scout/blob/7811f6d025cd0024da19a065b0eefb43e6e7eb69/src/Engines/MeiliSearchEngine.php#L99-L105

mrABR:
I've recently opened an issue (#586) and found out this was a pending PR that wasn't submitted yet

Fixes #572

### DIFF
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -118,6 +118,7 @@ class MeiliSearchEngine extends Engine
             'filters' => $this->filters($builder),
             'limit' => (int) $perPage,
             'offset' => ($page - 1) * $perPage,
+            'sort' => $this->buildSortFromOrderByClauses($builder),
         ]));
     }
 
--- a/tests/Unit/MeiliSearchEngineTest.php
+++ b/tests/Unit/MeiliSearchEngineTest.php
@@ -325,6 +325,29 @@ class MeiliSearchEngineTest extends TestCase
         $engine->paginate($builder, $perPage, $page);
     }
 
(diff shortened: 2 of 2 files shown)


mgralikowski:
Thanks for fixing and merging. My fault in the original PR - I forget about paginated results and I am here to make a fix, but you beat me to it :)

PR #591, opened 2022-02-18
Call queryCallback in DatabaseEngine

Alanaktion:
The `->query()` callback function on searches is helpful for things like eager loading relations, and is typically triggered via `getScoutModelsByIds` in engine implementations. Since the Database engine works directly with the models and does not call that method, I've added an explicit call to `queryCallback` when it's defined.

In my limited testing, this allows things like the following to work consistently on several engines including MeiliSearch and Database without requiring any engine-specific end-user code changes:

```php
Post::search('example query')
    ->query(function ($builder) {
        $builder->with('comments');
    })
    ->paginate(15);
```

It doesn't seem like that callback is documented so it may not be widely used, but since it is implemented consistently in the other official engines this seemed like a good contribution.

### DIFF
--- a/src/Engines/DatabaseEngine.php
+++ b/src/Engines/DatabaseEngine.php
@@ -206,6 +206,8 @@ class DatabaseEngine extends Engine implements PaginatesEloquentModels
             foreach ($builder->whereIns as $key => $values) {
                 $query->whereIn($key, $values);
             }
+        })->when(! is_null($builder->queryCallback), function ($query) use ($builder) {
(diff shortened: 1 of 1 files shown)


Alanaktion:
That does require the user to write engine-specific search queries rather than have a generalized solution that works consistently with one implementation, but if that's what's preferred by the Laravel team that's fine I guess. I'll probably just not use Scout at all for the database engine if I have to write custom logic to perform simple searches across multiple engines either way.

PR #592, opened 2022-02-18
Remove redundant `return` key like for all `when` methods for databas…

siarheipashkevich:


driesvints (approved):

siarheipashkevich:
@driesvints I wanna ask you, in the spec can we remove also `return` key https://laravel.com/docs/9.x/queries#conditional-clauses ?

driesvints:
@siarheipashkevich yeah think so 👍

PR #609, opened 2022-03-30
Add config for appends() parameter name

OrangePixelLtd:
### Small change to add the ability to customize the query string parameter

**Example**  https://domain.com?search=term INSTEAD OF https://domain.com?query=term 
I've seen a few requests for this and didn't find a solution out there.  This may not be the best solution but it works for me without breaking anything else.

**Added**
I have added a config setting with a default of 'query' so it shouldn't break anything and updated the Builder class methods.

`'parameter_name' => env('SCOUT_PARAMETER_NAME', 'query')`

**Why?**
Pagination links were overriding my original parameter. Using Laravel / Inertia / React

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -102,6 +102,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Query Parameter Name
+    |--------------------------------------------------------------------------
+    |
+    | This option allows you to control what name string will be used to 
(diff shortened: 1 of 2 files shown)


OrangePixelLtd:
Works but couldn't get tests to pass and not got a lot of time unfortunately.  Sorry all

PR #610, opened 2022-04-01
Add a method to convert a search result into an Eloquent query builder

Synchro:
I'm using scout with Algolia, and I'm implementing post-search filters. Scout only has very rudimentary abilities for that, but eloquent has lots. Once scout has done its thing and given you a response, it would be useful to be able to convert that into an eloquent builder and carry on adding filtering operations from there.

This PR adds a simple method to convert a scout search result into an eloquent builder. It works by extracting the keys from the scout result and using them in a `whereIn` method. It uses the `keys` method to get the model IDs directly from the search result and doesn't look them up again unnecessarily. It's obviously somewhat inefficient because you're doing two searches, but it does make it very easy to work with.

Example:

    $result = SearchableUserModel::search('Laravel')
        ->toEloquent()
        ->where('created_at', '>', Carbon::yesterday());

The test I initially wrote failed, but I discovered that this was not my doing, so I've included a failing test for that (with no other code changes) as a separate commit. It seems that calling `search()->keys()` triggers an error in Meilisearch that I don't understand. My method calls the same function, and so suffers the same failure – if you fix it, both will work!

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -255,6 +255,21 @@ class Builder
     }
 
     /**
+     * Convert a search result into an Eloquent builder.
+     *
+     * @return \Illuminate\Database\Eloquent\Builder
+     */
+    public function toEloquent()
+    {
+        return $this->model
+            ->query()
+            ->whereIn(
+                $this->model->getQualifiedKeyName(),
+                $this->keys()
+            );
+    }
+
+    /**
      * Get the first result from the search.
      *
      * @return \Illuminate\Database\Eloquent\Model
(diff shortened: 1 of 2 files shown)


taylorotwell:
Personally I just suggest filtering the collection outside of the DB after you retrieve it. This is going to break pagination counts, etc.

Synchro:
I'm aware of that, but after the conversion I can use Eloquent's own pagination instead. It also means I can use Scout with things that are not otherwise compatible with it, for example it makes it much easier to use Filament's filter UI components (which only work with Eloquent) with Scout.

JerryBels:
I find this idea very interesting. Scout is awesome, but Eloquent capabilities in sorting, filtering etc... are on an entirely other level. It would be very useful.

websitevirtuoso:
Sad that this request failed. Idea great. Hope to get it work with another implementations at lease. I will copy this code as good idea of using scout and Eloquent together

johncarter-:
The other problem with this is that the order of the search results (i.e. relevancy) becomes jumbled when using a `whereIn` as far as I have tested. I'm still trying to find a solution to this 🤔

PR #624, opened 2022-05-10
[9.x] Add import all option to import command

mozex:
Before this PR, you had to run `php artisan scout:import "App\Models\{Model}"` for every model manually which could be exhausting for existing projects with a lot of models, especially when you had to make some changes and then import them again.

This PR makes it possible to just run `php artisan scout:import` without any argument and it finds all the models that use the Searchable trait and imports them but it also accepts an optional model argument that acts like the original command.

### DIFF
--- a/src/Console/ImportCommand.php
+++ b/src/Console/ImportCommand.php
@@ -4,7 +4,12 @@ namespace Laravel\Scout\Console;
 
 use Illuminate\Console\Command;
 use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Arr;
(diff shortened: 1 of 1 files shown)


taylorotwell:

mozex:
Following the feedback on this pull request, I've released the proposed feature as a standalone package called "[Laravel Scout Bulk Actions](https://github.com/mozex/laravel-scout-bulk-actions)".

This package provides additional commands that extend Laravel Scout's functionality when dealing with multiple models.

You can explore and use the package from its GitHub repository: [Laravel Scout Bulk Actions](https://github.com/mozex/laravel-scout-bulk-actions).

I hope you find it useful for your projects. Feedback and contributions are always appreciated.

PR #632, opened 2022-07-13
Meilisarch v0.28.0 Breaking Change Adjustment

frankyso:
Hello there based on the meilisearch release on version [v0.28.0](https://github.com/meilisearch/meilisearch/releases/tag/v0.28.0)

there are some breaking changes in the total record counts, and it affects the laravel scout pagination.

thanks

PR #633, opened 2022-07-15
fix(meilisearch): change nbHits to estimatedTotalHits

Carlwirkus:
# Scope
Meilisearch [v0.28.0](https://github.com/meilisearch/MeiliSearch/releases) has a breaking change which renames `nbHits` to `estimatedTotalHits`. This breaks when trying to paginate the response. Functionally they have the same implementation.

I've looked to see if the engine uses the other parameters that were changed but it doesn't look like it does.


driesvints:
Please see https://github.com/laravel/scout/pull/632

Issue #615, opened 2022-04-06
Meilisearch engine pagination total broken


eliasjtg:
- Scout Version: 9.4.6
- Scout Driver: MeiliSearch
- Laravel Version: 8.83.6
- PHP Version: 8.0.16
- Database Driver & Version: Mysql 8.0.28-0ubuntu0.20.04.3
- SDK Version (If using a third-party service): 0.23.1

### Description:

This issue is related with https://github.com/laravel/scout/issues/535 and map id with key name, the solution was https://github.com/laravel/scout/commit/8a7782c5ce72b9407a3cb2e29a4865cf4efc964b creating mapIdsFrom method

The same issue is occurring with the method keys in Engine, does not take into account mapIdsFrom and call directly mapIds, which breaks the pagination count.

A solution is create keysFrom method:

```
/**
     * Get the results of the query as a Collection of primary keys.
     *
     * @param  \Laravel\Scout\Builder  $builder
     * @param  string  $key
     * @return \Illuminate\Support\Collection
     */
    public function keysFrom(Builder $builder, $key)
    {
        return $this->mapIdsFrom($this->search($builder), $key);
    }
```

and modify the getTotalCount method:

```
/**
     * Get the total number of results from the Scout engine, or fallback to query builder.
     *
     * @param  mixed  $results
     * @return int
     */
    protected function getTotalCount($results)
    {
        $engine = $this->engine();

        $totalCount = $engine->getTotalCount($results);

        if (is_null($this->queryCallback)) {
            return $totalCount;
        }

        $ids = $engine->mapIdsFrom(
            $results,
            $key = Str::afterLast($this->model->getScoutKeyName(), '.')
        )->all();

        if (count($ids) < $totalCount) {
            $ids = $engine->keysFrom(tap(clone $this, function ($builder) use ($totalCount) {
                $builder->take(
                    is_null($this->limit) ? $totalCount : min($this->limit, $totalCount)
                );
            }), $key)->all();
        }

        return $this->model->queryScoutModelsByIds(
            $this, $ids
        )->toBase()->getCountForPagination();
    }
```

Changing `$key = Str::afterLast($this->model->getScoutKeyName(), '.')` and calling $engine->keysFrom reusing the `$key` variable

### Steps To Reproduce:

Using query() reproduce the bug

return Post::search($search)->query(function($query) {
    $query->with('image');
})->paginate(5);

driesvints:
@mmachatschek is this a Meilisearch incompatibility again?

[...]

PR #640, opened 2022-08-05
Added ability to pass primary key into mapping methods

MatusBoa:
As mentioned here https://github.com/laravel/scout/issues/615#issuecomment-1191016297, there is a bug that causes pagination with Meilisearch to not work at all.

Whole problem is that, sometimes meilisearch (found out that it may be due to changing searchable fields) does not return primary key as first key hits result causing pagination to not work at all.

In my case, meilisearch `hits` array looks something like this

```
[
  [
    'name' => 'FooBar',
    'uuid' => '...'
  ]
]
```

so Scout thinks `name` is the primary key for model.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -448,7 +448,7 @@ class Builder
         )->all();
 
         if (count($ids) < $totalCount) {
-            $ids = $engine->keys(tap(clone $this, function ($builder) use ($totalCount) {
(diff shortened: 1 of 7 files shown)


MatusBoa:
Not sure why tests are failing, I think it's not related to changes in this PR 

https://github.com/laravel/scout/runs/7688304690?check_suite_focus=true

taylorotwell:
This PR would break every third-party Scout driver on a patch release because of changing method signatures.

PR #643, opened 2022-08-17
Attempt to resolve model by name

MortenDHansen:
It would be a friendly gesture to the user if scout tried to find the model typed in the command.

Example. I tried to flush a searchable today. First writing `artisan scout:flush MyModelName`. Then i remembered it had to be qualified and wrote `artisan scout:flush App/Models/MyModelName`. Then i remembered it had to be backslashes `artisan scout:flush App\Models\MyModelName` ... etc 

Let scout be the friend that checks the Models directory to see if you meant "App\Models\MyModelName" ❤️

### DIFF
--- a/src/Console/FlushCommand.php
+++ b/src/Console/FlushCommand.php
@@ -27,7 +27,12 @@ class FlushCommand extends Command
      */
     public function handle()
     {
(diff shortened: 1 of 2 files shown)


taylorotwell:

Issue #651, opened 2022-09-23
Scout won't return correct model `->keys()` when using custom `searchableAttributes` from Meilisearch


flexchar:
- Scout Version: 9.4.10
- Scout Driver: MeiliSearch
- Laravel Version: 9.27.0
- PHP Version: 8.1.8
- Meilisearch CLI Version (If using Meilisearch): 0.28.1

### Description:

I've discovered that `Searchable::search('query terms')->keys();` fails when customizing [ `searchableAttributes` ](https://docs.meilisearch.com/reference/api/settings.html#searchable-attributes) by Meilisearch.

Let's say you have comments blog that contains more than just message body. You want to only search based on its contents so you only list those attributes. Doing so will force Meilisearch to return raw results in different order. Below is the snippet of `Searchable::serach()->raw()`. **Notice how 'id` is no longer the first element in the array.**

```php
array:7 [
  "hits" => array:20 [
    0 => array:11 [
      "body" => "Lorem Ipsum blabla"
      "translation" => ""
      "id" => 42769
      "media_id" => 111
      "conversation_id" => 19
      "language" => null
      "user_id" => 1
      "visibility" => 3
    ]
    1 => array:11 [
      "body" => "Lorem Ipsum blabla"
      "translation" => ""
      "id" => 44648
      "media_id" => 258
      "conversation_id" => 31
      "language" => null
      "user_id" => 1
      "visibility" => 3
    ]
 ],
  "estimatedTotalHits" => 96
  "query" => "blabla"
  "limit" => 20
  "offset" => 0
  "processingTimeMs" => 3
  "nbHits" => 96
]
```

As such it forces the following code to break search: https://github.com/laravel/scout/blob/48aeabdd80763708ac704f6053f7b83e222fe007/src/Engines/MeiliSearchEngine.php#L219.

Maybe we could use `->mapIdsFrom(string $key)` and pass `->getScoutKeyName()` into? 

### Steps To Reproduce:

I hope I explained well. If not, please ping me and I will set up a repo.

driesvints:
Thanks @flexchar. @mmachatschek I'm not sure it was intended that Scout supported this? The related PR does look good though.

mmachatschek:
@driesvints the response seems legit, BUT I think @curquiza mentioned somewhere, that the search response will return the primaryKey as the first element in the search payload of the document.

@curquiza did this recently change? I do remember to have tested this a few times were the primaryKey was the first element in the payload

curquiza:
[...]

PR #654, opened 2022-09-27
[10.x] Use scout key when mapping keys from search results #652 

flexchar:
This is the fix against https://github.com/laravel/scout/issues/651 targeted for changes for the release tree 10.x. Adapted from https://github.com/laravel/scout/pull/652

**Context**

This PR overloads the `keys()` method in the base Engine class providing `MeilisearchEngine` the exact Scout key to be used when mapping search results.

Previously the key was determined by taking the first key of the array in the search response. However as recently discovered in the aforementioned issue - Meilisearch will return random array order depending on its index configuration.

@driesvints please let me know if it's ready for review or needs improvement. :)

### DIFF
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -220,6 +220,19 @@ class MeiliSearchEngine extends Engine
                 : collect($results['hits'])->pluck($key)->values();
     }
 
+     /**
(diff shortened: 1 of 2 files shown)


driesvints:
Thanks but this isn't needed anymore. I'll merge 9.x into master after your PR gets merged and do the changes.

PR #655, opened 2022-09-28
Proof of issues with calculated custom key values

DarronEngelbrechtEdge:
Currently this PR only serves to prove issues with calculated custom key values provided through `getScoutKey()` for a Searchable model through minimal alternation of test "Fixtures" and expected changes in the tests.

The unit tests in this branch now fail with the following output. Note that the actual value received is an md5 hash of the expected value which is an md5 hash of the models `other_id` attribute.

```
There was 1 failure:

1) Laravel\Scout\Tests\Unit\RemoveFromSearchTest::test_models_are_deserialized_without_the_database_using_custom_scout_key
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'81dc9bdb52d04dc20036dbd8313ed055'
+'ec6a6536ca304edf844d1d248a4f08dc'

~/scout/tests/Unit/RemoveFromSearchTest.php:69
```

### DIFF
--- a/tests/Unit/RemoveFromSearchTest.php
+++ b/tests/Unit/RemoveFromSearchTest.php
@@ -66,7 +66,7 @@ class RemoveFromSearchTest extends TestCase
         $this->assertCount(1, $job->models);
         $this->assertInstanceOf(SearchableModelWithCustomKey::class, $job->models->first());
         $this->assertTrue($model->is($job->models->first()));
-        $this->assertEquals(1234, $job->models->first()->getScoutKey());
+        $this->assertEquals(md5(1234), $job->models->first()->getScoutKey());
         $this->assertEquals('searchable_model_with_custom_keys.other_id', $job->models->first()->getScoutKeyName());
     }
 
@@ -80,8 +80,8 @@ class RemoveFromSearchTest extends TestCase
         ]);
 
         $this->assertEquals([
-            1234,
-            2345,
+            md5(1234),
+            md5(2345),
             3456,
             7891,
         ], $collection->getQueueableIds());
--- a/tests/Fixtures/SearchableModelWithCustomKey.php
+++ b/tests/Fixtures/SearchableModelWithCustomKey.php
@@ -18,7 +18,7 @@ class SearchableModelWithCustomKey extends Model
 
     public function getScoutKey()
     {
-        return $this->other_id;
+        return md5($this->other_id);
     }
 
     public function getScoutKeyName()


DarronEngelbrechtEdge:
@mmachatschek @driesvints - I'm wondering if the `test_models_are_deserialized_without_the_database_using_custom_scout_key` test should be present. If my understanding is correct the SerializesModels trait in the RemoveFromSearch job will ensure that only the model ID is stored in the job and the model must then be retrieved from the database to fetch the custom attribute?

Removing the SerializesModels trait from the RemoveFromSearch job gets the tests to pass but is likely a terrible solution.

driesvints:
These changes were made here: https://github.com/laravel/scout/pull/480 to solve https://github.com/laravel/scout/pull/479.  @DarronEngelbrechtEdge did you also set the `getScoutKeyName` in your model as detailed in PR #480? (before the changes in this PR).

Ping @stevebauman can you maybe have a look here?

DarronEngelbrechtEdge:
@driesvints - No I do not modify `getScoutKeyName` other than to specify it now as the string of `id` which is the primary key for the model.

Had a look at those PRs and #479 explains the problem I'm having but #480 fixes the use case where both `getScoutKeyName` and `getScoutKey` have been modified not the more simple case where only `getScoutKey` is modified?

I believe this is not caught in testing because SCOUT_QUEUE is false and the models are not serialized. 

Perhaps smarter people can figure out how if it's possible that "models_are_deserialized_without_the_database_using_custom_scout_key" so we can restore that test?

tests/Unit/RemoveFromSearchTest.php
```diff
@@ -54,20 +54,20 @@ public function test_models_are_deserialized_without_the_database()
...
 
         $this->assertInstanceOf(Collection::class, $job->models);
         $this->assertCount(1, $job->models);
-        $this->assertInstanceOf(SearchableModelWithCustomKey::class, $job->models->first());
+        $this->assertInstanceOf(MeiliSearchCustomKeySearchableModel::class, $job->models->first());
         $this->assertTrue($model->is($job->models->first()));
-        $this->assertEquals(1234, $job->models->first()->getScoutKey());
-        $this->assertEquals('searchable_model_with_custom_keys.other_id', $job->models->first()->getScoutKeyName());
+        $this->assertEquals('my-meilisearch-key.1234', $job->models->first()->getScoutKey());
+        $this->assertEquals('meili_search_custom_key_searchable_models.id', $job->models->first()->getScoutKeyName());
```
stevebauman, line 70: This test will pass because the `MeiliSearchCustomKeySearchableModel` fixture is being instantiated with primary key that is being used in the `getScoutKey()` method:

```php
public function getScoutKey()
{
    return 'my-meilisearch-key.'.$this->getKey();
}
```

https://github.com/DarronEngelbrechtEdge/scout/blob/246bf96e1042f1b1adc586f69f6260e389a1dccb/tests/Unit/MeiliSearchEngineTest.php#L475-L478

If another model attribute is used to construct the scout key, then this will fail, as it cannot be retrieved from the faux model that is created after the job is unserialized. All other model properties will not exist.

src/Jobs/RemoveFromSearch.php
```diff
@@ -38,7 +38,9 @@ public function __construct($models)
     public function handle()
     {
         if ($this->models->isNotEmpty()) {
-            $this->models->first()->searchableUsing()->delete($this->models);
+            $this->models->first()->searchableUsing()->delete(
+                RemoveableScoutCollection::make($this->models)
+            );
```
stevebauman, line 43: We can't make this change. The `RemoveableScoutCollection` instance must be present in the constructor for the `getQueueableIds()` method to return the proper scout keys. This collection method will be called when the job is serialized and placed onto the queue:

https://github.com/laravel/scout/blob/3352fbc5c071cbd3d5392222363f06cf660ca003/src/Jobs/RemoveableScoutCollection.php#L15

PR #656, opened 2022-09-30
[9.x] Fix custom scout keys not being utilized when deleting from queue

### LINKED ISSUE #653: MeiliSearchEngine does not work with custom Scout keys
- Scout Version: 9.4.10
- Scout Driver: MeiliSearch
- Laravel Version: 9.21.6
- PHP Version: 8.1.9
- Database Driver & Version: MySQL 8.0
- SDK Version (If using a third-party service): N/A
- Meilisearch CLI Version (If using Meilisearch): meilisearch-http 0.28.0

### Description:

When using custom index keys with `getScoutKey()` in a Searchable model the MeiliSearchEngine does not correct add the custom key to index nor delete entries from the index.

1. The `update()` method in the MeiliSearchEngine class uses array_merge() to merge the Scout key index with the toSearchableArray() data but references the Scout key first so the custom key gets overwritten by $searchableData unless a custom key is used.

```php
           return array_merge(
               // [$model->getKeyName() => $model->getScoutKey()],
                $searchableData,
                $model->scoutMetadata(),
                [$model->getKeyName() => $model->getScoutKey()]
            );
```

2. RemoveFromSearch job makes a RemoveableScoutCollection which runs the models through `getScoutKey()` and passes the result to the the `delete()` method in MeiliSearchEngine which then runs `getScountKey()` a second time.
ie.
```php
    /**
     * Remove the given model from the index.
     *
     * @param  \Illuminate\Database\Eloquent\Collection  $models
     * @return void
     */
    public function delete($models)
    {
        $index = $this->meilisearch->index($models->first()->searchableAs());

        $index->deleteDocuments(
            // $models->map->getScoutKey()
            //     ->values()
            //     ->all()
            $models->pluck($models->first()->getKeyName())->toArray()
        );
    }
```
PS. We may also want to look at replacing getKeyName() with getScoutKeyName() in this MeiliSearchEngine class?

### Steps To Reproduce:

1. Create a searchable model which overrides the `getScoutKey` method to prepend a something...
```php
    /**
     * Get the custom search index key
     *
     * @return void
     */
    public function getScoutKey()
    {
        return 1 . $this->getKey();
    }
```
3. Create and `->save()` the model, notice in Meilisearch the ID is not prepended with 1.

4. Delete the model, notice the model is not removed from Meilisearch.

stevebauman:
Closes #655 
Closes #653 

This PR resolves an issue described in #653 where the `MeiliSearchEngine` (as well as the `AlgoliaEngine`, it turns out) does not utilize the custom scout key when removing models from the search index when deleting from the queued `RemoveFromSearch` job.

The issue, is that when the faux model instances (faux, because they don't actually exist in the database) are instantiated from the `RemoveFromSearch` job after *unserialization*, all of the model's attributes are (of course) not available. The models are instead force-filled with the *restored from queue* custom scout keys:

https://github.com/laravel/scout/blob/3352fbc5c071cbd3d5392222363f06cf660ca003/src/Jobs/RemoveFromSearch.php#L59-L65

This means, that the `getScoutKey()` method on these faux model instances will not return the appropriate keys inside of the `delete()` methods of their respective `Engine` when they are retrieved, as they will be overridden, and all other model attributes will be missing. 

Another issue, was that using a custom scout key that returned as a string, would interfere with Eloquent's primary key type casting. If the primary key of the searchable model matches the unqualified scout key (which is likely always the case), then Eloquent would attempt to cast the string as an integer when retrieved from the model naturally. This would result in `getScoutKey()` returning `0` if the returned custom key is prefixed with a string.

To resolve this, I've implemented a check to determine if the Scout key is a string, and alter the primary key type of the faux model instance to suit, so this issue no longer occurs:

https://github.com/stevebauman/scout/blob/8b9270f812e8de28e6371ecd5ecb20e61f491116/src/Jobs/RemoveFromSearch.php#L59-L68

```php
return tap(new $value->class, function ($model) use ($id) {
    // The scout key may not be an integer. In this case,
    // we will force a key type of string so it is not
    // cast when retrieving it from the model.
    $model->setKeyType(
        is_string($id) ? 'string' : 'int'
    )->forceFill([
        $model->getUnqualifiedScoutKeyName() => $id
    ]);
});
```

@DarronEngelbrechtEdge This should resolve your issue -- can you review this PR? I've included tests that cover your use case.

Let me know if you'd like anything adjusted, thanks for your time! ❤️


driesvints:
Thanks a lot @stevebauman! Great work.

Did you keep in mind the changes that are present on master? Do we need to adjust a lot when merging this into master later on? 

I'd love a review by you @mmachatschek if you're willing.

stevebauman:
Happy to help @driesvints! 🙏 

> Did you keep in mind the changes that are present on master? 

I didn't -- let me checkout master and see what the differences are and I'll comment back when I'm done!

DarronEngelbrechtEdge:
@stevebauman - Many thanks this PR seems like it will sort all uses cases and both engines. I will have time again on Monday to test it in my project and confirm that it's working for me.

src/Engines/MeiliSearchEngine.php
```diff
@@ -64,9 +64,9 @@ public function update($models)
             }
 
             return array_merge(
-                [$model->getKeyName() => $model->getScoutKey()],
                 $searchableData,
-                $model->scoutMetadata()
+                $model->scoutMetadata(),
+                [$model->getKeyName() => $model->getScoutKey()],
```
DarronEngelbrechtEdge, line 69: @stevebauman - PS. If people are already adding the key in their toSearchableArray() with some value other than what's returned from `getScoutKey()`. Worth considering perhaps? Might also then want to make this change in the MeiliSearchEngine to ensure the `objectID` key is not overridden?

DarronEngelbrechtEdge (commented):
Thanks!

DarronEngelbrechtEdge:
@stevebauman  - Thanks again I can confirm this resolves issue #653 in my project (using MeiliSearchEngine).

src/Engines/MeiliSearchEngine.php
mmachatschek: this is also something that jumped to my eye. The order here was definitly on purpose the way it was before. is there a specific reason why this was change?

driesvints: @stevebauman can you maybe answer this question?

stevebauman: Sorry for the late reply on this one guys! I modified this due to conflicts with the searchable data array. The priority here was intended.

`$searchableData` will contain the model's attributes, including its primary key, alongside other data:

```php
['id' => 1, '...']
```

This means, the custom Scout key will always be overridden (unless the model uses a different primary key). 

Without this, the `array_merge()` structure would be:

```php
array_merge(
    ['id' => 'custom-scout-key:1'],
    ['id' => 1],
    ['metadata'],
)
```

Leading to the custom scout key always being overridden by the primary key.

> @DarronEngelbrechtEdge PS. If people are already adding the key in their toSearchableArray() with some value other than what's returned from getScoutKey(). Worth considering perhaps?

I'm not really sure how we can account for that due to the above scenario...

> Might also then want to make this change in the MeiliSearchEngine to ensure the `objectID` key is not overridden?

Don't you specify the primary key with the MeilisearchEngine? I can't find any reference to objectID inside of Meilisearch's documentation.

stevebauman:
That's excellent to hear @DarronEngelbrechtEdge! Thank you for taking the time to review the PR and testing it out ❤️

src/Engines/MeiliSearchEngine.php
stevebauman: The implementations the above priority for the `MeiliSearchEngine` and `AlgoliaEngine` differ, as `Algolia` requires a primary key of `objectID`. You cannot change it (to my knowledge).

Since `objectID` is unlikely to be a models primary key, collisions in the `array_merge()` are unlikely, so I haven't adjusted it. Maybe we should to ensure the same priority consistency across the included engines?

**AlgoliaEngine**:

https://github.com/laravel/scout/blob/3352fbc5c071cbd3d5392222363f06cf660ca003/src/Engines/AlgoliaEngine.php#L194


**MeiliSearchEngine**:

https://github.com/laravel/scout/blob/3352fbc5c071cbd3d5392222363f06cf660ca003/src/Engines/MeiliSearchEngine.php#L266

DarronEngelbrechtEdge: > I'm not really sure how we can account for that due to the above scenario... The primary key will likely always be present, so the custom Scout key will never be included in the indexed data. Correct me if I'm wrong?

@stevebauman - You're correct, the array_merge() must decide if it wants to override the `toSearchableArray()` or use the value from there if the user has specified the key returned by `getScoutKeyName()`.

I am in favor of your method change which will override `toSearchableArray()` as this will ensure the rest of the functionality will operate on the correct ID giving the user some protection against bad configurations? Either way it might be worth a mention somewhere in the docs if it's not there already? 

> Don't you specify the primary key with the MeilisearchEngine? I can't find any reference to objectID inside of Meilisearch's documentation.

@stevebauman - Sorry noticed a typo in my previous suggestion, I meant to make the change in the AlgoliaEngine to be sure that the user can't override the `objectID` key in the `toSearchableArray()` but can only modify the value with `getScoutKey()` as intended.

DarronEngelbrechtEdge:
> That's excellent to hear @DarronEngelbrechtEdge! Thank you for taking the time to review the PR and testing it out heart

@stevebauman Only a pleasure, thanks to you lot for saving so many people so much time!

driesvints:
So where are we with this? Is this ready for review?

src/Engines/MeiliSearchEngine.php
stevebauman: > Sorry noticed a typo in my previous suggestion, I meant to make the change in the AlgoliaEngine to be sure that the user can't override the objectID key in the toSearchableArray() but can only modify the value with getScoutKey() as intended.

No worries! Understood, I agree with this change. I've moved the order of this in the `AlgoliaEngine` as well. 👍

stevebauman:
Ready for review now @driesvints! I've made one last change in regards to @DarronEngelbrechtEdge comment. I've moved the `array_merge()` priority for the `AlgoliaEngine` as well for unification across the engines:

https://github.com/laravel/scout/pull/656/commits/3fdb281699430a16a065519d73987628c5d870ab

```php
return array_merge(
    $searchableData,
    $model->scoutMetadata(),
    ['objectID' => $model->getScoutKey()], // Moved below so it cannot be overridden.
);
```

driesvints:
Thanks @stevebauman this looks good. @mmachatschek does this also look good to you?

mmachatschek (commented):
@driesvints The changes look good to me after @stevebauman [answered my concerns](https://github.com/laravel/scout/pull/656#discussion_r985928157)

driesvints (approved):
Thanks everyone for your work/feedback on this!

taylorotwell:
Thanks 👍

PR #657, opened 2022-09-30
[10.x] Fix custom scout keys not being utilized when deleting from queue

stevebauman:
Related #656 

Sending this in for tests to run and ensure everything passes with the recent changes of the above PR. 👍


driesvints:
Thanks @stevebauman. I'll use this PR as a reference once 9.x is updated and I merge it into master.

driesvints:
@stevebauman @mmachatschek I've merged 9.x into master now but unsure if everything is now okay. If you have the time, can you maybe check if everything looks in order? Looks like this PR is also conflicting now.

Here's my merge commit: https://github.com/laravel/scout/commit/0337e4c130f9eb37ef47b19627e36b6b29722655

stevebauman:
@driesvints Okay I've completed reviewing and have updated my PR with master.

I have removed the `getUnqualifiedScoutKeyName()` inside the `Searchable` trait. We no longer need it, because in v9, `getScoutKey()` was returning the fully qualified key name for the Eloquent model (i,e. `{table}.id`):

https://github.com/laravel/scout/blob/8489facee067ae656b7e309b07cd88eb17c1f5af/src/Searchable.php#L388-L391

In v10, this was adjusted to return only the key name instead (i.e. `id`):

https://github.com/laravel/scout/blob/0337e4c130f9eb37ef47b19627e36b6b29722655/src/Searchable.php#L390-L393

This should not cause any side-effects for those upgrading to v10. `getScoutKeyName()` was only used in one location in v9:

https://github.com/laravel/scout/blob/8489facee067ae656b7e309b07cd88eb17c1f5af/src/Searchable.php#L247-L249

**However**, if a developer has overrode `getScoutKeyName()` on their `Searchable` model with a hard-coded string matching the previous format for some reason (the fully qualified naming structure), such as:

```php
// app/Models/Post.php

{
    return 'posts.id';
}
```

Then they must update their formatting to return only the DB column itself, excluding the DB table to maintain compatibility with the v10 update:

```diff
public function getScoutKeyName()
{
-    return 'posts.id';
+    return 'id';
}
```

Let me know if there's anything else I can help with! 🙏

driesvints (approved):
This looks good @stevebauman. Thank you for your work on this!

taylorotwell:
Hi @stevebauman - I see this is sent to `master` branch. Can you elaborate on any breaking changes or notes we would need to make in the upgrade guide about this? Basically any relevant notes to what this changes would be helpful to document in this thread.

stevebauman:
Hey @taylorotwell, I can definitely do that, but I'm leaving my home for the weekend. I'll be able to write something up on Monday! 👍

taylorotwell:
OK - mark as ready for review when that is done. Thanks!

stevebauman:
Okay I've done a run though. There are two breaking changes here:

1. Removal of the `Searchable::getUnqualifiedScoutKeyName()` method.

This method was introduced in v9 in a PR I wrote (https://github.com/laravel/scout/pull/656) to prevent logic duplication of retrieving the key name from the fully qualified key name that was being returned from the `getScoutKeyName()` method.

2. Change of the `Searchable::getScoutKeyName()` return value.

The `getScoutKeyName()` method now returns the _unqualified_ key name from the model by default, instead of the _fully_ qualified name that includes the model's database table. As mentioned above, developers who have overridden the method will need to update their definition to return the unqualified key name to maintain compatibility with v10:

```diff
public function getScoutKeyName()
{
-    return 'posts.id';
+    return 'id';
}
```

This is because this method is now being used in the built-in search engine implementations (such as to store the models primary key in the indexed record), where it was not being used previously. Ex:

**MeiliSearch Before**:
https://github.com/laravel/scout/blob/2bc384eb383330dd762efb7a7c4923bd0bc5b7fd/src/Engines/MeiliSearchEngine.php#L69

**MeiliSearch After**:
https://github.com/laravel/scout/blob/0337e4c130f9eb37ef47b19627e36b6b29722655/src/Engines/MeiliSearchEngine.php#L69

Developers that have not overridden the `getScoutKeyName()` method, or have not utilized the `getUnqualifiedScoutKeyName()` method, will not have to adjust anything.

Let me know if you have any further questions or would like me to tackle additional related tasks!

PR #666, opened 2022-11-10
fix getTotalCount for meilisearch

Falki14:
getTotalCount fix for meilisearch
nbHits has be changed to estimatedTotalHits in meilisearch

### DIFF
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -314,7 +314,7 @@ class MeiliSearchEngine extends Engine
      */
     public function getTotalCount($results)
     {
-        return $results['nbHits'];
+        return $results['estimatedTotalHits'];
     }
 
     /**
--- a/tests/Feature/BuilderTest.php
+++ b/tests/Feature/BuilderTest.php
@@ -107,7 +107,7 @@ class BuilderTest extends TestCase
                     'name' => $result->name,
                 ];
             }),
-            'nbHits' => $query->count(),
+            'estimatedTotalHits' => $query->count(),
         ]);
     }
 }
--- a/tests/Unit/MeiliSearchEngineTest.php
+++ b/tests/Unit/MeiliSearchEngineTest.php
@@ -146,7 +146,7 @@ class MeiliSearchEngineTest extends TestCase
             'hits' => [],
             'offset' => 0,
             'limit' => 20,
-            'nbHits' => 0,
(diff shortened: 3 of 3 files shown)


Falki14:
Yes it will be break all meilisearch versions older then 0.28. This was a breaking change on version 0.28. https://github.com/meilisearch/meilisearch/releases/tag/v0.28.0

Last version now 0.29.1.

mmachatschek:
> @mmachatschek can you confirm this? Will this break stuff for people on lower meilisearch versions?

@driesvints the nbHits usage doesn't break existing implementations as we have a specific fix on the `meilisearch/meilisearch-php` package on this:

https://github.com/meilisearch/meilisearch-php/blob/03b78fd2d8e2f1e2472907f1146cf11be1646b2f/src/Endpoints/Indexes.php#L210-L214

Anyhow @Falki14 it would make sense to send this to the 10.x branch instead as we could require a more recent version of the php-sdk there and drop support for all older meilisearch and meilisearch sdk versions.

driesvints:
Thanks @mmachatschek 

@Falki14 feel free to send this to master with passing tests. 

We should also think about maybe warning through an exception when people are using incompatible Meilisearch versions. Does Meilisearch PHP library offer a version we can check somewhere?

mmachatschek:
@driesvints the sdk outputs a semver compatible version. we already use this for another workaround regarding the filter attribute.

The next release of meilisearch (0.30) will have a new pagination api which we could also use for this. after that the meilisearch API is pretty stable and we can expect them to stick to the naming schema.
Because of that I suggest we remove the hotfixes entirely and require only newer version of the sdk on the 10.x branch.
A PR to switch to the pagination api as soon as v0.30 is release, will be sent right after the SDK supports it

driesvints:
Cool, yes we can focus Scout v10 for Meilisearch v1. 

I do wonder how we're going to handle this in the meantime because afaik this means people won't be able to use Meilisearch >= v0.30 for now?

curquiza:
> I do wonder how we're going to handle this in the meantime because afaik this means people won't be able to use Meilisearch >= v0.30 for now?

People can still use Meilisearch v0.30.0 and later the same way they used it before (with `estimatedTotalHits`, `offset` and `limit`). The new way of handling pagination is an addition, not a replacement. So they still can use Meilisearch v0.30.0 and later with Laravel 9.

driesvints:
Cool 👍

PR #669, opened 2022-11-23
[9.x] Support Meilisearch index settings

driesvints:
This PR adds support for Meilisearch's index settings: 

https://docs.meilisearch.com/learn/configuration/settings.html
https://docs.meilisearch.com/reference/api/settings.html

Through a new `settings` key on the `scout.meilisearch` config option, users will be able to define all the settings they need for their model indexes. These settings will be synced when the index is manually created through the `scout:index` command ~or whenever models get updated through the engine~ or when they use the new `scout:sync-settings` command.

```php
'meilisearch' => [
    'host' => env('MEILISEARCH_HOST', 'http://localhost:7700'),
    'key' => env('MEILISEARCH_KEY', null),
    'settings' => [
        'users' => [
            'filterableAttributes'=> ['id', 'name', 'email'],
        ],
    ],
],
```

### DIFF
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -26,25 +26,16 @@ class MeiliSearchEngine extends Engine
     protected $softDelete;
 
     /**
-     * The index settings.
-     *
-     * @var array
-     */
-    protected $settings;
-
-    /**
      * Create a new MeiliSearchEngine instance.
      *
      * @param  \MeiliSearch\Client  $meilisearch
      * @param  bool  $softDelete
-     * @param  array  $settings
      * @return void
      */
-    public function __construct(MeiliSearchClient $meilisearch, $softDelete = false, array $settings = [])
+    public function __construct(MeiliSearchClient $meilisearch, $softDelete = false)
     {
         $this->meilisearch = $meilisearch;
         $this->softDelete = $softDelete;
-        $this->settings = $settings;
     }
 
     /**
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -54,6 +54,10 @@ class MeiliSearchEngine extends Engine
 
         $index = $this->meilisearch->index($models->first()->searchableAs());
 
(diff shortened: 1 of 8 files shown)


driesvints:
@mmachatschek I'd appreciate a review here if you're up for it : )

src/Engines/MeiliSearchEngine.php
```diff
@@ -54,6 +63,10 @@ public function update($models)
 
         $index = $this->meilisearch->index($models->first()->searchableAs());
 
+        if ($settings = $this->settings[$models->first()->searchableAs()] ?? []) {
+            $index->updateSettings($settings);
+        }
+
```
mmachatschek, line 69: I don't think this would be super beneficial to users. Yes, this would keep index settings up to date. BUT everytime you update a model, this will add a update task to the meilisearch queue too (if there were no changes to the settings, there are would be no consequences and no performance degradation for meili users).

Still every model update will create the following tasks in meilisearch:

- update index settings
- update model/s

In any case, I would recommend everyone that deploys meilisearch to production with Laravel Scout to create a custom command that handles updates to the index settings (just like any migration command for the database).

mmachatschek (commented):
The code looks good to me 👍 I'm just concerned about one change:

src/Engines/MeiliSearchEngine.php
driesvints: Thanks @mmachatschek. You highlighted the one thing I was concerned about. I'll think about this a bit more and perhaps introduce a new command to update an index.

driesvints: I've added a new `scout:sync-settings` command that can be used during deployment. Can you review again?

mmachatschek: @driesvints this is a nice addition. thats basically the same code that I use in my environment. I didn't test your code but it looks rock solid 👍

mmachatschek (commented):
https://github.com/laravel/scout/pull/669#discussion_r1032408171

driesvints:
Thanks!

PR #670, opened 2022-12-06
[9.x] Fix the sync index settings command not using the scout prefix

tonysm:
### Changed

- Make the `scout:sync-index-settings` command aware of the scout prefix being used.

---

I was able to work around this by assigning the scout prefix to a variable inside the config file, like so:

```php
return [
    'prefix' => $scoutPrefix = env('SCOUT_PREFIX', 'local_'),

    // ...

    'meilisearch' => [
        'host' => env('MEILISEARCH_HOST', 'http://localhost:7700'),
        'key' => env('MEILISEARCH_KEY', null),
        'index-settings' => [
            $scoutPrefix . 'users' => [
                'filterableAttributes'=> [
                    '__soft_deleted',
                    'id',
                    'name',
                ],
                'sortableAttributes' => [
                    'id',
                ],
            ],
        ],
    ],
];
```

But I figured the command could be made aware of the prefix instead. This change also allows defining the key as a model, so something like:

```php
return [
    'meilisearch' => [
        'host' => env('MEILISEARCH_HOST', 'http://localhost:7700'),
        'key' => env('MEILISEARCH_KEY', null),
        'index-settings' => [
            App\Models\User::class => [
                'filterableAttributes'=> [
                    '__soft_deleted',
                    'id',
                    'name',
                ],
                'sortableAttributes' => [
                    'id',
                ],
            ],
        ],
    ],
];
```

Which uses the `searchableAs()` method inside the `Searchable` trait, which is already aware of the scout prefix.

When a string is passed instead of an FQCN, it will check if it already starts with the prefix before prepending it.

### DIFF
--- a/src/Console/SyncIndexSettingsCommand.php
+++ b/src/Console/SyncIndexSettingsCommand.php
@@ -64,6 +64,6 @@ class SyncIndexSettingsCommand extends Command
 
         $prefix = config('scout.prefix');
 
-        return ! Str::startsWith($name, $prefix) ? $prefix . $name : $name;
+        return ! Str::startsWith($name, $prefix) ? $prefix.$name : $name;
     }
 }
--- a/src/Console/SyncIndexSettingsCommand.php
+++ b/src/Console/SyncIndexSettingsCommand.php
@@ -4,6 +4,7 @@ namespace Laravel\Scout\Console;
 
 use Exception;
 use Illuminate\Console\Command;
+use Illuminate\Support\Str;
 use Laravel\Scout\EngineManager;
 
 class SyncIndexSettingsCommand extends Command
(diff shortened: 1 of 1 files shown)


driesvints (approved):
Thanks @tonysm!

driesvints:
I released this as v9.5.1

tommy66374:
I want upgrade to Scout v9.5.1

so we need change to use

```
        'index-settings' => [
            App\Models\User::class => [
```

instead of hardcoded the index name ?

Post Model

```
    public function searchableAs()
    {
        return 'posts_index';
    }
```

config\scout.php

```
        'index-settings' => [
            'posts_index' => [
```

driesvints:
@tommy66374 the old way should still work.

tommy66374:
OK, Thank you very much, this very cool

PR #675, opened 2022-12-15
[10.x] Drop old PHP and Laravel versions

driesvints:
Drops support for PHP 7.3 & 7.4 and Laravel 8.x

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,8 +16,11 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [8.1, 8.2]
+        php: ['8.0', 8.1, 8.2]
         laravel: [9]
+        exclude:
+          - php: '8.0'
+            laravel: 9
 
     name: PHP ${{ matrix.php }} - Laravel ${{ matrix.laravel }}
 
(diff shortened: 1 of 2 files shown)


driesvints:
@mmachatschek do you know when meilisearch-php v1 hits? Depending on that we might hold off on releasing v10 of Scout until that happens but if it's too far in the future it might not be worth it.

mmachatschek:
@driesvints the v1 milestone is due 2022-02-06
https://github.com/meilisearch/meilisearch/milestone/37

most sdks are also ready by that time

I suppose most api changes were made with the current v0.30 release and therefore the most recent SDK version would be sufficient I guess.

Maybe @curquiza can add a comment what the major changes for the v1 release for the SDKs would be

driesvints:
> Edit: I think its worth dropping all workarounds on the 9.x/10.x branch that were implemented to support most meilisearch versions from >0.19

On 9.x we probably cannot do this but we can on 10.x. Do you maybe want to send in a PR that removes those? You can up the minimum version also to 0.30

mmachatschek:
> On 9.x we probably cannot do this but we can on 10.x. Do you maybe want to send in a PR that removes those? You can up the minimum version also to 0.30

I'll send a PR ASAP (probably tomorrow)

driesvints:
Thanks @mmachatschek!

driesvints:
@taylorotwell you can merge this one already.

PR #678, opened 2022-12-16
[10.x] Meilisearch v1 support 

mmachatschek:
- [x] Notable changes for the migration guide
  - [x] casing of the the client in the service container has been renamed (use the correct `\Meilisearch\Client`)
    - Reference: https://github.com/meilisearch/meilisearch-php/releases/tag/v0.27.0
  - [x] minimum required meilisearch-php library version v1
  - [x] Usage of the new pagination implementation of meilisearch (page/hitsPerPage) which makes the search results exhausitve instead of estimated
    - Reference: https://docs.meilisearch.com/learn/advanced/pagination.html#numbered-page-selectors 
    - Reference: https://github.com/meilisearch/meilisearch/releases/tag/v0.30.0
  - [x] No workarounds for earlier meilisearch-php library versions (`filters => filter`)
    - Reference: https://github.com/meilisearch/meilisearch/releases/tag/v0.21.0
  - [x] Also consult the changelog on https://github.com/meilisearch/meilisearch-php for any breaking changes if you are customizing engine searches like so https://laravel.com/docs/9.x/scout#customizing-engine-searches
    - Reference: https://github.com/meilisearch/meilisearch-php/releases

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -29,17 +29,17 @@ PR: https://github.com/laravel/scout/pull/657
 
 Due to the `getScoutKeyName` change discussed above, the `getUnqualifiedScoutKeyName` method was removed as it is no longer necessary.
 
-### Meilisearch v1.0 Changes
+### Meilisearch 1.0 Changes
 
-Scout v10 now supports Meilisearch PHP v1.0 as its new minimum version. You should upgrade your dependency:
+Scout 10.x requires Meilisearch PHP 1.0 as its minimum supported SDK version; therefore, you should upgrade your dependency via your application's `composer.json` file:
 
 ```json
 "meilisearch/meilisearch-php": "^1.0",
 ```
 
-Further more, all casing has been changed from the old `MeiliSearch` to the new `Meilisearch`. This means that all references to Meilisearch code you have in your codebase should adopt this new casing.
+In this SDK update, all namespace and class references to "MeiliSearch" have been updated to "Meilisearch". Please review your code for any reference to the old capitalization and update those references accordingly.
 
-Please see [the full Meilisearch PHP v1.0 changelog](https://github.com/meilisearch/meilisearch-php/releases/tag/v1.0.0) for all changes.
+In addition, please consult the full [Meilisearch PHP v1.0 changelog](https://github.com/meilisearch/meilisearch-php/releases/tag/v1.0.0) for further details.
 
 ## Upgrading To 9.0 From 8.x
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -29,6 +29,18 @@ PR: https://github.com/laravel/scout/pull/657
 
 Due to the `getScoutKeyName` change discussed above, the `getUnqualifiedScoutKeyName` method was removed as it is no longer necessary.
 
+### Meilisearch v1.0 Changes
+
+Scout v10 now supports Meilisearch PHP v1.0 as its new minimum version. You should upgrade your dependency:
+
+```json
+"meilisearch/meilisearch-php": "^1.0",
+```
+
+Further more, all casing has been changed from the old `MeiliSearch` to the new `Meilisearch`. This means that all references to Meilisearch code you have in your codebase should adopt this new casing.
+
+Please see [the full Meilisearch PHP v1.0 changelog](https://github.com/meilisearch/meilisearch/releases/tag/v1.0.0) for all changes.
+
 ## Upgrading To 9.0 From 8.x
 
 ### Minimum Laravel Version
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -39,7 +39,7 @@ Scout v10 now supports Meilisearch PHP v1.0 as its new minimum version. You shou
 
 Further more, all casing has been changed from the old `MeiliSearch` to the new `Meilisearch`. This means that all references to Meilisearch code you have in your codebase should adopt this new casing.
 
-Please see [the full Meilisearch PHP v1.0 changelog](https://github.com/meilisearch/meilisearch/releases/tag/v1.0.0) for all changes.
+Please see [the full Meilisearch PHP v1.0 changelog](https://github.com/meilisearch/meilisearch-php/releases/tag/v1.0.0) for all changes.
 
 ## Upgrading To 9.0 From 8.x
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -29,7 +29,7 @@ PR: https://github.com/laravel/scout/pull/657
 
 Due to the `getScoutKeyName` change discussed above, the `getUnqualifiedScoutKeyName` method was removed as it is no longer necessary.
 
-### Meilisearch 1.0 Changes
+### Meilisearch 1.0
 
 Scout 10.x requires Meilisearch PHP 1.0 as its minimum supported SDK version; therefore, you should upgrade your dependency via your application's `composer.json` file:
 
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "illuminate/support": "^9.0|^10.0"
     },
     "require-dev": {
-        "meilisearch/meilisearch-php": "^0.23",
+        "meilisearch/meilisearch-php": "^1.0",
         "mockery/mockery": "^1.0",
         "php-http/guzzle7-adapter": "^1.0",
         "orchestra/testbench": "^7.0|^8.0",
@@ -52,7 +52,7 @@
     },
     "suggest": {
         "algolia/algoliasearch-client-php": "Required to use the Algolia engine (^3.2).",
-        "meilisearch/meilisearch-php": "Required to use the MeiliSearch engine (^0.23)."
+        "meilisearch/meilisearch-php": "Required to use the Meilisearch engine (^1.0)."
     },
     "config": {
         "sort-packages": true,
--- a/src/EngineManager.php
+++ b/src/EngineManager.php
@@ -10,9 +10,10 @@ use Illuminate\Support\Manager;
 use Laravel\Scout\Engines\AlgoliaEngine;
 use Laravel\Scout\Engines\CollectionEngine;
 use Laravel\Scout\Engines\DatabaseEngine;
-use Laravel\Scout\Engines\MeiliSearchEngine;
+use Laravel\Scout\Engines\MeilisearchEngine;
 use Laravel\Scout\Engines\NullEngine;
-use MeiliSearch\Client as MeiliSearch;
+use Meilisearch\Client as MeilisearchClient;
+use Meilisearch\Meilisearch;
 
 class EngineManager extends Manager
 {
@@ -74,10 +75,10 @@ class EngineManager extends Manager
         }
 
         if (class_exists('AlgoliaSearch\Client')) {
-            throw new Exception('Please upgrade your Algolia client to version: ^2.2.');
+            throw new Exception('Please upgrade your Algolia client to version: ^3.2.');
         }
 
-        throw new Exception('Please install the Algolia client: algolia/algoliasearch-client-php.');
+        throw new Exception('Please install the suggested Algolia client: algolia/algoliasearch-client-php.');
     }
 
     /**
@@ -107,38 +108,34 @@ class EngineManager extends Manager
     }
 
     /**
-     * Create an MeiliSearch engine instance.
+     * Create a Meilisearch engine instance.
      *
-     * @return \Laravel\Scout\Engines\MeiliSearchEngine
+     * @return \Laravel\Scout\Engines\MeilisearchEngine
      */
     public function createMeilisearchDriver()
     {
-        $this->ensureMeiliSearchClientIsInstalled();
+        $this->ensureMeilisearchClientIsInstalled();
 
-        return new MeiliSearchEngine(
-            $this->container->make(
-                class_exists(MeiliSearch::class)
-                    ? MeiliSearch::class
-                    : \Meilisearch\Client::class
-            ),
+        return new MeilisearchEngine(
+            $this->container->make(MeilisearchClient::class),
             config('scout.soft_delete', false)
         );
     }
 
     /**
-     * Ensure the MeiliSearch client is installed.
+     * Ensure the Meilisearch client is installed.
      *
      * @return void
      *
      * @throws \Exception
      */
-    protected function ensureMeiliSearchClientIsInstalled()
+    protected function ensureMeilisearchClientIsInstalled()
     {
-        if (class_exists(MeiliSearch::class) || class_exists(\Meilisearch\Client::class)) {
+        if (class_exists(Meilisearch::class) && version_compare(Meilisearch::VERSION, '1.0.0') >= 0) {
             return;
         }
 
-        throw new Exception('Please install the MeiliSearch client: meilisearch/meilisearch-php.');
+        throw new Exception('Please install the suggested Meilisearch client: meilisearch/meilisearch-php.');
     }
 
     /**
--- a/src/Engines/MeilisearchEngine.php
+++ b/src/Engines/MeilisearchEngine.php
@@ -113,6 +113,7 @@ class MeilisearchEngine extends Engine
 
     /**
      * Perform the given search on the engine.
+     *
      * page/hitsPerPage ensures that the search is exhaustive.
      *
      * @param  int  $perPage
--- a/src/Engines/MeilisearchEngine.php
+++ b/src/Engines/MeilisearchEngine.php
@@ -5,15 +5,17 @@ namespace Laravel\Scout\Engines;
 use Illuminate\Support\LazyCollection;
 use Laravel\Scout\Builder;
 use Laravel\Scout\Jobs\RemoveableScoutCollection;
-use MeiliSearch\MeiliSearch;
-use MeiliSearch\Search\SearchResult;
+use Meilisearch\Client as MeilisearchClient;
+use Meilisearch\Contracts\IndexesQuery;
+use Meilisearch\Meilisearch;
+use Meilisearch\Search\SearchResult;
 
-class MeiliSearchEngine extends Engine
+class MeilisearchEngine extends Engine
 {
     /**
-     * The MeiliSearch client.
+     * The Meilisearch client.
      *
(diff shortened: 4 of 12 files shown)


mmachatschek:
@driesvints I'll bump you for an initial look at the changes

also maybe @curquiza or @brunoocasali can take a look too (the changes also have something to do with https://github.com/meilisearch/meilisearch-php/pull/431)

driesvints:
@mmachatschek we're going to postpone the entire v10 release until Meilisearch PHP v1 is released but that's under the idea that it'll happen in early February. We don't want to wait longer than that.

brunoocasali:
Hi @mmachatschek, I approved and released the latest version https://github.com/meilisearch/meilisearch-php/releases/tag/v0.26.1

@driesvints our plan is to release a v1 of Meilisearch PHP in 2023 after the Meilisearch v1 which will occur in February 6th, 2023. We are now, planning the first quarter of 2023 and I'll personally include the PHP release as a top priority.

driesvints:
Thank you @brunoocasali!

brunoocasali:
@mmachatschek, I want to ask you to add the analytics information like Algolia's implementation does, if @driesvints agrees, to meilisearch-php.

The information can be sent as an array of strings like this:

`['Meilisearch Laravel Scout (v10.0.0)']` there is a constructor argument you can send the information https://github.com/meilisearch/meilisearch-php/blob/main/src/Client.php#L52.

This information is essential for the Integrations Team to help maintain the integrations. By the way:

Analytics is enabled by default in the server, but you can disable them by following [this guide](https://docs.meilisearch.com/learn/what_is_meilisearch/telemetry.html#how-to-disable-data-collection)
Also, of course, every analytics data we collect are **ANONYMOUS** [read the guide for more information](https://docs.meilisearch.com/learn/what_is_meilisearch/telemetry.html). So the users have nothing to worry about it.

driesvints:
@brunoocasali good idea. Would be cool if we could do that. Is that something we can already tackle in v9?

brunoocasali:
> @brunoocasali good idea. Would be cool if we could do that. Is that something we can already tackle in v9?

Yes @driesvints! The analytics support started on Meilisearch v0.21, and Meilisearch-PHP added support for custom agents on v0.24.2.

driesvints:
In that case the only thing we’ll have to see is if we can add it on 9.x in a backwards compatible way so it doesn’t conflict with older meilisearch php installs that users might have.

mmachatschek:
@driesvints we can do that on the 9.x branch by making a version compare and then add the agent/or not if the version is lower than <0.24.2

composer.json
```diff
@@ -24,7 +24,7 @@
         "illuminate/support": "^9.0"
     },
     "require-dev": {
-        "meilisearch/meilisearch-php": "^0.23",
+        "meilisearch/meilisearch-php": ">=0.26.1 <1.0 | ^1.0",
```
mmachatschek, line 27: @driesvints I added this version constraint, so that tests are still possible. As soon as meilisearch and meilisearch-php hit v1. we could set this to `^1.0` only

src/EngineManager.php
```diff
@@ -107,34 +108,34 @@ protected function defaultAlgoliaHeaders()
...
      *
      * @return void
      *
      * @throws \Exception
      */
-    protected function ensureMeiliSearchClientIsInstalled()
+    protected function ensureMeilisearchClientIsInstalled()
     {
-        if (class_exists(MeiliSearch::class)) {
+        if (version_compare(Meilisearch::VERSION, '0.26.1') >= 0) {
```
mmachatschek, line 134: @driesvints I added this version constraint, so that tests are still possible. As soon as meilisearch and meilisearch-php hit v1. we could set this to `^1.0` only

src/Engines/MeilisearchEngine.php
```diff
@@ -314,7 +308,10 @@ public function lazyMap(Builder $builder, $results, $model)
      */
     public function getTotalCount($results)
     {
-        return $results['nbHits'];
+        // It may happen, that users send search requests to meilisearch with limit/offset
+        // totalHits exists as soon as the user performs a search with page/hitsPerPage
+        // totalHits is an exhausitive value
+        return $results['totalHits'] ?? $results['estimatedTotalHits'];
```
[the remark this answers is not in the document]
mmachatschek: I checked the code. there should not be the possibility to have a limit/offset search (except you use the meilisearch client directly, but the method will always receive results from a paginated search).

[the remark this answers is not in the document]
mmachatschek: I checked the code. The `public function getTotalCount($results)` method on the engine is only called in the context of pagination (which will always force the parameters page, hitsPerPage -> and therefore return meilisearch paginated results)

There may be users that apply custom logic and access the engine themselves. @driesvints maybe we could encourage users to use page/hitsPerPage parameters in there customized engine searches https://laravel.com/docs/9.x/scout#customizing-engine-searches via the upgrade guide?

src/Engines/MeilisearchEngine.php
```diff
@@ -138,13 +139,6 @@ protected function performSearch(Builder $builder, array $searchParams = [])
...
-        // meilisearch-php 0.19.0 is compatible with meilisearch server 0.21.0...
-        if (version_compare(MeiliSearch::VERSION, '0.19.0') >= 0 && isset($searchParams['filters'])) {
-            $searchParams['filter'] = $searchParams['filters'];
-
-            unset($searchParams['filters']);
-        }
-
         if ($builder->callback) {
             $result = call_user_func(
                 $builder->callback,
```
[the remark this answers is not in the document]
mmachatschek: @brunoocasali I dont remember the exact logic behind using rawSearch but as far as I can tell, all methods expect a plain PHP array and therefore this was the way.

furthermore users can modify the query and apply custom logic https://laravel.com/docs/9.x/scout#customizing-engine-searches if a user returns a `SearchResult` object by returning the result of `$meilisearchClient->search()` we read the `getRaw()` of the `SearchResult`object

additionally the `rawSearch` method has less overhead. If there is value in switching to `->search()` instead of `->rawSearch()` please tell me. for v1, I suppose there is no change in the result object so we can safely stick to the plain array.

the implementation tries to stick as close to the aloglia driver as possible which also returns a plain php array.

src/EngineManager.php
mmachatschek: upgraded the version to v1 now so this is rdy for when meilisearch-php v1 is released

composer.json
mmachatschek: upgraded the version to v1 now so this is rdy for when meilisearch-php v1 is released

mmachatschek:
@driesvints will you add the migration guide?

driesvints:
@mmachatschek could you give me a tl;dr of the breaking changes?

mmachatschek:
@driesvints is the migration guide list in the PR description enough? I will take another look on the changes and check if there are more breaking changes for users other than the ones I already listed

driesvints:
@mmachatschek yeah looks good probably. Thanks!

src/Engines/MeilisearchEngine.php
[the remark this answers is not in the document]
brunoocasali: The search methods will stay as they were during the v1 transition.

I'm taking this opportunity to clean up the repository. I'll list every possible breaking-change in the next changelog, but as far as I checked, nothing will impact the work of this branch.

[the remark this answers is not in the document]
mmachatschek: @brunoocasali as far as I can tell from the open PR https://github.com/meilisearch/meilisearch-php/pull/454 only the `deleteAllIndexes` impacts the code in the console commands of this repo.

I subscribed to your PR so I can update the code here as soon as that one was merged

src/Engines/MeilisearchEngine.php
```diff
@@ -378,7 +372,13 @@ public function deleteIndex($name)
      */
     public function deleteAllIndexes()
     {
-        return $this->meilisearch->deleteAllIndexes();
+        $tasks = [];
+        $indexes = $this->meilisearch->getIndexes();
```
brunoocasali, line 376: @mmachatschek, just to inform you, this does not remove all the indexes. It only removes the first page of indexes, which will probably be 20.

Since this method did not remove all the indexes correctly and PHP would be the only one with it, I opted to remove it.
A correct implementation in my POV is to correctly iterate over all the pages of the response :)

driesvints:
Seems like there's still work on this PR. Please ping me when it's ready and I'll write the upgrade guide then.

src/Engines/MeilisearchEngine.php
mmachatschek: I thought so, just read through the changelog of the SDK. at least the implementation of the method here on scout side behaves the same, although its wrong. I'll update the code to address this issue

mmachatschek: @brunoocasali the PR is ready for another round of review

src/Engines/MeilisearchEngine.php
```diff
@@ -378,7 +373,28 @@ public function deleteIndex($name)
...
+            $query->setOffset($offset)->setLimit($limit);
+
+            $indexes = $this->meilisearch->getIndexes($query);
+            foreach ($indexes->getResults() as $index) {
+                $tasks[] = $index->delete();
+            }
+
+            if ($indexes->count() < $limit) {
+                $hasMorePages = false;
+            }
```
pyrou, line 392: if number of search exactly match a multiple of $limit, this implementation would perform an unnecessary extra request.

This can be prevented with following additional check :

```suggestion
            if ($indexes->count() < $limit || count($tasks) === $indexes->getTotal()) {
                $hasMorePages = false;
            }
```

src/Engines/MeilisearchEngine.php
```diff
@@ -378,7 +373,28 @@ public function deleteIndex($name)
...
+
+            $offset += $limit;
```
pyrou, line 394: ⚠️ A big warning here

To explain the issue let's suppose $limit = 3; and you have 8 indexes in meilisearch named from A to H

A B C D E F G H

- A first request, you will delete A B C
- But here you changed the $offset to 3
- now meilisearch have index D E F G H

At the second request, you will actually delete G and H,
Now, because $indexes->count() (2) < 3, hasMorePages will becomes false,
indexes D E and F will remain on meilisearch

tests/Unit/MeilisearchEngineTest.php
```diff
@@ -117,12 +118,12 @@ public function test_search_sends_correct_parameters_to_meilisearch()
         $client = m::mock(Client::class);
         $client->shouldReceive('index')->with('table')->andReturn($index = m::mock(Indexes::class));
         $index->shouldReceive('search')->with('mustang', [
-            'filter' => 'foo=1 AND bar=2',
+            'filters' => 'foo=1 AND bar=2',
```
brunoocasali, line 121: It is supposed to be `"filter"` 🤔 

https://github.com/meilisearch/meilisearch-php/blob/19351d7ddff1cf66fe4db33bfdde55e601faa554/tests/Endpoints/SearchTest.php#L254

brunoocasali: There are more occurrences of this, if I'm correct.

brunoocasali (commented):
Hi @mmachatschek, lovely work!

I left some comments to check if we're in the same page 🥇

src/Engines/MeilisearchEngine.php
```diff
@@ -105,13 +106,14 @@ public function search(Builder $builder)
     {
         return $this->performSearch($builder, array_filter([
             'filters' => $this->filters($builder),
-            'limit' => $builder->limit,
+            'hitsPerPage' => $builder->limit,
```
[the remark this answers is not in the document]
mmachatschek: Users could still override the parameters `hitsPerPage/page` by customizing the search https://laravel.com/docs/9.x/scout#customizing-engine-searches but in any case, it should have always been exhaustive for the meilisearch engine in scout.

The algolia driver also uses page/hitsPerPage which makes the search exhaustive and is the recommended method for doing pagination.

This is basically the "fix" for previous issues where the pagination didn't display the correct pages as the total pages was an estimation by meilisearch

src/Engines/MeilisearchEngine.php
mmachatschek: Thats true, thanks for pointing this out 👍

brunoocasali (commented):
Looks good to me! 🎉

driesvints:
Thanks! I'll try to review this soon.

src/Engines/MeilisearchEngine.php
```diff
@@ -378,7 +373,26 @@ public function deleteIndex($name)
      */
     public function deleteAllIndexes()
     {
-        return $this->meilisearch->deleteAllIndexes();
+        $tasks = [];
+        $offset = 0;
```
pyrou, line 377: Final thought on this $offset,

This variable becomes useless, also, creating a new instance of $query isn't necessary anymore.

Here is a proposition of something possibly simpler

```php
public function deleteAllIndexes()
{
    $tasks = [];
    $batchSize = 20;

    $query = new IndexesQuery();
    $query->setLimit($batchSize);

    do {
        $indexes = $this->meilisearch->getIndexes($query);

        foreach ($indexes->getResults() as $index) {
            $tasks[] = $index->delete();
        }
    }
    while ($indexes->count() >= $batchSize);

    return $tasks;
 }
 ```

Comment (mmachatschek) [reply src/Engines/MeilisearchEngine.php:377]: I thought about this a bit more and came to the conclusion that either of our solutions don't work correctly.

Index deletion is an async task, and the method will not correctly work as meili's tasks queue may not have finished deleting the task when we hit the next iteration in the loop. So the code will try to delete the same index multiple times and spam the meili task queue.

I think it would be better to just set a very high batch size and just take what ever is returned on the first page? what do you think @brunoocasali?

ping @driesvints

Comment (pyrou) [reply src/Engines/MeilisearchEngine.php:377]: The same light comes to me this morning..

Are the order of index guaranteed ? either name, or creation date ?
If so, a way to address the async issue would be, to get the total number of index first (using `$indexes->getTotal()`). Then crawl page and delete per page but from the last page to the first.

By doing this, whenever the deletion is completed or not at next batch, you should not overlap anymore.

Maybe this is completely overkill, and just setting a very high batch size is the best way to go.

Comment (mmachatschek) [reply src/Engines/MeilisearchEngine.php:378]: Hi there, I kind of went with my first solution just fetching all indexes first and when we have all of them in memory I iterate over them.

A bigger batch size makes sense. I think the (deleteAllIndexes) feature as is, probably is completely sufficient.

Comment (mmachatschek) [reply src/Engines/MeilisearchEngine.php]: Thanks @brunoocasali for clearing this up. I went with high batch number that should be sufficient. Hopefully this is something that a dedicated endpoint on meilisearch will handle 👍

**Review by brunoocasali — COMMENTED** [review]
From my POV it is good, but I may lack more knowledge from the laravel side :)

**Comment by driesvints** [comment]
Cool. I just merged 9.x into 10.x so can you rebase with it? If that's done I'll start my review.

**Comment by mmachatschek** [comment]
ping @driesvints rebase done

File: src/Engines/MeilisearchEngine.php [hunk 11]
```diff
@@ -103,14 +105,15 @@ public function delete($models)
...
-            'limit' => $builder->limit,
+            'filter' => $this->filters($builder),
+            'hitsPerPage' => $builder->limit,
             'sort' => $this->buildSortFromOrderByClauses($builder),
         ]));
     }
 
     /**
      * Perform the given search on the engine.
+     * page/hitsPerPage ensures that the search is exhaustive.
```
driesvints, line 116:
```suggestion
     *
     * page/hitsPerPage ensures that the search is exhaustive.
```

driesvints (approved):

driesvints:
Code looks good. Thanks for all the work all! I'll write the migration guide soon before marking this as ready for review for Taylor.

driesvints:
> Usage of the new pagination implementation of meilisearch (page/hitsPerPage) which makes the search results exhausitve instead of estimated

@mmachatschek can you give examples of how end users are affected? What did they get before and what will they get now? What Scout code would be affected?

mmachatschek:
@driesvints 
> @mmachatschek can you give examples of how end users are affected? What did they get before and what will they get now?

The pagination used in the new implementation, ensures that users get the correct counts of totalHits and totalPages which leads to more accurate paginations. There will not be any inconsistency between totalHits count on the first page, and the pages that come after that.

It may have happened that meilisearch e.g. returned a number of 1.000 totalHits on the first page and 950 on the 10th page.
So the previous implementation tried to give a faster estimate of how many results are available but that could lead to different counts on the paginations.

> What Scout code would be affected?

I don't think there are any actions needed by end-users (that set limit/offset in a custom scout query) as meilisearch prefers the usage of (page/hitsPerPage) in the pagination that we call in the engine.

driesvints:
Cool, in that case I won't mention it in the upgrade guide.

Thanks a lot you all for all your work on this! @taylorotwell this is now ready for you to review. After merging this we can release Scout v10.

UPGRADE.md
```diff
@@ -29,6 +29,18 @@ PR: https://github.com/laravel/scout/pull/657
...
+
+Scout v10 now supports Meilisearch PHP v1.0 as its new minimum version. You should upgrade your dependency:
+
+```json
+"meilisearch/meilisearch-php": "^1.0",
+```
+
+Further more, all casing has been changed from the old `MeiliSearch` to the new `Meilisearch`. This means that all references to Meilisearch code you have in your codebase should adopt this new casing.
+
+Please see [the full Meilisearch PHP v1.0 changelog](https://github.com/meilisearch/meilisearch/releases/tag/v1.0.0) for all changes.
```
brunoocasali, line 42: I'm just fixing the link. The old one was pointing to the Meilisearch release, not PHP release.

```suggestion
Please see [the full Meilisearch PHP v1.0 changelog](https://github.com/meilisearch/meilisearch-php/releases/tag/v1.0.0) for all changes.
```

By the way, should we mention, the Meilisearch engine release release?

driesvints: > By the way, should we mention, the Meilisearch engine release release?

The what?

Issue #682, opened 2022-12-28
Support Search Parameters in Scout for MeiliSearch


lucasRolff:
It would be super nice to support the various Search Parameters from MeiliSearch, especially the matching strategy ( https://docs.meilisearch.com/reference/api/search.html#matching-strategy ) - in some cases the "last" strategy isn't ideal when one wants to do somewhat strict searches.

driesvints:
Heya, I sent in a PR. Can you try it out to see if that works for you and reply on the PR? https://github.com/laravel/scout/pull/683

PR #683, opened 2022-12-29
[9.x] Allow options for search requests

driesvints:
Closes https://github.com/laravel/scout/issues/682

This PR allows users to set extra options for their search request. Right now, Algolia and Meilisearch are supported.

```php
$results = User::search('hello world')->options([
    'matchingStrategy' => 'all',
])->get();
```

### DIFF
--- a/src/Engines/MeiliSearchEngine.php
+++ b/src/Engines/MeiliSearchEngine.php
@@ -145,6 +145,8 @@ class MeiliSearchEngine extends Engine
             unset($searchParams['filters']);
         }
 
+        $searchParams = array_merge($builder->options, $searchParams);
+
         if ($builder->callback) {
             $result = call_user_func(
                 $builder->callback,
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -76,6 +76,13 @@ class Builder
     public $orders = [];
 
     /**
+     * Extra options that should be applied to the search.
+     *
+     * @var int
+     */
+    public $options = [];
+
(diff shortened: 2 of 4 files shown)


driesvints:
@mmachatschek does this look okay to you?

lucasRolff:
I can confirm this works, my applications search experience just got 10x better👌 thank you so much @driesvints!

src/Engines/MeiliSearchEngine.php
```diff
@@ -145,6 +145,8 @@ protected function performSearch(Builder $builder, array $searchParams = [])
             unset($searchParams['filters']);
         }
 
+        $searchParams = array_merge($builder->options, $searchParams);
```
mmachatschek, line 148: I think it would be preferable to us, that users can not override the search params that we send to the engine.

if we do it like this, then our parameters for `limit/offset` (later `page/hitsPerPage` would still get to the underlyling engine.

(this suggestion also relates to the algolia engine)

```suggestion
        $searchParams = array_merge($searchParams, $builder->options);
```

driesvints: Hey @mmachatschek, I thought of that and that's what `array_merge` does: https://3v4l.org/a1JBo

mmachatschek: sorry about that @driesvints 👍 I mixed the contents of `$searchParams` and `$builder->options`

PR #686, opened 2023-01-04
Fix Meilisearch class name case

n8man:
Composer is able to autoload improperly-cased namespaces if configured to do so but will not autoload an improperly-cased class name if the filesystem is case-sensitive.


n8man:
Oh, nevermind, I can see now that meilisearch/meilisearch-php renamed `MeiliSearch\MeiliSearch` to `Meilisearch\Meilisearch` 3 weeks ago at version 0.26.1. Now lame.

n8man:
I have a PR (meilisearch/meilisearch-php#441) that's mean't to fix compatibility with Scout and other projects that use meilisearch-php

PR #687, opened 2023-01-05
[9.x] Make scout compatible with new meilisearch casing

mmachatschek:
This adds workarounds for the new casing of class names that we introduced on `meilisearch/meilisearch-php` which had more implications than we thought. This should also solve issues like https://github.com/laravel/scout/pull/685

Unfortunately PHP handles case sensitivity on some platforms differently

These changes are not necessary for the 10.x branch which (when released) will already handle these case changes

### DIFF
--- a/src/EngineManager.php
+++ b/src/EngineManager.php
@@ -117,8 +117,8 @@ class EngineManager extends Manager
 
         return new MeiliSearchEngine(
             $this->container->make(
-                class_exists(MeiliSearch::class) 
-                    ? MeiliSearch::class 
+                class_exists(MeiliSearch::class)
+                    ? MeiliSearch::class
                     : \Meilisearch\Client::class
             ),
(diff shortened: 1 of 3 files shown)


driesvints:
That's indeed a bit unfortunate.. I personally would have waited with those changes until v1. Thanks for the PR!

driesvints (approved):

siddiquiazam:
@mmachatschek
I was using scout and I came across a rather unusual error, "Undefined variable $meilisearchClientClassName in vendor/laravel/scout/src/ScoutServiceProvider.php:46"

siddiquiazam:
@mmachatschek
I was using scout and I came across a rather unusual error, "Undefined variable $meilisearchClientClassName in vendor/laravel/scout/src/ScoutServiceProvider.php:46"
image

Issue #693, opened 2023-01-08
Undefined variable inside closure (ScoutServiceProvider.php:46)


siddiquiazam:
- Scout Version: 9.1
- Scout Driver: MeiliSearch
- Laravel Version: ^9
- PHP Version: 8.1.6

### Description:
Just installed Laravel scout's latest version with Laravel 9. I was syncing data on MeiliSearch using queue, I observed that my jobs were failing. Upon investigating the error this is what I got: 
![image](https://user-images.githubusercontent.com/72437438/211196629-791215e1-6741-49bb-a5aa-b0c5e8db67cd.png)

The error was referring to this piece of code:

![image](https://user-images.githubusercontent.com/72437438/211196690-b380cd24-10b8-4556-9492-c5f3c515b967.png)

As I am aware that we need to `use` variables inside a closure, so I have raised a PR #691. Please review it, and correct me if I am wrong.

Thanks.

mohamed-foly:
Same issue with me 
Message: Undefined variable $meilisearchClientClassName
Path: vendor/laravel/scout/src/ScoutServiceProvider.php:46

@taylorotwell @driesvints 
It's a blocking issue needs a merge as soon as possible

siddiquiazam:
> Same issue with me Message: Undefined variable $meilisearchClientClassName Path: vendor/laravel/scout/src/ScoutServiceProvider.php:46
> 
> @taylorotwell @driesvints It's a blocking issue needs a merge as soon as possible

I am not able to merge my code on production. Will try to go back to scout version 8.x until this issue is resolved.

Shailu016:
Same issue with me Message: Undefined variable $meilisearchClientClassName Path: vendor/laravel/scout/src/ScoutServiceProvider.php:46

@taylorotwell @driesvints It's a blocking issue needs a merge as soon as possible

I am not able to merge my code on production. Will try to go back to scout version 8.x until this issue is resolved.

siddiquiazam:
@driesvints Can you take a look at PR #691 ?

driesvints:
Hi all, we fixed this in v9.7.2, thanks for reporting.

PR #694, opened 2023-01-09
[9.x] Fix missing variable in closure

driesvints:
This PR fixes a missing variable import for the Meilisearch client in its service provider.

Fixes https://github.com/laravel/scout/issues/693


mmachatschek (commented):
🙌 thanks @driesvints I assumed that loading the service provider was already covered by the tests

Note to myself, always check twice

siddiquiazam:
@mmachatschek I checked the code many times assuming that I was understanding it wrong. 
The new test will help a lot.

PR #695, opened 2023-01-11
[9.x] Enable order by for `collection` & `database` engines

stein-j:
This PR enables `->orderBy()` for the `collection` & `database` engines.

Currently, when using those engines, results are by default ordered by the model key or by any full text columns. Now you may call `->orderBy('column', 'asc');` on the Scout Builder, as you would for other engines, it will be added to the query.

Before:
```php
Users::search('otwell')->orderBy('created_at', 'desc')->get(['id', 'created_at']);
//[{
//       id: 1,
//       date: 2023-01-11 00:00:00.0,
//    },
//    {
//        id: 2,
//        date: 2023-03-11 00:00:00.0,
// }],
```
Now:
```php
Users::search('otwell')->orderBy('created_at', 'desc')->get(['id', 'created_at']);
//[{
//       id: 2,
//       date: 2023-03-11 00:00:00.0,
//    },
//    {
//       id: 1,
//       date: 2023-01-11 00:00:00.0,
// }],
```

### DIFF
--- a/src/Engines/CollectionEngine.php
+++ b/src/Engines/CollectionEngine.php
@@ -101,11 +101,11 @@ class CollectionEngine extends Engine
                             }
                         })
                         ->when($builder->orders, function ($query) use ($builder) {
-                            foreach ($builder->orders as $order){
+                            foreach ($builder->orders as $order) {
                                 $query->orderBy($order['column'], $order['direction']);
                             }
                         }, function ($query) use ($builder) {
-                                $query->orderBy($builder->model->getKeyName(), 'desc');
+                            $query->orderBy($builder->model->getKeyName(), 'desc');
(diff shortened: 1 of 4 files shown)


driesvints:
@mmachatschek subsequently: I notice we don't use `getScoutKeyName` in the database engine. I'm wondering if it makes sense to use it there as well?

mmachatschek:
@driesvints looking at the changes on the 10.x branch it seems you already fixed this issue for the collection engine. The database engine looks like it doesn't need the change.

To verify everything works for the collection engine I suggest adding a test for that case 🤔 

https://github.com/laravel/scout/blob/10.x/src/Engines/CollectionEngine.php#L108

PR #703, opened 2023-02-14
[9.x] Ensure proper garbage collection when running `scout:import`

nicoverbruggen:
Apologies for making the PR to to the `9.x` branch, `10.x` is still very fresh, and as far as I can tell this file is exactly the same on the `10.x` branch.

We have rather large Eloquent models that we are using in combination with Scout, and we started encountering memory issues on our production server when importing large amounts of large models. Now, we are using chunking when importing (in combination with redis) so at first I was unsure as to why memory usage kept going up.

I forked Scout and did some tinkering and used the code below to identify and fix the issue we were seeing.

(I was able to identify the issue by logging the memory usage like I did here: https://github.com/dive-be/scout/commit/69ce819bdd964ddd204d0c685f328ec89f906ce3). I could not do the PR via the organisation, so I made a new one here, with the one line that fixes the issue on our end.

Without this change, in our case (on PHP 8.1, if that matters) the models do not appear to be garbage collected after the closure is finished running. 

As a result, the memory usage of this command keeps growing, even though the actual models in `$event->models` inside the closure are only 250 items (which our chunk size). In our use case, the first 250 items would take up about ~100 MB but grow to beyond 4 GB after plenty of chunks.

Once we use the fix which replaces `$event->models` with an empty collection (or use `unset()` on the actual Collection instance which contain the models) the memory usage significantly drops, and memory usage remains below 200 MB, even after reindexing ~15k records.

I also described the issue here: https://twitter.com/nicoverbruggen/status/1625533566925914119

Now, I'm a bit stumped as to what exactly might be causing this. Perhaps the nature of our own implementation of `toSearchableArray()` is causing this, or some cyclical Eloquent relationships?

Regardless, since the event that is being listened to in this context is only used to log how many models were (re-)indexed when running the command, this might be a fix that can help lower memory usage for others as well, without any negative side effects, as far as I can tell. 

I also noticed a significant speedup in record indexing as a result of this modification as well.

Any initial feedback or thoughts are very welcome!

### DIFF
--- a/src/Console/ImportCommand.php
+++ b/src/Console/ImportCommand.php
@@ -38,6 +38,7 @@ class ImportCommand extends Command
 
         $events->listen(ModelsImported::class, function ($event) use ($class) {
             $key = $event->models->last()->getScoutKey();
+            $event->models = collect();
 
             $this->line('<comment>Imported ['.$class.'] models up to ID:</comment> '.$key);
         });
--- a/src/Console/ImportCommand.php
+++ b/src/Console/ImportCommand.php
@@ -38,9 +38,10 @@ class ImportCommand extends Command
 
         $events->listen(ModelsImported::class, function ($event) use ($class) {
             $key = $event->models->last()->getScoutKey();
-            $event->models = collect();
 
             $this->line('<comment>Imported ['.$class.'] models up to ID:</comment> '.$key);
+
+            unset($event->models);
         });
 
         $model::makeAllSearchable($this->option('chunk'));


taylorotwell:
@nicoverbruggen can you tell me what other packages you are using? Are you using Telescope?

nicoverbruggen:
Yes, we do use Telescope. We are also using a lot of packages... Let me think, in particular, relevant to Scout I'd say we are using [this Elastic driver](https://github.com/babenkoivan/elastic-scout-driver). Apologies for the wall of text, I figure all of this might help you understand what's going on in our app.

We are using Horizon with Redis, and it appears as if the `Laravel\Scout\Events\ModelsImported` event is not being recorded, no.

When we run the command, we can see in Horizon that for each chunk Scout attempts to make searchable, a `Laravel\Scout\Jobs\MakeSearchable` job is being dispatched (and handled if the queue worker is active).

The only event that is being logged in Telescope is `Laravel\Horizon\Events\JobDeleted`, which is logged when the queue has successfully executed one of those jobs and 250 models were successfully updated (made searchable) in Elastic.

Because we are using the queue to actually handle the updating of the index, we don't even need to have the queue be handling jobs for this issue to occur: running the command directly seems to 1) queue the jobs and 2) call the closure to report in real-time that entries up to ID X have been imported. 

When the PHP process handling the command runs out of memory, the command simply terminates, I don't always get a `Whoops\Exception\ErrorException` that reports that we ran out of memory either. (Initially, we didn't even realize anything was wrong until one of our clients noticed a record wasn't being returned through search. I then ran this command locally and bumped into the issue. Only when I changed the chunk size of Scout to 1 did we eventually see an exception that informed us we ran out of memory.)

The models that are being queried and end up in the command's closure seem to be retained in any case. We are loading a ton of potential relationships on each of those records, as the data we include in the index can be quite complex.

taylorotwell:
Can you try disabling Telescope entirely and seeing if you still have the memory leak?

nicoverbruggen:
Aha, disabling Telescope (by setting `TELESCOPE_ENABLED` to false in the `.env` file) completely gets rid of the memory leak, yes!

nicoverbruggen:
Yes, Telescope is definitely causing the memory leak. If I don't record the job in `JobWatcher`'s `recordJob` method by adding this:

```diff
public function recordJob($connection, $queue, array $payload)
{
    if (! Telescope::isRecording()) {
        return;
    }

+    // Quick edit to check if this fixes the issue
+    if (str_starts_with(get_class($payload['data']['command']), 'Laravel\Scout')) {
+        return;
+    }
}
```

to exit early when a Scout event is being handled, the memory leak is gone.

taylorotwell:
Yep, let's fix that in Telescope then. Will take care of that now.

taylorotwell:
https://github.com/laravel/telescope/commit/057104e51834f672eb19525229b5637b4ae47635

nicoverbruggen:
Hmm, that doesn't seem to fix the problem: with that fix I still have the same problem.

Is there currently any way to ignore particular jobs from being logged in Telescope? 

(It's actually the `Laravel\Scout\Jobs\MakeSearchable` job being fired that's causing the leak in our case. I'm not sure if that's a job you want to prevent from being logged in all cases...)

taylorotwell:
@nicoverbruggen ah - can you just PR the change that helps you and then I will review it from there?

nicoverbruggen:
@taylorotwell I will do that, thanks!

Issue #710, opened 2023-03-04
Call to undefined method getUnqualifiedScoutKeyName()


G33RY:
- Scout Version: 10.0.0
- Scout Driver: Meilisearch
- Laravel Version: 10.0.2
- PHP Version: 8.1.12
- Database Driver & Version: mysql
- Meilisearch CLI Version (If using Meilisearch): 1.0.2

### Description:

The search works great when I am using simple characters, but if I use some special characters like "~", "-", "." I get this error:
```
Call to undefined method App\\Models\\Image::getUnqualifiedScoutKeyName() at vendor/laravel/framework/src/Illuminate/Support/Traits/ForwardsCalls.php:71)
```

### Steps To Reproduce:
I have a model named `Image` that has the `Searchable` trait, and `toSearchableArray` method: 
```
public function toSearchableArray(): array
{
        return [
            'id'   => $this->getKey(),
            'slug' => $this->slug,
            'type' => $this->type,
            'keywords' => $this->keywords,
            'is_visible' => $this->is_visible,
            'labels' => $this->getLabelsAttribute(),
            'created_at' => $this->created_at,
            'updated_at' => $this->updated_at,
        ];
}
```

`scout.php`:
```
[
   'soft_delete' => true,
   'queue'  => true,
   'after_commit' => true,
   'meilisearch' => [
        'host' => 'http://localhost:7700',
        'key' => '',
        'index-settings' => [
             'images' => [
                 'filterableAttributes'=> ['labels', 'keywords', 'slug', 'type'],
                 'sortableAttributes' => ['created_at', "updated_at"],
             ],
        ],
    ],
]
```

`ImageController`
```
public function index() 
{
    $search = request('search'); //If this is "." or some special character it throws that error 
    return Image::search($search)->get()->all();
}

```

driesvints:
Thanks, this is resolved now

PR #711, opened 2023-03-05
Fix missing function issue

JerryBels:
`getUnqualifiedScoutKeyName` doesn't exist anymore. Since `getScoutKeyName` sends back to `\Illuminate\Database\Eloquent\Model::getKeyName` which is unqualified, the fix is straightforward.

### DIFF
--- a/src/Engines/MeilisearchEngine.php
+++ b/src/Engines/MeilisearchEngine.php
@@ -248,7 +248,7 @@ class MeilisearchEngine extends Engine
      */
     public function keys(Builder $builder)
     {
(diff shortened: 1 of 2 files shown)


JerryBels:
Resolves #710

driesvints:
Thanks @JerryBels. 

@mmachatschek this is correct, right

mmachatschek:
This looks ok 👍 probably was some rebase issue on my side

driesvints:
No worries. I could have maybe missed this as well during merging 9.x into 10.x, thanks!

PR #713, opened 2023-03-06
Use `newQuery()` instead of `query()` in `DatabaseEngine`

### LINKED ISSUE #712: Issue with dynamic schema
- Scout Version: 9.8.1
- Scout Driver: Database
- Laravel Version: 9.52.0
- PHP Version: 8.1.14
- Database Driver & Version: 11.18
- ~SDK Version (If using a third-party service): #.#.#~
- ~Meilisearch CLI Version (If using Meilisearch): #.#.#~

### Description:

It's not really a bug as such but the Laravel Scout search doesn't work in my specific use case.

I would like to enable Laravel Scout search on a model stored in a specific schema in PostgreSQL.
That schema is set dynamically via a `SchemaTrait` (see below). The issue happens in `initializeSearchQuery()` function defined in `DatabaseEngine` class: it calls `$builder->model->query()` which will create a new query from the static model ; using `$builder->model->newQuery()` fixes this "issue".

That's what my `SchemaTrait` looks like (that trait is added to my model): 
```php
<?php

trait SchemaTrait
{
    public function setSchema(string $schema): self
    {
        $this->table = sprintf('%s.%s', $schema, $this->table);

        return $this;
    }

    public function getSchema(): string
    {
        if (strpos($this->table, '.') !== false) {
            [$schema, $table] = explode('.', $this->table);
        } else {
            $schema = 'public';
        }

        return $schema;
    }
}
```

Would you consider replacing `$builder->model->query()` by `$builder->model->newQuery()` in `initializeSearchQuery()` function ?

Thanks.

### Steps To Reproduce:

1. Create a model `SchemaTest` (`$table` property should be set to the name of the table **without** the schema, the schema will be set dynamically)
2. Add `SchemaTrait` to the model by adding `use SchemaTrait;` in your `SchemaTest` class
3. Create a new instance of that model: `$instance = new SchemaTest();`
4. Set the schema dynamically: `$instance->setSchema('test');`
5. Try to call a Laravel Scout search: `$instance->search('test');`

It will complain it can't find the table because the schema is not set.

jbelien:
The issue with `$builder->model->query()` is the fact that it creates a new query from a "new" **static** model; `$builder->model->newQuery()` will create a query from the current object (see [`Illuminate\Database\Eloquent\Model` abstract class](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Eloquent/Model.php#L1469-L1487)).

If you set the schema (or table name) of your model dynamically, `$builder->model->query()` will not work.
See #712 for more details.

Close #712

### DIFF
--- a/src/Engines/DatabaseEngine.php
+++ b/src/Engines/DatabaseEngine.php
@@ -162,10 +162,10 @@ class DatabaseEngine extends Engine implements PaginatesEloquentModels
     protected function initializeSearchQuery(Builder $builder, array $columns, array $prefixColumns = [], array $fullTextColumns = [])
     {
         if (blank($builder->query)) {
-            return $builder->model->query();
+            return $builder->model->newQuery();
         }
 
-        return $builder->model->query()->where(function ($query) use ($builder, $columns, $prefixColumns, $fullTextColumns) {
+        return $builder->model->newQuery()->where(function ($query) use ($builder, $columns, $prefixColumns, $fullTextColumns) {
             $connectionType = $builder->model->getConnection()->getDriverName();
 
             $canSearchPrimaryKey = ctype_digit($builder->query) &&


driesvints (approved):
I think this looks okay to me.

PR #715, opened 2023-03-22
Added support for complex where clauses in scout builder, as original…

alibaddd:
As can be seen in the scout documentation[1], when using the database driver, scout only supports simple where clauses that work only for basic numeric equality checks.
I edited the addAdditionalConstraints method found in the DatabaseEngine and the where method found in scout/src/Builder.php in order to add the following functionalities :

- An array of arrays can be passed as a parameter : where([['id', '>', 5],['id', '<', 10]]).
- Any operator supported by the Eloquent Builder is now supported as well : '=', '<', '>', '<=', '>=', '!=', etc.
- Clauses containing the operator like so : where('id', '>', 5). 
- All changes are fully backward compatible, If no operator is passed as a parameter, it will be assumed that the desired operator is '=' (where('id', 5) <=> where('id', '=', 5)).

[1]: https://laravel.com/docs/10.x/scout#where-clauses

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -161,7 +161,7 @@ class Builder
         // If no operator is passed as a parameter, it will be assumed that the desired operator is '='
         // where('id', 5) <=> where('id', '=', 5)
         $this->wheres[$column] = $operator;
-        
+
         return $this;
     }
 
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #716, opened 2023-03-24
[10.x] Add `getColumn` method to `Builder`

alirezasalehizadeh:
In this PR, added `getColumn` method to Builder class. (Database driver)

By default in each search request we recived all columns in our query like this:
```php
Post::search('php')->get();
```

img1

This is not good and for better performance it is better to use the required columns for each search.
The `getColumn` method help us to get all needed columns in query like this:
```php
Post::search('php')->getColumn('title');
```
img2

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -3,6 +3,7 @@
 namespace Laravel\Scout;
 
 use Illuminate\Container\Container;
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #726, opened 2023-04-06
[11.x] Adds support for custom page names using the database engine

lukeraymonddowning:
Hey there!

## Reason

This PR allows for specifying a page name when performing scout pagination using the database engine so that you can have two or more searchable indexes on a single page. 

## Rationale for approach

This only really applies to the `DatabaseEngine`, so adding the property to the `paginate` method might seem overkill, but as far as I can tell, there isn't really another way to add the page name without doing something a funky like having a `withPageName($pageName)` method on the `PaginatesEloquentModels` contract. I also think that this ends up ultimately being the cleanest solution.

If you'd rather me take the other approach btw, just say and I'll change this PR.

## Why not v10?

Obviously, either way, it's a breaking change because whilst it doesn't break any user facing code (the `Builder` remains identical), it does change the signature of `Engine` and `PaginatesEloquentModels`, which will break 3rd party drivers for Scout. As such, I've targeted v11.

Thanks for all the hard work, time and effort you put into everything.

Kind Regards,
Luke

### DIFF
--- a/src/Contracts/PaginatesEloquentModels.php
+++ b/src/Contracts/PaginatesEloquentModels.php
@@ -11,18 +11,20 @@ interface PaginatesEloquentModels
      *
      * @param  \Laravel\Scout\Builder  $builder
      * @param  int  $perPage
+     * @param  string  $pageName
      * @param  int  $page
      * @return \Illuminate\Contracts\Pagination\LengthAwarePaginator
      */
-    public function paginate(Builder $builder, $perPage, $page);
+    public function paginate(Builder $builder, $perPage, $pageName, $page);
 
     /**
      * Paginate the given search on the engine using simple pagination.
      *
      * @param  \Laravel\Scout\Builder  $builder
      * @param  int  $perPage
+     * @param  string  $pageName
      * @param  int  $page
      * @return \Illuminate\Contracts\Pagination\Paginator
      */
-    public function simplePaginate(Builder $builder, $perPage, $page);
+    public function simplePaginate(Builder $builder, $perPage, $pageName, $page);
 }
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -316,7 +316,7 @@ class Builder
         $engine = $this->engine();
 
         if ($engine instanceof PaginatesEloquentModels) {
-            return $engine->simplePaginate($this, $perPage, $page)->appends('query', $this->query);
+            return $engine->simplePaginate($this, $perPage, $pageName, $page)->appends('query', $this->query);
         }
 
         $page = $page ?: Paginator::resolveCurrentPage($pageName);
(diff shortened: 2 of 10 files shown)


src/Contracts/PaginatesEloquentModels.php
```diff
@@ -11,18 +11,20 @@ interface PaginatesEloquentModels
      *
      * @param  \Laravel\Scout\Builder  $builder
      * @param  int  $perPage
+     * @param  string  $pageName
      * @param  int  $page
      * @return \Illuminate\Contracts\Pagination\LengthAwarePaginator
      */
-    public function paginate(Builder $builder, $perPage, $page);
+    public function paginate(Builder $builder, $perPage, $pageName, $page);
```
driesvints, line 18: Wondering if we can instead do the below to make it less breaking?

```suggestion
    public function paginate(Builder $builder, $perPage, $page, $pageName = 'page');
```

lukeraymonddowning: Yeah, I did consider this option. However, being that it's an interface, all implementations will still break won't they? And by adding to the end, we break uniform with how all the other `paginate` methods are defined 🤔

lukeraymonddowning: However, happy to change to this if we think that's the best option; just providing my reasoning. LMK 👍

driesvints: It's not so much about the implementation but about usage. End users who use scout and not implement their own engine will have a smooth upgrade path. Otherwise you still need to move the "page" argument to the fourth one.

lukeraymonddowning: Don't the end users use the `Builder` rather than the engines directly? So their usage doesn't change.

driesvints: Yeah that's true. Let's see what @taylorotwell says.

taylorotwell:
Retarget this to current branch - I can make it non-breaking.

taylorotwell:
@lukeraymonddowning let me know if that makes sense.

lukeraymonddowning:
@taylorotwell that does make sense. I guess the only thing is that the abstract `Engine` class (which all engines extend), has the `paginate` method on it, which doesn't have the `$pageName` parameter. Any thoughts on overcoming that obstacle? Or are you thinking of having an additional (database only) method like `paginateDatabase` on the `PaginatesEloquentModelsUsingDatabase` contract?

taylorotwell:
Yeah - true, we will probably need a new method there as you said.

lukeraymonddowning:
@taylorotwell I've opened up a new PR #728, so I'll close this one.

PR #728, opened 2023-04-07
[10.x] Adds support for custom page names using the database engine

lukeraymonddowning:
Hey there!

This is a follow up PR to https://github.com/laravel/scout/pull/726 which uses the suggested approach to introduce this ability in a non-breaking fashion.

It does this by introducing a new contract: `PaginatesEloquentModelsUsingDatabase`, which includes support for specifying a `$pageName` for both the `paginate` and `simplePaginate` methods. Technically, this replaces `PaginatesEloquentModels`, which is no longer used in the codebase at all. However, I have left the logic in place in case a 3rd party driver uses that contract. Perhaps we could remove it in a future version of Scout?

I have left the old `simplePaginate` method on the `DatabaseEngine` so that should anybody be doing something strange like injecting it directly, the signature remains identical. The old method simply defers to the new method, passing manual `$pageName` of 'page', which replicates previous behaviour.

### DIFF
--- a/src/Engines/DatabaseEngine.php
+++ b/src/Engines/DatabaseEngine.php
@@ -71,7 +71,7 @@ class DatabaseEngine extends Engine implements PaginatesEloquentModelsUsingDatab
      */
     public function paginate(Builder $builder, $perPage, $page)
     {
-        return $this->paginateDatabase($builder, $perPage, 'page', $page);
+        return $this->paginateUsingDatabase($builder, $perPage, 'page', $page);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Engines/DatabaseEngine.php
```diff
@@ -8,10 +8,10 @@
 use Laravel\Scout\Attributes\SearchUsingFullText;
 use Laravel\Scout\Attributes\SearchUsingPrefix;
 use Laravel\Scout\Builder;
-use Laravel\Scout\Contracts\PaginatesEloquentModels;
+use Laravel\Scout\Contracts\PaginatesEloquentModelsUsingDatabase;
 use ReflectionMethod;
 
-class DatabaseEngine extends Engine implements PaginatesEloquentModels
+class DatabaseEngine extends Engine implements PaginatesEloquentModelsUsingDatabase
```
driesvints, line 14: Don't we need to use the both interfaces here? Technically this is still a breaking change for people relying on the previous contract to be implemented.

driesvints (approved):

PR #732, opened 2023-04-24
[10.x] Add 'makeSearchableUsing' method (to allow eager loading when making specific models searchable)

gdebrauwer:
In the projects at work, we use Laravel's lazy loading prevention. But when installing and using Scout, I noticed that Scout does not use `makeAllSearchableUsing` when making specific models searchable (which I thought at first). This causes lazy loading errors when the `toSearchableArray` method expects certain relations to be loaded (as is the case when importing all models because it then uses the `makeAllSearchableUsing` method).
```php
public function toSearchableArray() : array
{
    return [
        // This causes lazy loading on every model that is made searchable
        'some_relation_attribute' => $this->someRelation->attribute
    ];
}

$someModels->searchable();
$parentModel->someRelation()->searchable();
```

So to fix this, I propose introducing a new method: `makeSearchableUsing`. This method allows you to modify models before the `toSearchableArray` method is called on each of those models 

```php
protected function makeSearchableUsing($models)
{
    // load relations or add extra data that will be used in the "toSearchableArray" method
    return $models->load(...);
}
```

### DIFF
--- a/src/Jobs/MakeSearchable.php
+++ b/src/Jobs/MakeSearchable.php
@@ -39,6 +39,6 @@ class MakeSearchable implements ShouldQueue
             return;
         }
 
-        $this->models->first()->searchableUsing()->update($this->models);
+        $this->models->first()->makeSearchableUsing($this->models)->first()->searchableUsing()->update($this->models);
     }
 }
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -152,7 +152,7 @@ trait Searchable
     }
 
     /**
-     * Modify the models used to make those models searchable.
+     * Modify the collection of models being made searchable.
      *
      * @param  \Illuminate\Support\Collection  models
      * @return \Illuminate\Support\Collection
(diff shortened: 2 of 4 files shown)


taylorotwell:
Can't you just call `load` on the collection before calling `searchable`?

gdebrauwer:
No that does not work when data syncing is queued. And even if it would work, it would be very annoying in my opinion. You might have to make models searchable in different places, which means you have to repeat the eager loading code in all those places.

taylorotwell:
Why `BaseCollection` instead of an Eloquent collection? Which one is it? Base collection doesn't have a `load` method.

gdebrauwer:
ah indeed it should be an eloquent collection

mstaack:
this broke my scout setup, had to revert to version before.

see: https://github.com/laravel/scout/issues/736

nemoitis:
@mstaack Same here. All on a sudden my tests are going red.

PR #737, opened 2023-05-10
[10.x]  Meilisearch v1 returns estimatedTotalHits or totalHits, but not both

RoccoHoward:
## Issue
When using pagination we get a `Undefined array key "totalHits"`

The Meilisearch engine provides `estimatedTotalHits` or `totalHits` but not both.

The logic is currently provided upstream in `meilisearch-php`, but is currently providing the old key name `nbHits`, instead of what is now expected as `totalHits`.

https://github.com/meilisearch/meilisearch-php/blob/78cec653dcd1a796596e947f9e02d889ec0a606b/src/Endpoints/Indexes.php#L206

I created a pull request to add this, but it would better to maintain this at the Scout side to provide BC.
https://github.com/meilisearch/meilisearch-php/pull/502

### DIFF
--- a/src/Engines/MeilisearchEngine.php
+++ b/src/Engines/MeilisearchEngine.php
@@ -316,6 +316,10 @@ class MeilisearchEngine extends Engine
      */
     public function getTotalCount($results)
     {
+        if (isset($result['estimatedTotalHits'])) {
+            return $result['estimatedTotalHits'];
+        }
+
         return $results['totalHits'];
     }
 
--- a/src/Engines/MeilisearchEngine.php
+++ b/src/Engines/MeilisearchEngine.php
@@ -316,8 +316,8 @@ class MeilisearchEngine extends Engine
      */
     public function getTotalCount($results)
     {
-        if (isset($result['estimatedTotalHits'])) {
-            return $result['estimatedTotalHits'];
+        if (isset($results['estimatedTotalHits'])) {
+            return $results['estimatedTotalHits'];
         }
 
         return $results['totalHits'];


mmachatschek:
@RoccoHoward how does your implementation look like? do you extend the pagination api of some sort? this can only happen if you override the parameters that we send to meilisearch. We fix the search params to `page` and `hitsPerPage` to which meilisearch will always respond with the `totalHits` key in the reponse.

RoccoHoward:
Thanks @mmachatschek  - I revised the code, and see that I was passing an empty options array via a closure - which isn't actually necessary for this specific search. I was adding a facet filter in another index search, but had copy-pasted and not cut it back to what was actually needed here.

```
$query = Display::search($this->search, function ($meilisearch, $query) {
            try {
                return $meilisearch->search($query, []);
            } catch (\Exception $exception) {
                abort(500);
            }
        })->within('display_index');
```
(passing $options from closure to search, also fixed it)

And now, just doing this without needing to make any adjustments to Scout code.
 
 ```
 $query = Display::search($this->search)->within('display_index');
 ```

taylorotwell:
@mmachatschek so is this actually needed?

mmachatschek:
@taylorotwell the upgrade guide didn't explicitly mention this. but it is expected that the custom code of @RoccoHoward breaks the implementation, as we really on an exhaustive search rather than an estimated result

zappzerapp:
I did a search with callback and overwrote the `$options` array instead of merging it.
This also gave me the ErrorException
```
Undefined array key "totalHits"
```


#### Wrong: 
```
return Model::search(callback: fn (Indexes $meilisearch, string $query, array $options) => $meilisearch->search($query, [
    'filter' => [...],
    'sort' => [...]
]));
```


#### Right: 
```
return Model::search(callback: fn (Indexes $meilisearch, string $query, array $options) => $meilisearch->search($query, [
    'filter' => [...],
    'sort' => [...],
    ...$options
]));
```

Issue #723, opened 2023-04-04
Import query duplicates key column in orderBy causing SQL exception


gdesoto:
- Scout Version: 10.0.2
- Scout Driver: tntsearch (12.4.0)
- Laravel Version: 10.5.1
- PHP Version: 8.1.13
- Database Driver & Version: SQL Server 2017 (Microsoft ODBC Driver 17 for SQL Server)
- SDK Version (If using a third-party service): #.#.#
- Meilisearch CLI Version (If using Meilisearch): #.#.#

### Description:
When running the scout:import command for a model, the following error is thrown while the command gets all searchable models from a MS SQL Server database.

> Illuminate\Database\QueryException
> 
> SQLSTATE[42000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]A column has been specified more than once in the order by list. Columns in the order by list must be unique. (Connection: default, SQL: select top 500 * from [users] order by [users].[id] asc, [id] asc)

The `makeAllSearchable` method in `src/Searchable.php` called by the import command adds an `orderBy` on the primary scout key prior to the calling `searchable()` macro. The searchable macro defined in `src/SearchableScope.php` then calls Eloquent's `chunkById` method which adds its own `orderBy` on the model's primary key. When both keys are the same, an error will be thrown.

This issue seems to have started with this change to use `qualifyColumn` method in the `orderBy` [#509](https://github.com/laravel/scout/pull/509#issuecomment-903721511). As a test, I reverted this line and no error was thrown.

Laravel's `chunkByID` method removed existing "orders" for a query prior to adding its own which explains why it didn't throw an error prior to this change. Since `[users].[id] !== [id]` it isn't removed from the query anymore, and now throws the error.

Assuming moving to qualifying column names is the future, then a possible solution may be to find a way to pass the results of `qualifyColumn()` to `chunkById` as the third parameter. Something like...
```diff
        $builder->macro('searchable', function (EloquentBuilder $builder, $chunk = null) {
            $builder->chunkById($chunk ?: config('scout.chunk.searchable', 500), function ($models) {
                $models->filter->shouldBeSearchable()->searchable();

                event(new ModelsImported($models));
-            });
+            }, $builder->qualifyColumn($builder->getModel()->getScoutKeyName()));
        });
```

I haven't tested in other database environments. I didn't come across any other reports of this issue when searching, so I'm thinking other database engines must ignore the duplicated column names in the orderBy query.

### Steps To Reproduce:

1. SQL Server environment
[...]

PR #740, opened 2023-05-18
[10.x] Fixes usage with `sqlsrv` databases

nunomaduro:
Scout is currently experiencing compatibility issues with  Microsoft SQL Server databases, as an error is thrown when attempting to import models on Scout. The error message is as follows:

```
  SQLSTATE[HY000]: General error: 20018 A column has been specified more than once in the order by list. Columns in the order by list must be unique. [20018] (severity 15) [select top 500 * from [users] order by [users].[id] asc, [id] asc] (Connection: sqlsrv, SQL: select top 500 * from [users] order by [users].[id] asc, [id] asc)
  ```
  
The regression was introduced in the changes made during the pull request at https://github.com/laravel/scout/pull/509. It happens because two "order by" statements are used with the same column name. While this doesn't pose any problems with other database drivers, it causes issues with Microsoft SQL Server databases, as they do not allow multiple "order by" clauses with the same column name.

Fixes https://github.com/laravel/scout/issues/723.

### DIFF
--- a/src/SearchableScope.php
+++ b/src/SearchableScope.php
@@ -32,19 +32,23 @@ class SearchableScope implements Scope
     public function extend(EloquentBuilder $builder)
     {
         $builder->macro('searchable', function (EloquentBuilder $builder, $chunk = null) {
+            $model = $builder->getModel();
+
             $builder->chunkById($chunk ?: config('scout.chunk.searchable', 500), function ($models) {
                 $models->filter->shouldBeSearchable()->searchable();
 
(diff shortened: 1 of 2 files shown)


nunomaduro:
@driesvints @mmachatschek Do you mind of reviewing this pull request, and let me know what do you think?

mmachatschek (commented):
@nunomaduro @driesvints LGTM ✅ 🚀

driesvints (approved):

Issue #741, opened 2023-05-24
Problem with Laravel Scout treating UUIDs as integers


ianfortier:
### Scout Version

10.2.1

### Scout Driver

Meilisearch

### Laravel Version

10.0.4

### PHP Version

8.1

### SDK Version

1.1

### Meilisearch CLI Version

1.1

### Description

Hi Nuno, I apologize for being the only open issue.

I've encountered an issue when using Laravel Scout with MeiliSearch. Specifically, the problem arises when the primary key for my Eloquent model is an integer, but I use a UUID for searching in MeiliSearch.

The issue appears to lie in this snippet in Laravel Scout's get method:

```
$whereIn = in_array($this->getKeyType(), ['int', 'integer']) ? 'whereIntegerInRaw' : 'whereIn';
```
In this case, the `getKeyType()` method returns 'int', causing Laravel Scout to treat the UUID, used for searching, as an integer.

As a result, the get method doesn't retrieve any results because it attempts to parse UUIDs as integers, which results in a failed SQL query. 

Conversely, using the raw method does return correct results, as it doesn't involve the said parsing.

To resolve this issue, I suggest adding a new method like getScoutKeyType which would provide the data type used for the searchable key in Laravel Scout. This would allow developers to specify a different key type for search indexing, distinct from the getKeyType method. 

For instance:

```
public function getScoutKeyType()
{
    return 'string'; // For UUIDs
}
```

Let me know if you need more details!

Thanks!

### Steps To Reproduce

You can reproduce this issue by using any model that implements the Searchable trait with a custom string key.

driesvints:
Heya. This seems more like a feature request so it's best that you send in a PR for us to look at. Right now this isn't supported, sorry.

PR #752, opened 2023-07-20
Add Configurable Scout Key Type

geisi:
Hello,

this PR improves the Scout integration when using custom keys for Searchable Models.

## Problem

This problem was already addressed in the following issue:
https://github.com/laravel/scout/issues/741

In one of our applications, we use incrementing IDs as the primary key for our models. Additionally, we are using the `HasUuids` trait to add a uuid column on top of the incrementing id. We did that because we didn't want to expose our incrementing IDs to the front end (and the search engine) due to security concerns. 

We have done that by modifying the `getScouteKey' method.
```PHP
    public function getScoutKey(): string
    {
        return $this->uuid;
    }
```

So far, so good indexing of these models works. But when we want to use the `Model::search()->get()` method within the query builder, we get a database exception because Scout thinks our primary key is an integer. 

This happens within the `queryScoutModelsByIds` method. The scout package determines the used key type (integer or string) via the Eloquent `getKeyType()` method. This behavior is not correct in this case. When you can define a custom Scout key column, you should also be able to define the column type of this custom key.

### Workaround

Before this PR, The workaround for this problem is to override the `queryScoutModelsByIds()` method within the models so the builder is doing a `whereIn()` instead of a `whereIntegerRaw()` condition.

## Solution

I have added a new `getScoutKeyType(): string` method, which, by default, returns the model key type. This new method allows you to customize the key type Scout uses for querying models.

This should not be a breaking change since it does not change the default behavior of Scout, but it adds a little more flexibility in how custom key types can be used in Scout. There is a minor risk of naming collisions, but since the new method is prefixed with the *scout* word, it should be no problem.

### DIFF
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -64,8 +64,8 @@ trait Searchable
         }
 
         dispatch((new Scout::$makeSearchableJob($models))
-                ->onQueue($models->first()->syncWithSearchUsingQueue())
-                ->onConnection($models->first()->syncWithSearchUsing()));
+            ->onQueue($models->first()->syncWithSearchUsingQueue())
+            ->onConnection($models->first()->syncWithSearchUsing()));
     }
 
     /**
@@ -253,7 +253,7 @@ trait Searchable
             call_user_func($builder->queryCallback, $query);
         }
 
-        $whereIn = in_array($this->getKeyType(), ['int', 'integer']) ?
+        $whereIn = in_array($this->getScoutKeyType(), ['int', 'integer']) ?
             'whereIntegerInRaw' :
             'whereIn';
 
@@ -391,6 +391,16 @@ trait Searchable
     public function getScoutKey()
     {
         return $this->getKey();
+    }
+
+    /**
+     * Get the auto-incrementing key type for querying models.
+     *
+     * @return string
+     */
+    public function getScoutKeyType()
+    {
+        return $this->getKeyType();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


driesvints:
Hey @mmachatschek, does this change look okay to you?

geisi:
I think if this PR gets merged, we should also create a PR for the Docs:

By default, Scout will use the primary key of the model as the model's unique ID / key that is stored in the search index. If you need to customize this behavior, you may override the `getScoutKey`,  `getScoutKeyName` and  `getScoutKeyType` methods on the model:

    <?php

    namespace App\Models;

    use Laravel\Scout\Searchable;

    class User extends Model
    {
        use Searchable;

        /**
         * Get the value used to index the model.
         */
        public function getScoutKey(): mixed
        {
            return $this->email;
        }

        /**
         * Get the key name used to index the model.
         */
        public function getScoutKeyName(): mixed
        {
            return 'email';
        }

        /**
         * Get the auto-incrementing key type for querying models.
         *
         * @return string
         */
        public function getScoutKeyType()
        {
            return 'string';
        }

This current example in the docs won't work with Model::search() when you use incrementing IDs as the model's primary key.

mmachatschek:
@driesvints the changes look good to me, even though I would prefer if we could add a test for this as well.

geisi:
@mmachatschek Yes, you are right.

In the current test suite, I did not find any tests for the currently existing behavior, so I have not added some.

But I can create tests for the whole scoutKey logic.

mmachatschek:
@geisi yes that would be great

geisi:
@mmachatschek Tests are done (at least) for the new logic

driesvints (approved):

PR #754, opened 2023-08-16
Update meilisearch documentation link

larsbo:
This documentation link inside the config file is outdated and returns a 404. Not sure, but this link seems to be the best alternative

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -125,7 +125,7 @@ return [
     | source search engine with minimal configuration. Below, you can state
     | the host and key information for your own Meilisearch installation.
     |
-    | See: https://docs.meilisearch.com/guides/advanced_guides/configuration.html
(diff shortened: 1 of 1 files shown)


config/scout.php
```diff
@@ -125,7 +125,7 @@
     | source search engine with minimal configuration. Below, you can state
     | the host and key information for your own Meilisearch installation.
     |
-    | See: https://docs.meilisearch.com/guides/advanced_guides/configuration.html
+    | See: https://www.meilisearch.com/docs/learn/configuration/instance_options#all-instance-options
```
mmachatschek, line 128: Even though thats the "new" link, it would be useful to have an additional link to https://www.meilisearch.com/docs/learn/configuration/settings below so one can immediately find the index settings as well

PR #757, opened 2023-08-20
Support backed enums

Carlwirkus:
Meilisearch supports backed enums when inserting data, by extension I think it would make sense to support also filtering by them in the MeilisearchEngine.


driesvints:
@mmachatschek does this look good to you?

mmachatschek:
@driesvints firstly, using `BackedEnum` will require to bump the minimum required version of PHP + this feature would also be useful for other engines. this is not tied to Meilisearch directly

driesvints:
Thanks @mmachatschek. 

@Carlwirkus can you send this to master instead? Thanks!

PR #758, opened 2023-08-22
Meilisearch: support filtering with backed enums

Carlwirkus:
Meilisearch supports backed enums when inserting data, by extension I think it would make sense to also support filtering by them.


driesvints:
Seems something went wrong with branching here.

PR #772, opened 2023-10-17
[10.x] Add Typesense engine

jasonbosco:
@taylorotwell @driesvints Following up from our conversation, this PR adds native support for [Typesense](https://github.com/typesense/typesense) (an open source alternative to Algolia) to Scout. 

It follows a code structure similar to the Algolia engine to maintain consistency.

CC: @karakhanyans @arayiksmbatyan

PR #773, opened 2023-10-17
[10.x] Add Typesense engine

jasonbosco:
@taylorotwell @driesvints Following up from our conversation, this PR adds native support for [Typesense](https://github.com/typesense/typesense) (an open source alternative to Algolia) to Scout. 

It follows a code structure similar to the Algolia engine to maintain consistency.

CC: @karakhanyans @arayiksmbatyan

P.S. This PR supersedes #772. 

---

> Edit: Adding documentation for the Typesense Scout engine, similar to the docs [here](https://laravel.com/docs/10.x/scout#meilisearch)

## Typesense

[Typesense](https://typesense.org/) is an open source, typo tolerant search engine that is optimized for instant sub-50ms searches,
while providing an intuitive developer experience.

When using the Typesense driver you will need to install the Typesense PHP SDK via the Composer package manager:

```bash
composer require typesense/typesense-php
```

Then, set the `SCOUT_DRIVER` environment variable as well as your Typesense host and API key credentials within your application's .env file:

```env
SCOUT_DRIVER=typesense
TYPESENSE_API_KEY=xyz
TYPESENSE_HOST=localhost
```

You can also modify `port`, `path` and `protocol` accordingly:

```env
TYPESENSE_PORT=8108
TYPESENSE_PATH=ts
TYPESENSE_PROTOCOL=https
```

For more information regarding Typesense, please consult the [Typesense documentation](https://typesense.org/docs/guide/#quick-start).

### `toSearchableArray`

To make your search compatible with Typsense, cast your model `id` to string and `created_at` to int32 timestamp as shown below.

```php
<?php

namespace App;

use Laravel\Scout\Searchable;

class Todo extends Model
{
    use Searchable;
    
     /**
     * Get the indexable data array for the model.
     *
     * @return array
     */
    {
        return array_merge(
            $this->toArray(), 
            [
                // Cast id to string and turn created_at into an int32 timestamp
                // in order to maintain compatibility with the Typesense index definition below
                'id' => (string) $this->id,
                'created_at' => $this->created_at->timestamp,
            ]
        );
    }
}
```

### Collection Schemas and Search Parameters

To setup Typesense collection schemas, in `config/scout.php` find the `typesense` driver configs and modify `model-settings` array as shown in example below.

To modify the model `query_by` attributes you have to setup `seach-parameters` in `model-settings`

Note: `query_by` is a required parameter.

```php
use App\Models\User;

 /*
    |--------------------------------------------------------------------------
    | Typesense Configuration
    |--------------------------------------------------------------------------
    |
    | Here you may configure your Typesense settings. Typesense is an open
    | source search engine with minimal configuration. Below, you can state
    | the host and key information for your own Typesense installation.
    |
    | See: https://typesense.org/docs/0.25.1/api/authentication.html
    |
    */

    'typesense' => [
        'client-settings' => [
            'api_key' => env('TYPESENSE_API_KEY', 'xyz'),
            'nodes' => [
                [
                    'host' => env('TYPESENSE_HOST', 'localhost'),
                    'port' => env('TYPESENSE_PORT', '8108'),
                    'path' => env('TYPESENSE_PATH', ''),
                    'protocol' => env('TYPESENSE_PROTOCOL', 'http'),
                ],
            ],
            'nearest_node' => [
                'host' => env('TYPESENSE_HOST', 'localhost'),
                'port' => env('TYPESENSE_PORT', '8108'),
                'path' => env('TYPESENSE_PATH', ''),
                'protocol' => env('TYPESENSE_PROTOCOL', 'http'),
            ],
            'connection_timeout_seconds' => env('TYPESENSE_CONNECTION_TIMEOUT_SECONDS', 2),
            'healthcheck_interval_seconds' => env('TYPESENSE_HEALTHCHECK_INTERVAL_SECONDS', 30),
            'num_retries' => env('TYPESENSE_NUM_RETRIES', 3),
            'retry_interval_seconds' => env('TYPESENSE_RETRY_INTERVAL_SECONDS', 1),
        ],
        'model-settings' => [
            User::class => [
                'collection-schema' => [
                    'fields' => [
                        [
                            'name' => 'name',
                            'type' => 'string',
                        ],
                        [
                            'name' => 'created_at',
                            'type' => 'int64',
                        ],
                        [
                            'name' => '__soft_deleted', // <==== When scout.soft_delete is set to true, this field is set to 1 if the record is deleted.
                            'type' => 'int32',
                            'optional' => true
                        ],
                    ],
                    'default_sorting_field' => 'created_at',
                ],
                'search-parameters' => [ 
                    'query_by' => 'name, title' // required
                ],
            ],
        ],
    ],
```

## Search Parameters On The Fly

Typesense supports the ability to set almost all search parameters on the fly, if needed. 

To modify search parameters on the fly you can use `setSearchParameters` method of the Typesense Engine.

```php
use App\Models\Todo;

Todo::search('Do grocceries')->setSearchParameters(['query_by' => 'title, description'])->get();
```

Here are some example search parameters:

```php
'highlight_start_tag' => '<mark>',
'highlight_end_tag' => '</mark>',
'snippet_threshold' => 30,
'exhaustive_search' => false,
'use_cache' => false,
'cache_ttl' => 60,
'prioritize_exact_match' => true,
'enable_overrides' => true,
'highlight_affix_num_tokens' => 4,
```

Full list of search parameters and descriptions can be found [here](https://typesense.org/docs/0.25.1/api/search.html#search-parameters).


config/scout.php
```diff
@@ -139,4 +139,38 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | Here you may configure your Typesense settings. Typesense is an open
+    | source search engine with minimal configuration. Below, you can state
+    | the host and key information for your own Typesense installation.
+    |
+    | See: https://typesense.org/docs/0.25.1/api/authentication.html
+    |
+    */
+    'typesense' => [
```
driesvints:
```suggestion
    */

    'typesense' => [
```

config/scout.php
```diff
@@ -139,4 +139,38 @@
...
+        'api_key'         => env('TYPESENSE_API_KEY', 'xyz'),
```
driesvints: Please only use one space before `=>` assignments.

```suggestion
        'api_key' => env('TYPESENSE_API_KEY', 'xyz'),
```

src/EngineManager.php
```diff
@@ -138,6 +140,30 @@ protected function ensureMeilisearchClientIsInstalled()
         throw new Exception('Please install the suggested Meilisearch client: meilisearch/meilisearch-php.');
     }
 
+    /**
+     * @return TypesenseEngine
+     * @throws \Typesense\Exceptions\ConfigError
+     */
```
driesvints: Please use FQN everywhere.

```suggestion
    /**
     * Create a Typesense engine instance.
     * 
     * @return \Laravel\Scout\Engines\TypesenseEngine
     *
     * @throws \Typesense\Exceptions\ConfigError
     */
```

src/Engines/TypesenseEngine.php
```diff
@@ -0,0 +1,498 @@
...
+use Typesense\Exceptions\TypesenseClientError;
+use Typesense\Collection as TypesenseCollection;
+
+/**
+ * Class TypesenseEngine.
+ *
+ * @date    4/5/20
+ *
+ * @author  Abdullah Al-Faqeir <abdullah@devloops.net>
+ */
```
driesvints:
```suggestion
```

src/Engines/TypesenseEngine.php
```diff
@@ -0,0 +1,498 @@
...
+class TypesenseEngine extends Engine
+{
+    /**
+     * @var Typesense
```
driesvints: Please use FQN everywhere and add a one line comment like other docblocks.

src/Engines/TypesenseEngine.php
```diff
@@ -0,0 +1,498 @@
...
+     */
+    private Typesense $typesense;
+
+    /**
+     * @var $searchOptions
```
driesvints:
```suggestion
     * @var array $searchOptions
```

src/Engines/TypesenseEngine.php
```diff
@@ -0,0 +1,498 @@
...
+            'snippet_threshold'          => 30,
+            'exhaustive_search'          => false,
+            'use_cache'                  => false,
+            'cache_ttl'                  => 60,
+            'prioritize_exact_match'     => true,
+            'enable_overrides'           => true,
+            'highlight_affix_num_tokens' => 4,
+        ];
+
+        if (!empty($this->searchOptions)) {
```
driesvints:
```suggestion
        if (! empty($this->searchOptions)) {
```

src/Engines/TypesenseEngine.php
```diff
@@ -0,0 +1,498 @@
...
+            $params = array_merge($params, $this->searchOptions);
+        }
+
+        if (!empty($builder->orders)) {
```
driesvints:
```suggestion
        if (! empty($builder->orders)) {
```

src/Engines/TypesenseEngine.php
```diff
@@ -0,0 +1,498 @@
...
+            if (!empty($params['sort_by'])) {
+                $params['sort_by'] .= ',';
+            } else {
+                $params['sort_by'] = '';
+            }
+            $params['sort_by'] .= $this->parseOrderBy($builder->orders);
```
driesvints:
```suggestion

            $params['sort_by'] .= $this->parseOrderBy($builder->orders);
```

src/Engines/TypesenseEngine.php
```diff
@@ -0,0 +1,498 @@
...
+        }
+
+        return $documents->search($options);
+    }
+
+    /**
+     * @param \Laravel\Scout\Builder $builder
+     * @param int $page
+     * @param int|null $perPage
+     *
```
driesvints, line 486:
```suggestion
```

driesvints (changes requested):
Hi @jasonbosco, thanks for your PR. There's still quite a bit of styling improvements needed here. Can you make sure the code style of this PR adheres to the rest of the library? When done we can mark it as ready for Taylor to review. Thanks

jasonbosco:
@driesvints Thank you for the quick review! 

@karakhanyans has fixed the linting issues and the `continuous-integration/styleci/pr` check now passes. Could you take a look now?

driesvints:
@jasonbosco a lot of the DocBlocks are still missing their first line comment. Could you add those?

driesvints:
Thanks all 👍

taylorotwell:
@jasonbosco You're calling a few model methods I've never heard of:

`getCollectionSchema`
`typesenseQueryBy`

Can you explain both of these methods? What are they supposed to do? Examples? More detail in general please.

I generally need full documentation for how to use this from start to finish. Please mark as ready for review when posted.

jasonbosco:
> `getCollectionSchema` 
> `typesenseQueryBy`
> 
> Can you explain both of these methods? What are they supposed to do? Examples? More detail in general please.

@taylorotwell In the current engine we have, these are methods that needed to be defined in the model. But @karakhanyans and I just discussed this, and we're going to take this opportunity to simplify this interface, and move it into the engine configuration, so these methods won't be required any more. Will share detailed docs once we make these changes.

taylorotwell:
Thanks

jasonbosco:
@taylorotwell We've updated the Typesense engine, and simplified the code to remove any dependencies on custom methods defined in the model. 

I've also updated the description of this PR with docs for the engine. 

Let me know what you think!

CC: @karakhanyans

driesvints:
@jasonbosco looks good. Will let @taylorotwell give the final review of this.

taylorotwell:
@jasonbosco I'm unable to get this working on a fresh Laravel application. I get the following error... have you tested in a real app?

To recreate I just made a new Laravel application using Scout with this PR applied, then uncommented the 10 created users in the `DatabaseSeeder` and ran `php artisan migrate:fresh --seed`.

CleanShot 2023-11-28 at 15 22 54@2x

taylorotwell:
Marking as draft. I tried to correct the error above which led to another error. Please test in fresh Laravel application then mark as ready for review.

jasonbosco:
@taylorotwell We've been looking into this. The root cause is that when we install [typesense-php](https://github.com/typesense/typesense-php) into a Laravel 10 application, it seems to pull in an older version of the package by default (4.0.1). But installing `typesense-php` in Laravel 8 and 9, pulls in the latest version of typesense-php (4.2) correctly. 

We suspect that there's some dependency in typesense-php which conflicts with a Laravel dependency. We're working on trying to identify this. I'll keep you posted.

jasonbosco:
@taylorotwell -

@karakhanyans has fixed the root cause of the issue in typesense-php. The latest code should now pull the correct version of typesense-php and fix the issue you saw.

taylorotwell:
@jasonbosco how do I handle soft deleted records? What do I add to my schema? Furthermore, if I'm using Typesense how can I delete all my collections for a fresh start?

Please mark as ready for review when questions are answered.

jasonbosco:
@taylorotwell 

> how do I handle soft deleted records? What do I add to my schema?

Soft deletes are handled [here](https://github.com/laravel/scout/pull/773/files#diff-30cbc8fb802094edb92485725f009b63a895bcc546b84381b550883a17441129R59-R70). 

Essentially if `soft_delete` is set to `true` in the Scout config, then the engine syncs the [soft delete metadata](https://github.com/laravel/scout/blob/5b5fc4f7a38c879d68e0fd31294672f61bd4c26f/src/Searchable.php#L357-L360) generated by Scout into Typesense. 

So you'd want to add a field called `__soft_deleted` to the Typesense Collection schema to handle soft deletes (see comment below):

```php

...

'typesense' => [
       ...
        'model-settings' => [
            User::class => [
                'collection-schema' => [
                    'fields' => [
                        [
                            'name' => 'name',
                            'type' => 'string',
                        ],
                        [
                            'name' => 'created_at',
                            'type' => 'int64',
                        ],
                        [ /****** Add this field for soft deletes ********************************/
                            'name' => '__soft_deleted', 
                            'type' => 'int32',
                            'optional' => true
                        ],
                    ],
                    'default_sorting_field' => 'created_at',
                ],
                'search-parameters' => [ 
                    'query_by' => 'name, title' 
                ],
            ],
        ],
    ],

```

I'll update the docs in the description with this information shortly.

> Furthermore, if I'm using Typesense how can I delete all my collections for a fresh start?

If you're running on dev, to get a clean slate, you'd want to stop the Typesense server, delete the contents of the `--data-dir` which you specified when starting the process, and then start the process again. 

In a production environment, you'd want to call `DELETE /collections/<name>` on each collection to drop each collection. [Docs](https://typesense.org/docs/0.25.1/api/collections.html#drop-a-collection)

taylorotwell:
@jasonbosco another bug related to soft delete.

If I do this: 

```php
return User::search('Delpha')->withTrashed()->get();
```

I get a result (as expected because Delpha is trashed).

But if I do this:

```php
return User::search('Delpha')->onlyTrashed()->get();
```

I don't get any results. I would expect to still get my result.

taylorotwell:
@jasonbosco drafting this again pending response on above.

jasonbosco:
@taylorotwell - could you give it a shot now? 

@karakhanyans has fixed the soft delete issue and also written a test for it.

taylorotwell:
@jasonbosco @karakhanyans 

OK - thanks. Another question. Let's imagine I have a schema for my user model. I later add another column to the database and want to add it to my schema. I update my Scout configuration file's Typesense schema configuration... then what? I assume the schema isn't automatically updated in Typesense? What is the user story here?

jasonbosco:
@taylorotwell 

> I update my Scout configuration file's Typesense schema configuration... then what? I assume the schema isn't automatically updated in Typesense?

That's correct. Schema changes need to be done in Typesense directly using the [alter schema endpoints](https://typesense.org/docs/0.25.2/api/collections.html#update-or-alter-a-collection).

Alternatively, there's also an [auto-schema detection](https://typesense.org/docs/0.25.2/api/collections.html#with-auto-schema-detection) mode in Typesense, where users can define field names using regex in the collection schema, so at least new field additions matching a pattern do not need schema alterations, but will be auto-added by Typesense when a document is indexed with that field name pattern. 

We can probably add a note to the configuration section in Scout saying that the schema defined here is only used when the collection is created the first time to make this clear?

taylorotwell:
@jasonbosco would it be reasonable to make the default config in Scout reflect auto-schema detection?

jasonbosco:
> would it be reasonable to make the default config in Scout reflect auto-schema detection?

@taylorotwell Setting that to the default could potentially lead to unnecessary fields being indexed and taking up RAM, without the user realizing. For eg, I've commonly seen URL fields getting indexed accidentally, when typically they are only used for display purposes and not for full-text search purposes. 

May be we can call out auto-schema detection as a commented-out field in the example schema, with a link to the docs, so users are aware of its existence?

taylorotwell:
@jasonbosco is there any kind of admin dashboard or something where users can make schema changes? It just feels cumbersome to have to manually craft HTTP calls to update the schema.

jasonbosco:
@taylorotwell 

The dashboard that the core Typesense team maintains is available as part of Typesense Cloud. Happy to set you up with a free account. If you want to sign up and let me know, I can provision a free cluster for you to use. 

There's a Postman collection available [here](https://www.postman.com/typesense/workspace/typesense-api/overview).

There's a community-maintained dashboard [here](https://github.com/bfritscher/typesense-dashboard), but it looks like it doesn't support the alter schema endpoint yet.

taylorotwell:
Thanks 👍

PR #778, opened 2023-11-02
[10.x] Add filter support with comparison operators for the case of Meilisearch Engine

rrakibul:
Add filter support with comparison operators for the case of Meilisearch Engine.

Meilisearch (v1.4) currently supports filtering features with comparison operators `(>, <, >=, <=, TO)`. Laravel scout only supports `=` operation with where clause.

However, in this PR
* I have extended the `where clause` capability by adding an extra optional parameter in the second position to mention the comparison operator. I've handled the case when user will only provide two parameters. This will keep the existing functionality as it is.  
* I've also added new `whereBetween` builder property and functions to facilitate where between comparison.
* Lastly I have modified the filter function for the meilisearch engine to support the comparison operations in the Engines/MeilisearchEngine.php file.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -70,6 +70,13 @@ class Builder
     public $whereNotIns = [];
 
     /**
+     * The "between" constraints added to the query.
+     *
+     * @var array
+     */
+    public $whereBetween = [];
+
+    /**
      * The "limit" that should be applied to the search.
      *
      * @var int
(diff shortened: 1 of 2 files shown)


mmachatschek:
Although I like more configurability I have to say that these changes were already tried multiple times:

https://github.com/laravel/scout/pull/747
https://github.com/laravel/scout/pull/715

Furthermore it doesn't make sense to update the Builder class without supporting the other engines (Algolia, etc.)

driesvints:
I'm sorry @rrakibul but it seems we won't be able to do this right now.

PR #780, opened 2023-11-15
[10.x] Add search engine meta data to results

tobz-nz:
When adding extra options to search queries, xtra data may be return in the raw result - this PR adds the relevant details to the resulting Models via the `withScoutMetadata()` method.

Example:

```php

$products = Product::search(
    $request->term,
    fn ($e, $q, $o) => $e->search($q, array_merge($o, [
        'attributesToHighlight' => ['title'],
        'showRankingScore' => true,
    ]))
)->paginate(48);

dd($products->first()->scoutMetadata());

// output
array:2 [
  "_formatted" => array:31 [
    "title" => "<em>HP</em> <em>Elitebook</em> 1040 G10 14 Inch i5-1335U 4.6GHz 16GB RAM 256GB SSD Laptop with Windows 10/11 Pro"
    ...
  ]
  "_rankingScore" => 0.86111111111111
]
```

### DIFF
--- a/src/Engines/MeilisearchEngine.php
+++ b/src/Engines/MeilisearchEngine.php
@@ -291,7 +291,16 @@ class MeilisearchEngine extends Engine
 
         return $model->getScoutModelsByIds(
             $builder, $objectIds
-        )->filter(function ($model) use ($objectIds) {
+        )->map(function ($model) use ($results, $objectIdPositions) {
+            $result = $results['hits'][$objectIdPositions[$model->getScoutKey()]] ?? [];
+            foreach ($result as $key => $value) {
+                if (substr($key, 0, 1) === '_') {
+                    $model->withScoutMetadata($key, $value);
+                }
+            }
+
+            return $model;
+        })->filter(function ($model) use ($objectIds) {
             return in_array($model->getScoutKey(), $objectIds);
         })->sortBy(function ($model) use ($objectIdPositions) {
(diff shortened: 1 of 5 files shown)


src/Engines/MeilisearchEngine.php
```diff
@@ -291,7 +291,16 @@ public function map(Builder $builder, $results, $model)
 
         return $model->getScoutModelsByIds(
             $builder, $objectIds
-        )->filter(function ($model) use ($objectIds) {
+        )->map(function ($model) use ($results, $objectIdPositions) {
+            $result = $results['hits'][$objectIdPositions[$model->getScoutKey()]] ?? [];
+            foreach ($result as $key => $value) {
```
driesvints:
```suggestion
            $result = $results['hits'][$objectIdPositions[$model->getScoutKey()]] ?? [];

            foreach ($result as $key => $value) {
```

mmachatschek:
@tobz-nz there is a higher chance to get this merged if you could add tests + implement the equivalent code for the algolia search driver

tobz-nz:
@mmachatschek I've added the same functionality to `AgloliaEngine` and added some assertions into the `test_map_correctly_maps_results_to_models` tests for both engines (I hope this is ok and a separate test is not required - it makes so me putting it there)

I've also added the same functionality to the `lazyMap` methods as well.

PR #794, opened 2024-01-23
[Typesense] Fix Search Parameters Issue

karakhanyans:
Fixed the problem with "withSearchParameters" method, removed that method, and used Scout/Builder $options to allow users to add additional options to the query.


src/Engines/TypesenseEngine.php
```diff
@@ -24,17 +24,10 @@ class TypesenseEngine extends Engine
...
-     *
-     * @var array
-     */
-    protected array $searchParameters = [];
-
     /**
      * Create new Typesense engine instance.
      *
-     * @param  \Typesense  $typesense
+     * @param  Typesense  $typesense
```
driesvints, line 30: Can you use FQN here?

src/Engines/TypesenseEngine.php
```diff
@@ -522,30 +515,6 @@ protected function usesSoftDelete($model): bool
...
-     *
-     * @param  array  $options
-     * @return $this
-     */
-    public function setSearchParameters(array $options): static
-    {
-        $this->searchParameters = $options;
-
-        return $this;
-    }
```
driesvints, line 547: Removing these methods would imply a breaking change. Are you sure this is the right call? Everyone who is already using Typesense in Scout now will see their implementation break. But we only launched recently so there might still be a window for this.

driesvints (changes requested):

driesvints:
@karakhanyans I'm gonna have to ask you to resend the PR. Seems something went wrong with branching here and lots of old commits are travelling along. Please also see my remarks above. Thanks!

src/Engines/TypesenseEngine.php
karakhanyans: The thing is, that these methods are inside the Engine and not accessible from the Builder instance. I can leave them inside Engine, just to not introduce any breaking changes.

driesvints: Ah good call. Yeah you can remove them 👍

karakhanyans:
@driesvints did the changes you requested here: https://github.com/laravel/scout/pull/795

PR #802, opened 2024-01-29
[10.x] Fix zero integer value in options parameter for Typesense

alignwebs:
Typesense accepts a zero integer value in a few search option parameters, such as `'drop_tokens_threshold': 0`.

However, when setting it inside `options()` due to the `array_filter()` function, it gets filtered out in the final search query.

```php
Occupation::search($q)->options([
          'limit' => 5,
          'drop_tokens_threshold' => 0,
])->get()
```

### DIFF
--- a/src/Engines/TypesenseEngine.php
+++ b/src/Engines/TypesenseEngine.php
@@ -190,7 +190,7 @@ class TypesenseEngine extends Engine
     {
         return $this->performSearch(
             $builder,
-            array_filter($this->buildSearchParameters($builder, 1, $builder->limit))
+            $this->buildSearchParameters($builder, 1, $builder->limit)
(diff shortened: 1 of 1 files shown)


taylorotwell:
@jasonbosco ?

karakhanyans:
@taylorotwell Let me test this and then we can merge.

karakhanyans:
@taylorotwell the PR verified and confirmed. We can merge it.

alignwebs:
Hey @taylorotwell @karakhanyans, I'm just wondering when the release will happen.

driesvints:
Today

alignwebs:
Thanks for the quick update! @driesvints

PR #808, opened 2024-03-01
fix: permit makeSearchableUsing to realy modify the collection

ArnaudBan:
The `makeSearchableUsing` method is aime for modifying the collection.

 I wanted to use the `load` method on my current collection but for that it need to assign the return to the current object. 

Without this changes you can't really modify the collection, all modifications are not passe to the last call `->update($this->models)`

### DIFF
--- a/src/Jobs/MakeSearchable.php
+++ b/src/Jobs/MakeSearchable.php
@@ -39,6 +39,7 @@ class MakeSearchable implements ShouldQueue
             return;
         }
 
-        $this->models->first()->makeSearchableUsing($this->models)->first()->searchableUsing()->update($this->models);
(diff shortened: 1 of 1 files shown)


driesvints:
There's no tests here.

ArnaudBan:
You are of course rigth. I will try to add tests as soon as i find how to do that. If you have some documentation to setup this package on my machine and run the tests it would help me a lot otherwise a will find how to do this. Thank you.

taylorotwell:
Sorry, I'm not sure why they wouldn't be modified? The `load` method loads the relationships onto the model instance and objects are by-reference in PHP.

PR #812, opened 2024-03-15
[10.x] Optional field definitions on model

MortenDHansen:
Check if field definitions has been added on the model and use if that is the case

**Reasoning**
Search engine would like to have a definition of fields and their type. According to documentation, this is done in a (potentially huge) array in scout config. I propose an optional alternative where you _can_ define the fields alongside your toSearchableArray() method for a searchable model. 

1. This has the benefit of colocation with the properties 
I feel this data belongs on the model rather in the global config, just like the toSearchableArray

2. It opens the option to use php to generate some logic around the schema definition
Consider you want to generate the index with a loop over the toSearchableArray keys detecting the typesense type automatically. 

**Notes**
I propose to replace the field definitions if they are in the model. This ensures there is no breaks with the base approach and the definitions are only used if they are specifically added

I will of course update the documentation accordingly.

### DIFF
--- a/src/Engines/TypesenseEngine.php
+++ b/src/Engines/TypesenseEngine.php
@@ -500,8 +500,9 @@ class TypesenseEngine extends Engine
             return $index;
         } catch (ObjectNotFound $exception) {
             $schema = config('scout.typesense.model-settings.'.get_class($model).'.collection-schema') ?? [];
-            if(method_exists($model, 'typesenseCollectionSchemaFields')) {
-                $schema['fields'] = $model::typesenseCollectionSchemaFields();
+
+            if (method_exists($model, 'typesenseCollectionSchemaFields')) {
+                $schema['fields'] = $model->typesenseCollectionSchemaFields();
             }
 
             if (! isset($schema['name'])) {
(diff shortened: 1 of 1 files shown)


jasonbosco:
> It opens the option to use php to generate some logic around the schema definition

This sounds interesting, but the reason we placed the "index settings" (aka schema definition) for Typesense in the Scout config is to maintain consistency with the other drivers, which place their own index settings in the Scout config as well.

So we wanted to maintain consistency. 

If we make this change in the Typesense driver, we might want to also make similar changes in the other drivers as well...  

But I'll defer to @driesvints or @taylorotwell if we want to have multiple ways to configure index settings - through the Scout config and also through model methods for all the drivers.

MortenDHansen:
> This sounds interesting, but the reason we placed the "index settings" (aka schema definition) for Typesense in the Scout config is to maintain consistency with the other drivers, which place their own index settings in the Scout config as well.
> 
> So we wanted to maintain consistency. 
> 
> If we make this change in the Typesense driver, we might want to also make similar changes in the other drivers as well...  
> 

I would like to do an approach for all drivers if it's interesting, maybe except Algolia who made a separate scout extension completely.

Issue #818, opened 2024-04-06
Typesense driver makes unnecessary api cals


ingria:
### Scout Version

10.8.5

### Scout Driver

Typesense

### Laravel Version

10

### PHP Version

8.3.3

### Description

At the moment, typesence-driver makes two http requests on any model update, even if the changed field is not in the document schema. This causes performance problems if there are a lot of updates.

Real life example: I have a collection with over 10000 documents in it. Each document is updated hourly. Each update changes one model attribute, which triggers 2 request to typesense server. That's 5.5 completely unnecessary requests per second.

This issue has been fixed in Typesense official driver (https://github.com/typesense/laravel-scout-typesense-driver/issues/23), but not in Laravel official driver.

### Steps To Reproduce

1) Create a model with two attributes:

```php
// MyModel.php
public function toSearchableArray(): array {
    return [
        'id' => (string) $this->id,
        'name' => $this->name,
    ];
}

// config/scout.php
[
  'typesense' => [
    'model-settings' => [
      'App\Models\MyModel' => [
        'collection-schema' => [
          'name' => 'my-model',
          'fields' => [
            [
              'name' => 'id',
              'type' => 'string',
            ],
            [
              'name' => 'name',
              'type' => 'string',
            ],
          ],
        ],
      ],
    ],
  ],
];
```

2) Update any non-existent attribute:

```php
MyModel::first()->update([
    'some_field' => 'test',
]);

// scout sends GET to /collections/my-model
// scout sends POST to /collections/my-model/documents/import

MyModel::first()->update([
    'some_date' => now(),
]);

// scout sends GET to /collections/my-model
// scout sends POST to /collections/my-model/documents/import
```

AbdullahFaqeir:
Hey @ingria,

I'll make sure to submit a PR to fix this in the scout official driver.

CC : @jasonbosco

AbdullahFaqeir:
Fixed with #820, should be closed.

PR #820, opened 2024-04-09
- Prevent unnecessary api calls on Collection (index). #818

AbdullahFaqeir:
Prevent unnecessary api calls on Collection (index) when updating/inserting ...etc. #818


AbdullahFaqeir:
I believe we also need to add a testing step for typesense, but I'm honestly not knowledgable in github actions.

PR #836, opened 2024-06-05
Added possibility to version indexes. 

Boorinio:
Hi everyone,
This is a simple pr that adds the indexableAs method in the searchable trait.

Why?
- Rolling updates: It ensures zero downtime by indexing data into a new index while searching continues on the old one. This is something I feel is missing from scout, especially for high traffic websites that want to use it and it is quite easy to achieve this way.
- Backwards compatible: users that don't care for this feature wont be affected since by default the indexableAs and searchableAs have the same return.

Usage scenario:
1. Users Model is searchable
2. You define the indexableAs with return 'users_v2'
3. You reimport everything in users_v2, meanwhile your website can still perform searches on the old index
4. You deploy a new version of your code with the searchableAs returning 'users_v2'
5. profit!

### DIFF
--- a/src/Console/DeleteIndexCommand.php
+++ b/src/Console/DeleteIndexCommand.php
@@ -51,7 +51,7 @@ class DeleteIndexCommand extends Command
     protected function indexName($name)
     {
         if (class_exists($name)) {
-            return (new $name)->searchableAs();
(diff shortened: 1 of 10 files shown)


Boorinio:
Thank you, much appreciated!! @taylorotwell

Issue #831, opened 2024-05-13
[Algolia] Invalid syntax for numeric condition:status=open


yueyuzhao:
### Scout Version

10.9.0

### Scout Driver

Algolia

### Laravel Version

11.7

### PHP Version

8.2.17

### Description

Invalid syntax for numeric condition:status=open {"exception":"[object] (Algolia\\AlgoliaSearch\\Exceptions\\BadRequestException(code: 400): Invalid syntax for numeric condition:status=open at /xxxx/vendor/algolia/algoliasearch-client-php/src/RetryStrategy/ApiWrapper.php:221)

```
Array
(
    [body] => Array
        (
            [numericFilters] => Array
                (
                    [0] => Array
                        (
                            [0] => status=open
                            [1] => status=paid
                        )

                )

            [hitsPerPage] => 15
            [page] => 0
            [query] => 
        )

    [headers] => Array
        (
            [X-Algolia-Application-Id] => xxxxxx
            [X-Algolia-API-Key] => xxxxxx
            [User-Agent] => Algolia for PHP (3.4.1); PHP (8.2.17); Guzzle (7); Laravel Scout (10.9.0)
            [Content-Type] => application/json
        )

    [method] => POST
    [query] => Array
        (
        )

    [retryNumber] => 1
    [host] => https://xxxxx-dsn.algolia.net/1/indexes/dev_order_index/query
)
```

The input filters are placed in `numericFilters`, but those are not numeric values, and should be place in `filters` as `status:open OR status:paid`

### Steps To Reproduce

1. use algolia engine
2. create an index
3. search with `whereIn` described in https://laravel.com/docs/11.x/scout#where-clauses

yueyuzhao:
For those who want to search with filters, you can set filters by `options` as:
```php
$orders = Order::search('Star Trek')->options([
    'filters' => 'status:open OR status:paid',
])->get();
```

driesvints:
Could you share an exact code sample to reproduce this?

yueyuzhao:
@driesvints 
1. Set the engine to algolia
2. Set `User` model searchable (or any searchable model)
3. Put the following code to your `routes/console.php`
4. Run `php artisan scout-where-in`
```php
// ......
use App\Models\User;
// ......

Artisan::command('scout-where-in', function () {
    User::search('hi')
        ->whereIn('status', ['open', 'paid'])
        ->get();
});

```
And you will get 
```

[...]

PR #837, opened 2024-06-11
Removed numericFilters from AlgoliaEngine.

Boorinio:
Hopefully fixes 
https://github.com/laravel/scout/issues/831
@driesvints

### DIFF
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -105,7 +105,7 @@ class AlgoliaEngine extends Engine
     public function search(Builder $builder)
     {
         return $this->performSearch($builder, array_filter([
-            'numericFilters' => $this->filters($builder),
+            'filters' => $this->filters($builder),
             'hitsPerPage' => $builder->limit,
         ]));
     }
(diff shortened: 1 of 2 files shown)


taylorotwell:
The docs say `where` is for numeric equality checks.

Boorinio:
> The docs say `where` is for numeric equality checks.

Indeed, which is inconsistent with how meilisearch works (which supports everything). The docs also mention this
```
$orders = Order::search('Star Trek')->whereIn(
    'status', ['open', 'paid']
)->get();

```
Which currently doesn't work

driesvints:
@taylorotwell there is indeed a deeper underlying issue here with inconsistency between drivers. I feel like we have an opportunity here to align the different drivers more and at the same time make the Algolia search more powerful. Could we reconsider?

taylorotwell:
I guess we'll give it a shot. Will have to roll it back if it breaks something.

taylorotwell:
@driesvints we need to test this on a real Laravel application before it can be released.

Boorinio:
> I guess we'll give it a shot. Will have to roll it back if it breaks something.

Thanks for merging this, if it does break something you can comment here, I will be happy to do a follow up!

driesvints:
@Boorinio unfortunately we had to revert this one as it's incompatible with https://github.com/algolia/scout-extended which still expects numericFilters. Can you re-attempt the PR to 11.x and also add an entry to the upgrade guide?

PR #840, opened 2024-06-20
Fix searchable queue configuration

tklie:
This PR fixes how the `Searchable` trait determines whether to queue jobs. Also, the published configuration file is updated to be compatible with the current version of the trait.

---

The `Searchable` trait currently determines the queue connection and queue using the following two methods:

```php
    {
    }

    {
    }
```

However, the config file that is published by this package only contained the following config option:

```php
return [
    'queue' => env('SCOUT_QUEUE', false),
];
```

This has been updated to the keys actually used by the trait:

```php
return [
    'queue' => [
        'connection' => env('SCOUT_CONNECTION'),
        'queue' => env('SCOUT_QUEUE', false),
    ],
]
```

Subsequently, another update to the `Searchable` trait itself was necessary, since this used the old `scout.queue` key to determine if jobs should be queued. It now correctly uses `scout.queue.queue`:

```php
    public function queueMakeSearchable($models)
    {
        if ($models->isEmpty()) {
            return;
        }

        if (! config('scout.queue.queue')) {
            $models->first()->makeSearchableUsing($models)->first()->searchableUsing()->update($models);
            return;
        }

        dispatch((new Scout::$makeSearchableJob($models))
                ->onQueue($models->first()->syncWithSearchUsingQueue())
                ->onConnection($models->first()->syncWithSearchUsing()));
    }
```

Edit: The same was done for the `queueRemoveFromSearch` method.

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -42,7 +42,10 @@ return [
     |
     */
 
-    'queue' => env('SCOUT_QUEUE', false),
+    'queue' => [
+        'connection' => env('SCOUT_CONNECTION'),
+        'queue' => env('SCOUT_QUEUE', false),
+    ],
 
     /*
     |--------------------------------------------------------------------------
--- a/src/Searchable.php
+++ b/src/Searchable.php
@@ -59,8 +59,9 @@ trait Searchable
             return;
         }
 
-        if (! config('scout.queue')) {
-            return $models->first()->makeSearchableUsing($models)->first()->searchableUsing()->update($models);
+        if (! config('scout.queue.queue')) {
+            $models->first()->makeSearchableUsing($models)->first()->searchableUsing()->update($models);
+            return;
         }
 
         dispatch((new Scout::$makeSearchableJob($models))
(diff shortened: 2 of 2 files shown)


taylorotwell:
Would break all existing apps.

tklie:
Will be fixed in future versions though? Because as it stands right now, the `Searchable` trait is inconsistent within itself, using two different config keys.

To clarify: You can use it as is right now by setting the entire `scout.queue` option in your config file to `false`. But having the config file like this

```php
return [
    'queue' => [
        'connection' => env('SCOUT_CONNECTION'),
        'queue' => env('SCOUT_QUEUE', false),
    ],
]
```

and disabling queueing via an env variable is not possible right now, because `scout.queue` will always be truthy since it's an array.

bedus-creation:
@taylorotwell it would be nice, if we could fix this by adding another options possibly like connection ?

```php
'connection' => env('SCOUT_CONNECTION')
```

and 
```php
 {
    return config('scout.connection') ?: config('queue.default');
}
```

PR #841, opened 2024-06-21
[11.x] Use scout prefix when deleting all indexes

macbookandrew:
I use Laravel Scout and Meilisearch with several local development sites and use `SCOUT_PREFIX` to avoid collisions between apps.

I recently wanted to drop and recreate the indexes for one app so used `php artisan scout:delete-all-indexes` assuming that it would delete just the indexes for the current app…but it actually deleted _all_ indexes in the Meilisearch server.

This PR respects the `scout.prefix` config value and only deletes the indexes that begin with the prefix.

The `php artisan scout:delete-index` is _not_ affected, since you have to supply the index name and it’s in theory more obvious which index you are deleting.

Alternative idea instead of this approach: the `scout:delete-all-indexes` command could instead notify the user that it will delete _all_ indexes, not just those with this app’s prefix, and ask them to confirm before deleting.

Alternative idea in conjunction with this approach: the `scout:delete-all-indexes` command could include a `--ignore-prefix` or similar flag to override the behavior in this PR for those cases where the user does in fact want to remove all indexes, even those without the prefix.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,6 +6,10 @@
 
 In previous Scout releases, the Algolia engine utilized `numericFilters` to power `where` conditions. However, `numericFilters` does not support simple string matching. In Scout 11.x, `filters` is now used instead of `numericFilters`.
 
+### Meilisearch Engine `scout:delete-all-indexes` command
+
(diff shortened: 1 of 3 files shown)


driesvints:
@macbookandrew thank you for your PR. Could you add an entry to the `UPGRADE.md` guide for this one?

macbookandrew:
@driesvints done

PR #846, opened 2024-07-10
[Typesense]  Sync server state in getOrCreateCollectionFromModel #845

tharropoulos:
Updates the `getOrCreateCollectionFromModel` method in `TypesenseEngine` to verify if a collection exists on the server. If not found, it is created. This ensures consistency between the worker and server, fixing queue workers not being able to import after flushing, as mentioned in #845.

Issue #845, opened 2024-07-08
Typesense Import: Index Race Condition


saibotk:
### Scout Version

10.9.0

### Scout Driver

Typesense

### Laravel Version

10.48.10

### PHP Version

8.3.9

### Database Driver & Version

Typesense 0.25.2

### SDK Version

Typesense PHP 4.9.3

### Description

When deleting the index and afterwards importing all models, we can reach a race condition, where the import fails because some jobs run in parallel and all try to create or access the index.
 
Only one succeeds and the others will, for example, throw this error:
`Typesense\Exceptions\ObjectNotFound: Not Found`
or interestingly, only rarely, `Typesense\Exceptions\ObjectAlreadyExists: A collection with name `<something>` already exists.`.

This happens when using queued imports. I don't exactly understand why it returns Not Found instead of the more rare collection already exists error, which is logical, but this might be some technical detail in the typesense engine.

E.g. using Horizon with `minProcesses = 5` and running the following command on a table with enough entries e.g. ~ 40.000 will sometimes result in those exceptions on some of the jobs.

We should try to make the import more robust and maybe ensure that the index is created before the documents are imported? I'm happy to hear your solutions on this, but i think we should fix this flaw in scout directly.

### Steps To Reproduce

In any scout repository with Horizon or some queue workers and a table with enough elements such that chunking happens, run:
```bash
php artisan scout:delete-index "App\Models\User" && php artisan scout:import "App\Models\User"
```

And sometimes the race condition will cause some jobs to fail.

driesvints:
@jasonbosco @AbdullahFaqeir

saibotk:
Seems like #822 is related

saibotk:
I just did some more deep-diving:

- `scout:flush` in typesense does also delete the schema / index entirely on the server, but the local `exists` bool is never reset to null, so it will assume that it exists. This is one of the issues that we need to solve. I think this is the cause of the ObjectNotFound exception. I found this by using flush instead of delete-index.
- the other one still needs investigation, maybe there are different code paths and some weird race conditions on what horizon process has which state etc. idk

A small workaround:
When calling `artisan scout:flush` or `scout:delete-index` run this afterwards (maybe create a wrapper artisan command):
```php
use Laravel\Scout\EngineManager;
use App\Models\Commodity;

$ts = app(EngineManager::class)->engine();
$model = new Commodity();

$ts->flush($model);
$ts->getCollections()->offsetUnset($model->searchableAs());
``` 

[...]

PR #847, opened 2024-07-12
refactor(typesense): remove unused exists checks

saibotk:
In https://github.com/laravel/scout/pull/820 we introduced an exists check method, which then lead to static state issues as described in https://github.com/laravel/scout/issues/845

Those where addressed in https://github.com/laravel/scout/pull/846

But now that the function always double-checks the existence, we can remove the exists check entirely and only rely on the Typesense server response for this state.

This "fixes" issues where the server already has a collection and the client would try to recreate it. E.g. when it has flushed the index and another process or worker then creates the collection.

### DIFF
--- a/src/Engines/TypesenseEngine.php
+++ b/src/Engines/TypesenseEngine.php
@@ -497,6 +497,7 @@ class TypesenseEngine extends Engine
     protected function getOrCreateCollectionFromModel($model, bool $indexOperation = true): TypesenseCollection
     {
         $method = $indexOperation ? 'indexableAs' : 'searchableAs';
+
         $collectionName = $model->{$method}();
         $collection = $this->typesense->getCollections()->{$collectionName};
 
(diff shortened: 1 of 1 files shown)


saibotk:
@tharropoulos could you review this, just to be sure?

saibotk:
We could even remove the 
`$collection->setExists(true)`
after creation, or i would also need to add `$collection->setExists(true)` right before returning the existing collection, to be sure that this field is set.

What are the plans regarding the exists field in the typesense sdk?

EDIT: Updated my code to also set the exists field when the server has the collection to be 100% correct.

tharropoulos:
Thanks for iterating upon the PR! I'll have to check it first thing on Monday, as I'm away from my laptop for this weekend. I'll get back to you as soon as I can

tharropoulos:
Looking back on it, the verbosity used on the previous PR might be overkill, and after spending some time messing with it, it seems to work just fine. 


CC: @jasonbosco  could you also take a look at the code yourself? Looks just fine to me

saibotk:
I feel like this is pretty much ready otherwise, so i will mark this as ready :)

PR #849, opened 2024-07-17
Sort models after fetching when orders conditions are applied to search

lubatti:
I've found that after applying a custom order logic to the search engine and retrieving a subset of ids, these ids can be re-sorted during the fetching of the models.


crynobone:
This is only supported by MySQL and SQLite and will cause syntax error in Postgres and MSSQL

taylorotwell:

PR #854, opened 2024-08-09
Add clear-cache option to the Flush command

xyluz:
The flush command currently only removes the saved records from the database (or whatever data storage is used), it does not 'flush' records in the cache. As the name suggests, flush should remove the record from the db, and clear cache (probably more of a personal opinion - but since it's not flush-from-storage, it's just flush). 

This is useful because (from my work experience), after flushing records, I always clear cache. And it'll be useful to just have that option along side the flush command. I'm not entirely sure how many people do the same, but i would personally find this useful.

Hopefully someone agrees. 

it's a simple addition to the codebase, and i hope this explanation above is sufficient, otherwise, please let me know if there are any feedback/questions regarding this.

**Fingers crossed**

### DIFF
--- a/src/Console/FlushCommand.php
+++ b/src/Console/FlushCommand.php
@@ -31,7 +31,7 @@ class FlushCommand extends Command
     {
         $class = $this->argument('model');
         $clearCache = $this->option('clear-cache');
-        
+
         $model = new $class;
 
         $model::removeAllFromSearch();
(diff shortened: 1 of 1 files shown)


driesvints:
v10 is closed for new features. Please attempt v11.

taylorotwell:

Issue #824, opened 2024-04-25
Incorrect Total Count in Laravel Scout Pagination


Evenerik:
### Scout Version

10.8.6

### Scout Driver

Typesense

### Laravel Version

10.48.3

### PHP Version

8.3.0

### Database Driver & Version

MySQL 8.0.36

### Description

This is pretty much a reopening of @razvaniacob's issue, who has encountered the same problem I am currently facing. I won't go into too much detail since he described the bug perfectly well in [his issue](https://github.com/laravel/scout/issues/819), but Scout's `LengthAwarePaginator` shows wrong total when using Typesense's engine

### Steps To Reproduce

See steps in https://github.com/laravel/scout/issues/819

driesvints:
@karakhanyans do you have any thoughts here?

karakhanyans:
@driesvints will check this and get back with more info 👌

raz-iacob:
I will be happy to see this solved guys. Good luck!

driesvints:
@karakhanyans any news here?

AbdullahFaqeir:
I'll be working on this

tamakoma1129:
Hello,  

I encountered the same issue reported here. In my case, I was using a model with a ULID as the primary key but initially forgot to set protected $keyType = 'string'; in my model. After updating this setting, I can confirm that the pagination now works correctly.

Here are the specifics of my development environment:
- Laravel Version: 10.48.4
- Scout Version: 10.8.6
- PHP Version: 8.1.27
- Scout Driver: Meilisearch
- Database: MySQL 8.0.36  

Apologies if this solution has already been mentioned. I hope this helps someone else experiencing the same problem.  

Thank you.

driesvints:
Ping @jasonbosco

jasonbosco:
@AbdullahFaqeir will be working on this going forward.

AbdullahFaqeir:
Hi @Evenerik!

Can you please dump the collection schema, and sample record in typesense and the same record in database.

Below you can see a dump of my env showing a working fresh sample, I've seeded 10K users.

# User model settings.

<img width="389" alt="screenshot 7" src="https://github.com/laravel/scout/assets/1428547/796086e1-403b-44d8-997e-4931344d2fe3">

# Search performed.

<img width="288" alt="screenshot 8" src="https://github.com/laravel/scout/assets/1428547/d4e16990-dbea-4476-9dae-72cf3e7098f6">

# Paginated search results.

[...]

PR #858, opened 2024-08-26
[Typesense] Fix Paginate Function Returning Limited Records in Laravel Scout with Typesense Engine (#824)

tharropoulos:
This pull request addresses an issue where the paginate function in Laravel Scout with the Typesense engine always returns a number of records limited to the perPage value, rather than reflecting the actual total number of matches from the Typesense server. This behavior was traced back to the limit attribute being set to the perPage value, which restricted the number of records returned. (#824)

## Changes Made:

* Removed the assignment of the `perPage` value to the limit attribute in the `TypesenseEngine` class to allow the pagination to reflect the actual total number of matches.

## Rationale:

The current behavior prevents users from receiving all the results they expect based on the actual total number of matches on the Typesense server. By removing the forced assignment of the `perPage` value to the `limit` attribute, we allow the pagination to work correctly, returning the correct number of records.

### DIFF
--- a/src/Engines/TypesenseEngine.php
+++ b/src/Engines/TypesenseEngine.php
@@ -205,8 +205,6 @@ class TypesenseEngine extends Engine
      */
     public function paginate(Builder $builder, $perPage, $page)
     {
-        $builder->take($builder->limit ?? $perPage);
-
         return $this->performSearch(
             $builder,
             $this->buildSearchParameters($builder, $page, $perPage)


Braunson:
Is there a bigger issue with this though? For example, this code works:

```php
$results = Names::search($this->keyword)
    ->options([
        'filter_by' => $this->buildFilterString(),
        'sort_by' => implode(',', $this->sortBy),
    ])
    ->paginate(48);
```

But if I drop in this line between options and paginate:
```php
->query(fn (Builder $query) => $query->with('someRelation'))
```

We then get this error back from Typesense:
```
Only up to 250 hits can be fetched per page.
```

The only solution to solve that is to again set the `per_page` option to match the `->paginate()` value.

PR #863, opened 2024-09-08
feat(typesense): add 'infix' option to search parameters

TitusKirch:
This PR adds the infix option to the Typesense search parameters in Laravel Scout.

The Infix Search option allows users to perform more flexible searches by matching substrings in the middle of words. This provides greater accuracy and control over search results, especially when dealing with partial matches. It improves the search experience in applications where infix matching is critical, such as autocomplete features or when searching for terms that may not be at the beginning or end of a string.

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -184,6 +184,7 @@ return [
             //             [
             //                 'name' => 'name',
             //                 'type' => 'string',
+            //                 'infix' => true,
             //             ],
             //             [
             //                 'name' => 'created_at',
@@ -193,7 +194,8 @@ return [
             //         'default_sorting_field' => 'created_at',
             //     ],
             //     'search-parameters' => [
-            //         'query_by' => 'name'
+            //         'query_by' => 'name',
+            //         'infix' => 'allways',
             //     ],
             // ],
         ],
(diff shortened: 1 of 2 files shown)


jasonbosco:
@TitusKirch You can already set any search parameters (including `infix`)  dynamically like this: https://laravel.com/docs/11.x/scout#typesense-dynamic-search-parameters

taylorotwell:

TitusKirch:
> @TitusKirch You can already set any search parameters (including `infix`) dynamically like this: https://laravel.com/docs/11.x/scout#typesense-dynamic-search-parameters

@jasonbosco  thanks for your quick reply. I know that you can set it via dynamic search parameters using the `options` method. However, it is not possible to set them globally depending on the `query_by` parameter. If you set the following globally:

```php
<?php
$options = [
    'search-parameters' => [
        'query_by' => 'title, description',
    ],
];
```

Now I would have to append the `options` method in every search operation to achieve this, as the default is `off`.

Or am I missing something?

TitusKirch:
I have found a way to avoid having to redefine this every time. If someone has the same problem, you can do the following:

In the respective model you can define a method `typesenseSearchParameters`. E.g. as follows:

```PHP
<?php

namespace App\Models;

use Laravel\Scout\Searchable;

class User extends Model
{
    use Searchable;

    //...
    
    /**
     * Return the Typesense search parameters for the model.
     */
    public function typesenseSearchParameters(): array
    {
        return [
            'query_by' => 'name',
            'infix' => 'always',
        ];
    }
    
    //...
}
```
This then also works. See also in the `TypesenseEngine`: https://github.com/laravel/scout/blob/10.x/src/Engines/TypesenseEngine.php#L262-L264

Issue #881, opened 2024-11-13
Cannot use object of type Algolia\\AlgoliaSearch\\Config\\SearchConfig as array


brocard:
### Scout Version

10.11.6

### Scout Driver

Algolia

### Laravel Version

11

### PHP Version

8.2

### Description

After insert in model with Searchable trait i receveid thir error

` "message": "Cannot use object of type Algolia\\AlgoliaSearch\\Config\\SearchConfig as array",
    "exception": "Error",
    "file": "/Users/brocard/Projects/VALLARTA/AppKovayVacations/AppKovayApi/vendor/laravel/scout/src/Engines/Algolia3Engine.php",
    "line": 42,`

### Steps To Reproduce

Just after insert using the model with the trait or run the command `php artisan scout:import Model Namespace`

brocard:
To fix this, we can use the methods provided by the SearchConfig object to access its properties, assuming they are accessible through methods or public properties. Here’s how you might approach it:

Inspect the SearchConfig object: Check which methods or properties are available. You can do this by running dd(get_class_methods($config)); to see the available methods or dd($config); to see its properties.

Replace array-style access with method calls: If, for example, getConnectTimeout() is a method of SearchConfig, you would replace $config['connect_timeout'] with $config->getConnectTimeout().

crynobone:
Hey there, thanks for reporting this issue.

We'll need more info and/or code to debug this further. Can you please create a repository with the command below, commit the code that reproduces the issue *as one separate commit* on the main/master branch and share the repository here?

Please make sure that you have the [latest version of the Laravel installer](https://github.com/laravel/installer) in order to run this command. Please also make sure you have both Git & [the GitHub CLI tool](https://cli.github.com/) properly set up.

    laravel new bug-report --github="--public"

Do not amend and create a separate commit with your custom changes. After you've posted the repository, we'll try to reproduce the issue.

Thanks!

brocard:
This line its broken too, because $config->getAppId() its return an array and not an string 

<img width="945" alt="Screenshot 2024-11-13 at 0 23 58" src="https://github.com/user-attachments/assets/5ab81d78-98d0-43ba-b892-5974e5eca62e">

Here is the problem the function create has $appid, and $apikey as parameter, and this moment is reciving and array
<img width="397" alt="Screenshot 2024-11-13 at 0 35 22" src="https://github.com/user-attachments/assets/5894939d-1919-4404-ae21-28edf6cab26a">

[...]

PR #882, opened 2024-11-13
Fix Cannot use object of type Algolia\\AlgoliaSearch\\Config\\SearchConfig as array #881

brocard:
fixed the error on Cannot use object of type Algolia\\AlgoliaSearch\\Config\\SearchConfig as array #881

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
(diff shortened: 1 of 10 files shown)


dwightwatson:
I don't think this is the right fix - you are just getting the connect timeout, read timeout, write timeout and batch size off the Algolia 3 config and setting it again.

Also, the PR includes your PHPStorm files.

crynobone:
Fixed via #884

PR #894, opened 2024-12-07
fix: numericFilters will be overwritten if softDelete = true

tomocrafter:
When additional options is set by `options` method on `Builder`, those value will be overwritten by following value:

https://github.com/laravel/scout/blob/b000080cf75da9ee5f7b936ca5df868e0aec5aef/src/Engines/AlgoliaEngine.php#L94

which results in lost all `numericFilters` manually set with `options` method.

This PR fixes it by using recursive version of array_merge to merge all `numericFilters`.

Thank you.


taylorotwell:
Potential breaking change.

PR #903, opened 2025-01-25
Update Algolia config to replace ALGOLIA_SECRET with ALGOLIA_API_KEY

yehuuu6:
Official Algolia docs give us this;
ALGOLIA_APP_ID=
ALGOLIA_API_KEY=

But the Laravel Scout has different fields for this. We should update this.


taylorotwell:
I don't see a big reason to change this atm.

PR #912, opened 2025-02-17
Resolve Race Condition In TypesenseEngine Collection Creation

### LINKED ISSUE #911: Hitting Race Condition In TypesenseEngine Due To Jobs Competing To Create Collection
### Scout Version

10.12.2

### Scout Driver

Typesense

### Laravel Version

11.31

### PHP Version

8.2

### Database Driver & Version

MariaDB

### SDK Version

4.9

### Meilisearch CLI Version

_No response_

### Description

I've identified a race condition taking place when attempting to do `scout:flush` and `scout:import`.  As best as I can tell the issue lies here:

https://github.com/laravel/scout/blob/10.x/src/Engines/TypesenseEngine.php#L618

as two competing jobs dispatched and picked up close enough, can run afoul of the behavior identified here.  I think a better solution would be to create a mutex to check to see if any other job is attempting to create it, and then wait for it to be created.

### Steps To Reproduce

I used a chunk size of about 1000, and had it being dispatched and pickup by horizon.

```php

$this->call('scout:flush', [
    'model' => User::class),
]);
$this->call('scout:import', [
    'model' => User::class,
]);
```

andyleach:
Resolves: https://github.com/laravel/scout/issues/911

### DIFF
--- a/src/Engines/TypesenseEngine.php
+++ b/src/Engines/TypesenseEngine.php
@@ -11,6 +11,7 @@ use Laravel\Scout\Builder;
 use stdClass;
 use Typesense\Client as Typesense;
 use Typesense\Collection as TypesenseCollection;
+use Typesense\Exceptions\ObjectAlreadyExists;
 use Typesense\Exceptions\TypesenseClientError;
 
 class TypesenseEngine extends Engine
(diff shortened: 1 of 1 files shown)


tharropoulos:
Yeah that's a way to do it as well hahaha, thanks!

PR #918, opened 2025-03-14
Add support for importing and flushing all models

grantholle:
This PR adds the ability to flush and import all models configured in a scout config. This is helpful to run a single command to handle all your model updates, such as during deployments when you may have adjusted model settings and need to have them re-indexed.

It checks to see if you have included the now optional `model` argument or new `--all` flag, falling back to the argument. If neither are provided, a `RuntimeException` is thrown alerting the user to the missing `model` argument.


taylorotwell:

PR #919, opened 2025-03-25
fix Builder.php - not actually using returned results from afterRawSearchCallback

The-Rabak:
Fix using afterRawSearchCallback without actually getting back the returned results. 
If the purpose of the afterRawSearchCallback is to allow changes to be made on preprocessed raw results then this part was probably left out by mistake.

I believe this was the intended use for this callback that was unintentionally missed.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -564,7 +564,7 @@ class Builder
     public function applyAfterRawSearchCallback($results)
     {
         if ($this->afterRawSearchCallback) {
-            $results = call_user_func($this->afterRawSearchCallback, $results);
+            $results = call_user_func($this->afterRawSearchCallback, $results) ?: $results;
         }
 
         return $results;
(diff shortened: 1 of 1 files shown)


src/Builder.php
```diff
@@ -564,7 +564,7 @@ protected function getTotalCount($results)
     public function applyAfterRawSearchCallback($results)
     {
         if ($this->afterRawSearchCallback) {
-            call_user_func($this->afterRawSearchCallback, $results);
+            $results = call_user_func($this->afterRawSearchCallback, $results);
```
gdebrauwer: @The-Rabak If the afterRawSearchCallback does not return anything, then it should still use the original $result value (similar to the afterQuery-method on eloquent's query builder). Otherwise, this will be a breaking change.
```suggestion
            $results = call_user_func($this->afterRawSearchCallback, $results) ?: $results;
```

The-Rabak:
@gdebrauwer thank you for the catch, you're definitely right, applied and committed your suggestion.

PR #929, opened 2025-07-05
[10.x] Add `scout:queue` command to optimize large imports

stevebauman:
## Description

This PR implements a new `scout:queue` command to optimize the importing of millions of rows into the configured engine.

It works by executing only two queries to retrieve the `MIN` and `MAX` range of the given searchable model, generating ranges based on the given chunk size, and dispatching a job to import the models between each range. This makes it significantly faster than the existing `scout:import` command because we're not actually iterating through the entire database table to dispatch the jobs. This means that there will be of course potential gaps between these ranges in the database, so jobs may not index the exact chunk size (or in some cases, index anything at all if the gaps exceed chunk size).

This command also assumes that the model's scout key is an auto-incremented ID. Though, this should cover a large majority of developer's applications using Scout. If a developer attempts to queue a model with a scout key that is not numeric, then the command is aborted with an error.

At my work, we have over 30 million rows that we're indexing at the moment, and the current `scout:import` takes us _days_ to finish, making it difficult to schedule without downtime. This command allows us to effectively spread the load horizontally in our queue worker and get our index back up and running much faster.

Thanks for your time, and let me know your thoughts!

## Usage

**Queuing all models**:

```php
php artisan scout:queue "App\Models\Post"
```

**Queuing by 1000**:

```php
php artisan scout:queue "App\Models\Post" --chunk=1000
```

**Queuing a custom range**:

```php
php artisan scout:queue "App\Models\Post" --min=10 --max=100
```

### DIFF
--- a/src/Console/QueueCommand.php
+++ b/src/Console/QueueCommand.php
@@ -16,6 +16,8 @@ class QueueCommand extends Command
      */
     protected $signature = 'scout:queue
             {model : Class name of model to bulk queue}
+            {--min= : The minimum ID to start queuing from}
+            {--max= : The maximum ID to queue up to}
             {--c|chunk= : The number of records to queue in a single job (Defaults to configuration value: `scout.chunk.searchable`)}';
 
     /**
@@ -38,8 +40,10 @@ class QueueCommand extends Command
 
         $query = $model::makeAllSearchableQuery();
 
-        $min = $query->min($model->getScoutKeyName());
-        $max = $query->max($model->getScoutKeyName());
+        $min = $this->option('min') ?? $query->min($model->getScoutKeyName());
+        $max = $this->option('max') ?? $query->max($model->getScoutKeyName());
+
+        $chunk = max(1, (int) ($this->option('chunk') ?? config('scout.chunk.searchable', 500)));
 
         if (! $min || ! $max) {
             $this->info('No records found for ['.$class.']');
@@ -53,12 +57,8 @@ class QueueCommand extends Command
             return;
         }
 
-        $chunkSize = $this->option('chunk') ?? config('scout.chunk.searchable', 500);
-
-        $chunkSize = max(1, (int) $chunkSize);
-
-        for ($start = $min; $start <= $max; $start += $chunkSize) {
-            $end = min($start + $chunkSize - 1, $max);
+        for ($start = $min; $start <= $max; $start += $chunk) {
+            $end = min($start + $chunk - 1, $max);
 
             dispatch(new MakeRangeSearchable($model, $start, $end))
                 ->onQueue($model->syncWithSearchUsingQueue())
--- a/src/Console/QueueCommand.php
+++ b/src/Console/QueueCommand.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Laravel\Scout\Console;
+
+use Illuminate\Console\Command;
+use Laravel\Scout\Jobs\MakeRangeSearchable;
+use Symfony\Component\Console\Attribute\AsCommand;
+
+#[AsCommand(name: 'scout:queue')]
+class QueueCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'scout:queue
+            {model : Class name of model to bulk queue}
+            {--c|chunk= : The number of records to queue in a single job (Defaults to configuration value: `scout.chunk.searchable`)}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Import the given model into the search index';
+
+    /**
+     * Execute the console command.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $class = $this->argument('model');
+
+        $model = new $class;
+
+        $query = $model::makeAllSearchableQuery();
+
+        $min = $query->min($model->getScoutKeyName());
+        $max = $query->max($model->getScoutKeyName());
+
+        if (! $min || ! $max) {
+            $this->info('No records found for ['.$class.']');
+
+            return;
+        }
+
+        if (! is_numeric($min) || ! is_numeric($max)) {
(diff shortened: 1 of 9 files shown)


src/Jobs/MakeRangeSearchable.php
```diff
@@ -0,0 +1,65 @@
...
+    public function handle()
+    {
+        $models = $this->model::makeAllSearchableQuery()
+            ->whereBetween($this->model->getScoutKeyName(), [$this->start, $this->end])
+            ->get();
+
+        dispatch(new Scout::$makeSearchableJob($models))
+            ->onQueue($this->model->syncWithSearchUsingQueue())
+            ->onConnection($this->model->syncWithSearchUsing());
+    }
```
stevebauman, line 70: In order to preserve functionality between this command and `scout:import`, I'm dispatching the `$makeSearchableJob` here instead of synchronously indexing using `searchable()`, as developers may have created their own `$makeSearchableJob` with custom logic.

src/Console/QueueCommand.php
```diff
@@ -0,0 +1,72 @@
...
+
+        $model = new $class;
+
+        $query = $model::makeAllSearchableQuery();
+
+        $min = $query->min($model->getScoutKeyName());
+        $max = $query->max($model->getScoutKeyName());
+
+        if (! $min || ! $max) {
+            $this->info('No records found for ['.$class.']');
```
devajmeireles, line 45:
```suggestion
            $this->components->info('No records found for ['.$class.']');
```

devajmeireles:
```suggestion
            $this->components->info('No records found for ['.$class.']');
```

src/Console/QueueCommand.php
```diff
@@ -0,0 +1,72 @@
...
+
+            return;
+        }
+
+        if (! is_numeric($min) || ! is_numeric($max)) {
+            $this->error('The primary key for ['.$class.'] is not numeric.');
```
devajmeireles:
```suggestion
            $this->components->error('The primary key for ['.$class.'] is not numeric.');
```

src/Console/QueueCommand.php
```diff
@@ -0,0 +1,72 @@
...
+            $end = min($start + $chunkSize - 1, $max);
+
+            dispatch(new MakeRangeSearchable($model, $start, $end))
+                ->onQueue($model->syncWithSearchUsingQueue())
+                ->onConnection($model->syncWithSearchUsing());
+
+            $this->line('<comment>Queued ['.$class.'] models up to ID:</comment> '.$end);
+        }
+
+        $this->info('All ['.$class.'] records have been queued.');
```
devajmeireles:
```suggestion
        $this->components->info('All ['.$class.'] records have been queued.');
```

devajmeireles (commented):
Suggesting the usage of `$this->components->`.

src/Console/QueueCommand.php
stevebauman: This would be inconsistent with the other commands in the package:

https://github.com/laravel/scout/blob/747532be6ed01ad510733065048dbaa59d4e1499/src/Console/ImportCommand.php#L51

I'll leave this up to the team if they want this 👍

taylorotwell:
Thanks!

DougSisk:
Was facing similar import issues, so thanks for this!

Unfortunately I'm getting an error when running this with Horizon and Redis. Seems to be an issue with `SerializesModels` on `MakeRangeSearchable`. It's trying to load a model from the database and it's erroring out since there isn't an ID set.

vinothkumar95:
👋 Just adding context for MongoDB users:

Right now, php artisan scout:queue-import only supports numeric primary keys (using range-based chunking). This makes it incompatible with MongoDB’s ObjectId, which causes the error:

`The primary key for [App\Models\YourModel] is not numeric.`

👉 Would it be possible to add support for non-numeric primary keys?
That would make queue-import usable across both SQL and MongoDB drivers.

stevebauman:
> Right now, php artisan scout:queue-import only supports numeric primary keys (using range-based chunking). This makes it incompatible with MongoDB’s ObjectId, which causes the error:

Yes this is detailed in the PR:

> **This command also assumes that the model's scout key is an auto-incremented ID.** Though, this should cover a large majority of developer's applications using Scout. If a developer attempts to queue a model with a scout key that is not numeric, then the command is aborted with an error.

> Would it be possible to add support for non-numeric primary keys?
> That would make queue-import usable across both SQL and MongoDB drivers.

How do you propose we do that?

vinothkumar95:
For MongoDB (and other drivers using non-numeric keys like ObjectId), we could add a fallback strategy:

If the Scout key is numeric, keep the current range-based chunking (fast path).

If the Scout key is not numeric, fall back to using Eloquent’s chunkById() (or even plain chunk()) — this works with ObjectId and any UUID-style primary key.

That way:
SQL users with auto-increment IDs still get the performance boost.
MongoDB / UUID users still get compatibility without hitting the “not numeric” error.

stevebauman:
> If the Scout key is not numeric, fall back to using Eloquent’s chunkById() (or even plain chunk()) — this works with ObjectId and any UUID-style primary key.

This is already being done in the existing `scout:import` command and would not improve performance:

https://github.com/laravel/scout/blob/230179272d5eb5dfd39948aa87bf307b62195412/src/Console/ImportCommand.php#L52

https://github.com/laravel/scout/blob/230179272d5eb5dfd39948aa87bf307b62195412/src/Searchable.php#L173-L176

https://github.com/laravel/scout/blob/230179272d5eb5dfd39948aa87bf307b62195412/src/SearchableScope.php#L34-L42

The performance improvement in this new command is due to avoiding iterating through the whole database table all together. If we implemented chunking here, there would be no difference between it and the previous command.

You may use the previous command (`scout:import`) if you are using non-numerically indexed rows.

vinothkumar95:
Thanks @stevebauman

PR #930, opened 2025-07-11
[10.x] Remove SerializesModels from MakeRangeSearchable

DougSisk:
This is a fix for PR #929.

The `SerializesModels` trait does not play well with an empty Eloquent model class. It's attempting to load a model and erroring out since there's no ID set. This removes that trait and changes the job to use the class' fully qualified name to avoid any errors from PHP attempting to serialize the class.


stevebauman:
Thanks @DougSisk!

PR #931, opened 2025-07-19
feat: can disable typesense prefix

Barbapapazes:
Hello 👋,

Typesense offers semantic search capabilities using embeddings. However, the `prefix` search option is not compatible with `vector` (embedding-based) search. This PR enables users to set the `prefix` option to `false` when performing a `vector` search.

Relevant documentation: https://typesense.org/docs/29.0/api/search.html#query-parameters:~:text=Indicates%20that%20the,for%20all%20fields).

Current error when using `vector` search with `prefix` set to `true`:

```txt
Prefix search is not supported for remote embedders. Please set `prefix=false` as an additional search parameter to disable prefix searching.
```

### DIFF
--- a/src/Engines/TypesenseEngine.php
+++ b/src/Engines/TypesenseEngine.php
@@ -345,6 +345,7 @@ class TypesenseEngine extends Engine
             'prioritize_exact_match' => true,
             'enable_overrides' => true,
             'highlight_affix_num_tokens' => 4,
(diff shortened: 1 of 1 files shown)


tharropoulos (commented):
Can't directly approve of changes here, but it looks right

Barbapapazes:
I was wondering why the parameters array couldn't be merged with the one in the configuration to avoid to add manually every parameters from typesense while being future proof?

PR #934, opened 2025-07-23
Feature: add typesense remote embedding settings

SoongMingLiang:
This PR adds the capability of customizing the behavior of typesense performing remote embedding through specifying parameters in config/scout.php file.

Related documentation:
https://typesense.org/docs/29.0/api/vector-search.html#remote-embedding-api-parameters

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -204,6 +204,11 @@ return [
             //     ],
             // ],
         ],
+        'remote-embedding-settings' => [
+            'remote_embedding_batch_size' => env('TYPESENSE_REMOTE_EMBEDDING_BATCH_SIZE', 200),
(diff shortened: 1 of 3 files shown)


taylorotwell:
I need a review from Typesense on this one. @tharropoulos

tharropoulos (commented):
These are runtime parameters on the `import` API, so adding them in the scout config falls in line with how we already handle the rest of the collection configuration. 

They're being initialized to their defaults, so it's correct.

SoongMingLiang:
Hi @taylorotwell , just a quick nudge on this PR. Whenever you have a moment, would love to hear your feedback. Thanks!

PR #935, opened 2025-07-30
Add scout:import-all command to import all Searchable models

ChristianJoniel:
This PR implements a new `scout:import-all` command to automatically import all models that contain the Searchable trait, or all the models that were specified in the scout.php config file.

I've worked in multiple projects where I've had to re-import all models, and run the php artisan scout:import command per each model. This PR resolves this, by automatically discovering the models that have the Searchable trait.
If you want to specify which models should be imported when running the command, then in the scout.php config file there is a models key with an empty array where you can specify the models you want this command to import.

Let me know your thoughts!
Thanks

## Summary
This PR adds a new `scout:import-all` command that automatically discovers and imports all searchable models to the search index, eliminating the need to run `scout:import` individually for each model.

## Problem
In projects with multiple searchable models, developers currently need to run `php artisan scout:import` for each model individually when re-indexing data.

## Solution
The new command provides two ways to import models:

1. **Auto-discovery** (default): Automatically finds all models using the `Searchable` trait
2. **Configuration-based**: Specify models in `config/scout.php`

## Usage
```bash
php artisan scout:import-all

```

## Configuration (Optional)
```php
// config/scout.php
'models' => [
    App\Models\User::class,
    App\Models\Post::class,
    App\Models\Product::class,
],
```

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -18,6 +18,10 @@ return [
 
     'driver' => env('SCOUT_DRIVER', 'algolia'),
 
+    'models' => [
+        // App\Models\User::class,
+    ],
+
     /*
     |--------------------------------------------------------------------------
     | Index Prefix
--- a/config/scout.php
+++ b/config/scout.php
@@ -18,6 +18,17 @@ return [
 
     'driver' => env('SCOUT_DRIVER', 'algolia'),
 
+    /*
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #939, opened 2025-08-22
fix(scout): Add configurable queue connection and queue name options

faraweilyas:
# Fix Scout queue configuration options

## Description

This PR fixes Laravel Scout's queue configuration by adding `queue_connection` and `queue_name`, allowing developers to specify custom queue connections and queue names while maintaining full backward compatibility.

Checkout the queue [doc section](https://laravel.com/docs/12.x/scout#queueing) to spot the bug.

## Changes Made

- **Expand queue configuration options**: Added `'queue_connection' => env('SCOUT_QUEUE_CONNECTION'),` and `'queue_name' => env('SCOUT_QUEUE_NAME'),` options
- **Update queue config getter in Searchable trait**: Modified `syncWithSearchUsing()` and `syncWithSearchUsingQueue()` methods to use `config('scout.queue_connection')` and `config('scout.queue_name')`
- **Add environment variable support**: Introduced `SCOUT_QUEUE_CONNECTION` and `SCOUT_QUEUE_NAME` for easier configuration management
- **Maintain backward compatibility**: Existing `SCOUT_QUEUE=true/false` continues to work as expected

## Benefits to End Users

- **Greater flexibility**: Developers can now route Scout operations to specific queue connections (e.g., separate Redis instance for search indexing)
- **Better queue organization**: Ability to use dedicated queue names for Scout operations, improving monitoring and job separation
- **Easier deployment management**: Environment-based configuration makes it simple to use different queue setups across environments
- **Zero breaking changes**: Existing applications continue to work without any code changes

## Code Changes

**config/scout.php file updates:**
```php
// Before
'queue' => env('SCOUT_QUEUE', false),

// After
'queue' => env('SCOUT_QUEUE', false),

'queue_connection' => env('SCOUT_QUEUE_CONNECTION'),

'queue_name' => env('SCOUT_QUEUE_NAME'),
```

**src/Searchable.php file updates:**
```php
// Before
{
}

// After  
{
    return config('scout.queue_connection') ?: config('queue.default');
}

// Before
{
}

// After  
{
    return config('scout.queue_name');
}
```

**src/Console/QueueImportCommand.php file updates:**
```php
// Before
{--queue= : The queue that should be used (Defaults to configuration value: `scout.queue.queue`)}'

// After  
{--queue= : The queue that should be used (Defaults to configuration value: `scout.queue_name`)}'
```

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -44,6 +44,10 @@ return [
 
     'queue' => env('SCOUT_QUEUE', false),
 
+    'queue_connection' => env('SCOUT_QUEUE_CONNECTION'),
+
+    'queue_name' => env('SCOUT_QUEUE_NAME'),
+
     /*
     |--------------------------------------------------------------------------
     | Database Transactions
--- a/src/Console/QueueImportCommand.php
+++ b/src/Console/QueueImportCommand.php
@@ -19,7 +19,7 @@ class QueueImportCommand extends Command
             {--min= : The minimum ID to start queuing from}
             {--max= : The maximum ID to queue up to}
             {--c|chunk= : The number of records to queue in a single job (Defaults to configuration value: `scout.chunk.searchable`)}
-            {--queue= : The queue that should be used (Defaults to configuration value: `scout.queue.queue`)}';
+            {--queue= : The queue that should be used (Defaults to configuration value: `scout.queue_name`)}';
 
     /**
      * The console command description.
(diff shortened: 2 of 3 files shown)


taylorotwell:
I think your original approach of using an array can still work but you just need to adjust the code that reads those values to account for both cases... I think that would be cleaner.

PR #947, opened 2025-10-09
Full-text improvements

taylorotwell:
This makes a couple of improvements to full-text support when using the `database` engine.

- Just pass multiple columns to `orWhereFullText` in a single call.
- When using Postgres and no other developer provided ordering has been defined, automatically add relevance based ordering by default to match MySQL's built-in ordering behavior.

### DIFF
--- a/src/Engines/DatabaseEngine.php
+++ b/src/Engines/DatabaseEngine.php
@@ -260,7 +260,7 @@ class DatabaseEngine extends Engine implements PaginatesEloquentModelsUsingDatab
 
         return $query->orderByRaw(
             sprintf(
-                "ts_rank(".$vectors.", %s(?)) desc",
+                'ts_rank('.$vectors.', %s(?)) desc',
                 match ($this->getFullTextOptions($builder)['mode'] ?? 'plainto_tsquery') {
                     'phrase' => 'phraseto_tsquery',
                     'websearch' => 'websearch_to_tsquery',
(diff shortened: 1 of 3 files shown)


src/Engines/DatabaseEngine.php
```diff
@@ -196,9 +205,11 @@ protected function initializeSearchQuery(Builder $builder, array $columns, array
             return $query;
         }
 
-        return $query->where(function ($query) use ($builder, $columns, $prefixColumns, $fullTextColumns) {
-            $connectionType = $builder->model->getConnection()->getDriverName();
+        [$connectionType] = [
+            $builder->model->getConnection()->getDriverName(),
```
MeiKatz: Shouldn't this be `$builder->modelConnectionType()` ?

src/Engines/DatabaseEngine.php
```diff
@@ -196,9 +205,11 @@ protected function initializeSearchQuery(Builder $builder, array $columns, array
             return $query;
         }
 
-        return $query->where(function ($query) use ($builder, $columns, $prefixColumns, $fullTextColumns) {
-            $connectionType = $builder->model->getConnection()->getDriverName();
+        [$connectionType] = [
+            $builder->modelConnectionType(),
+        ];
```
MeiKatz, line 210: Sorry to bother you again but why are you converting the model connection type to an array with one element just to retrieve the first and only element right after? Wouldn't it be easier to assign the connection type directly?

```php
$connectionType = $builder->modelConnectionType();
```

taylorotwell: Yeah I was assigning multiple variables here in an earlier version.

PR #949, opened 2025-10-31
Feat: Laravel-friendly model lookup

macaws:
This simple PR allows users to type commands such as `php artisan scout:import User` instead of `php artisan scout:import "App\\Models\\User"` by doing a quick lookup if the user's input is not immediately found.

It seems more in-line with other Artisan CLI commands from Laravel without breaking existing functionality.

I'm not sure if the PHPDoc `@throws` tag is wanted or not - PHPStorm suggested it after adding a throw. I *did* need to update the test, since it was excepting an `Error` class and we now pass a ScoutException, which seems more appropriate in this situation and should be easier for a dev to debug since these commands run within CI/CD, forge deploy scripts, crons, etc.


macaws:
Also, this may warrant a docs update

taylorotwell:
Can re-submit this without the one-line stuff?

PR #955, opened 2025-12-13
Typesense Import Action - default to upsert

dbpolito:
## Problem

The default action was changed from `upsert` to `emplace` in https://github.com/laravel/scout/pull/936

According to the Typesense docs (https://typesense.org/docs/29.0/api/documents.html#index-multiple-documents):
> Creates a new document or updates an existing document if a document with the same id already exists. You can send either the whole document or a partial document for update.

While `emplace` supports partial updates, this causes a critical issue: **when removing optional fields from a document, those fields are not deleted from the index**. Instead, they retain their old values and remain searchable.

### Example

Initial document:
```php
[
   'id' => 1,
   'a' => 'a',
   'b' => 'b',
]
```

After update (removing field `b`):
```php
[
   'id' => 1,
   'a' => 'a',
]
```

I even experienced some cases that even sending the field with null values would still not update it, when using `enable_nested_fields`.

```php
[
   'id' => 1,
   'a' => 'a',
   'b' => null,
]
```

**Expected:** Field `b` should be removed from the index
**Actual:** Field `b` still contains `'b'` and remains searchable

## Solution

This PR adds a config option to change the action, but reverts the default back to `upsert` to prevent this issue.

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -204,6 +204,7 @@ return [
             //     ],
             // ],
         ],
+        'import_action' => env('TYPESENSE_IMPORT_ACTION', 'upsert'),
     ],
 
 ];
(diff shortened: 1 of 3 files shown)


dbpolito:
hey @SoongMingLiang

pinging you here since you introduced the other PR to get your thoughts

PR #961, opened 2026-01-29
[10.x] Separate scout key name from database column name

### LINKED ISSUE #960: scout:import fails when getScoutKeyName() differs from primary key
### Scout Version

10.23.0

### Scout Driver

Meilisearch

### Laravel Version

12.38.1

### PHP Version

8.2

### Database Driver & Version

MySQL 8.0 via Laravel Sail

### SDK Version

1.16.1

### Meilisearch CLI Version

_No response_

## Problem

  `getScoutKeyName()` is currently used in two different contexts:
  1. Field name in search index (Meilisearch, Algolia)
  2. Column name in database queries (SQL)

  This causes failures when a model needs different names for each context.

  ## Why this matters

  Consider two models indexed together:

  | Model | Database PK | Index field |
  |-------|-------------|-------------|
  | `User` | `id` (int) | `id` |
  | `Invitation` | `uuid` (string) | `id` |

  Both models have `id` field in Meilisearch, but in database it's `id` or `uuid` depending on the model. This is where it breaks - `getScoutKeyName()` cannot
  return both `id` (for index) and `uuid` (for database) at the same time.

  ## Affected code

  | Location | Uses `getScoutKeyName()` for |
  |----------|------------------------------|
  | `MeilisearchEngine::update()` | Index field name ✓ |
  | `SearchableScope::searchable()` | `chunkById()` column ✗ |
  | `Searchable::queryScoutModelsByIds()` | `whereIn()` column ✗ |
  | `QueueImportCommand::handle()` | `min()`/`max()` column ✗ |
  | `DatabaseEngine::searchModels()` | `orderBy()` column ✗ |

### Steps To Reproduce

  1. Create two models with different primary keys:

  ```php
  // User.php - standard id
  class User extends Model
  {
      use Searchable;

      protected $primaryKey = 'id';
  }

  // Invitation.php - uuid primary key
  class Invitation extends Model
  {
      use Searchable;

      protected $primaryKey = 'uuid';
      protected $keyType = 'string';

      {
          return 'id'; // Need 'id' in Meilisearch
      }

      public function getScoutKey()
      {
          return $this->uuid;
      }

      {
          return [
              'id' => $this->uuid, // Map uuid to 'id' field
              // ...
          ];
      }
  }
```

  2. Run import:

  `php artisan scout:import "App\Models\Invitation"`

  3. Result:

```
  SQLSTATE[42S22]: Column not found: 1054 Unknown column 'id' in 'where clause'

  The chunkById() uses getScoutKeyName() which returns id, but database column is uuid.
  ```

waszkowski:
Fixes #960

## Problem

When a model overrides `getScoutKeyName()` to return a value different from `getKeyName()` (the primary key), database queries break. This is because `getScoutKeyName()` was used both as the key name sent to the search engine **and** as the column name in database queries (`orderBy`, `whereBetween`, `chunkById`, etc.).

For example, a model with `uuid` as the primary key (`getKeyName()`) but `id` as the scout key (`getScoutKeyName()`) would generate invalid SQL queries referencing a non-existent `id` column.

### Before
```php
// Model with uuid primary key and custom scout key
public function getKeyName() { return 'uuid'; }
public function getScoutKeyName() { return 'id'; }

// Queries incorrectly use 'id' column instead of 'uuid':
// SELECT * FROM models ORDER BY models.id  ← fails
```

### After
```php
// New method separates the database column from the scout key:
public function getScoutKeyColumnName() { return $this->getKeyName(); } // 'uuid'
public function getScoutKeyName() { return 'id'; } // sent to search engine

// Queries correctly use 'uuid' column:
// SELECT * FROM models ORDER BY models.uuid  ← works
```

## Solution

Introduce `getScoutKeyColumnName()` method on the `Searchable` trait that defaults to `getKeyName()`. All internal database queries now use `getScoutKeyColumnName()` instead of `getScoutKeyName()`, keeping the two concerns separated:

- `getScoutKeyName()` → key name for the search engine index
- `getScoutKeyColumnName()` → column name for database queries

Updated locations: `Searchable`, `SearchableScope`, `DatabaseEngine`, `QueueImportCommand`, `MakeRangeSearchable`, `RemoveFromSearch`.

## Tests

- Verifies `getScoutKeyColumnName()` defaults to `getKeyName()`
- Verifies a model with different key names returns correct values from each method
- Verifies `queryScoutModelsByIds` uses column name, not scout key name
- Verifies `makeAllSearchable` orders by column name

### DIFF
--- a/src/Console/QueueImportCommand.php
+++ b/src/Console/QueueImportCommand.php
@@ -48,8 +48,8 @@ class QueueImportCommand extends Command
 
         $query = $model::makeAllSearchableQuery();
 
-        $min = $this->option('min') ?? $query->min($model->getScoutKeyName());
-        $max = $this->option('max') ?? $query->max($model->getScoutKeyName());
+        $min = $this->option('min') ?? $query->min($model->getScoutKeyColumnName());
+        $max = $this->option('max') ?? $query->max($model->getScoutKeyColumnName());
 
         $chunk = max(1, (int) ($this->option('chunk') ?? config('scout.chunk.searchable', 500)));
 
--- a/src/Engines/DatabaseEngine.php
+++ b/src/Engines/DatabaseEngine.php
@@ -81,7 +81,7 @@ class DatabaseEngine extends Engine implements PaginatesEloquentModelsUsingDatab
                 }
             })
             ->when(! $this->getFullTextColumns($builder), function ($query) use ($builder) {
-                $query->orderBy($builder->model->getTable().'.'.$builder->model->getScoutKeyName(), 'desc');
+                $query->orderBy($builder->model->getTable().'.'.$builder->model->getScoutKeyColumnName(), 'desc');
             })
             ->when($this->shouldOrderByRelevance($builder), function ($query) use ($builder) {
                 $this->orderByRelevance($builder, $query);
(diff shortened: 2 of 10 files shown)


taylorotwell:
Not sure I want to add another configuration for this. I would just probably just put an auto-incrementing column and a UUID uniquely indexed column on the model so it has both.

PR #967, opened 2026-02-07
[10.x] Ensure MakeSearchable is only dispatched when required

jackbayliss:
Currently the `MakeSearchable` Job is dispatched regardless of what attributes were updated.

This means, if we're only indexing certain attributes via the `toSearchableArray()`, updating anything outside of those still triggers a reindex.

This becomes extremely inefficient and also costly when doing large bulk updates of other attributes, as it can re-index thousands of items for no reason.

This PR ensures:
- Newly created items are still indexed as expected
- We only dispatch the job if we need to  / only index on the engine if its required

All other tests and new tests pass as expected. 

No breaking change as the default `toSearchableArray()` returns `toArray()`, so `wasChanged()` will work as expected for models that haven't customized it. The behavior only changes for models that have explicitly narrowed their `toSearchableArray()`, which is what I would expect.

There is `searchIndexShouldBeUpdated` but, this is not documented - and this PR feels (to me) like it's a sensible default behavior. 

Let me know if you need any changes!

### DIFF
--- a/src/ModelObserver.php
+++ b/src/ModelObserver.php
@@ -106,6 +106,10 @@ class ModelObserver
             return;
         }
 
+        if (! $model->wasRecentlyCreated && ! $model->wasChanged(array_keys($model->toSearchableArray()))) {
+            return;
+        }
+
         $model->searchable();
     }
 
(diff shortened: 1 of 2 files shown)


jackbayliss:
I realised customised array keys won't update if someone does: 


```
{
    return [
        'name' => $this->name,
        'full_address' => $this->street . ', ' . $this->city,
    ];
}
```

So.. I guess I'll just update the docs

PR #968, opened 2026-02-10
[10.x] Run Typesense integration tests in github action

gdebrauwer:
Based on https://typesense.org/docs/guide/github-actions.html, I updated the GitHub action to run the Typesense integration tests.

I also replaced the `@group ...` docblocks with PHPUnit attributes, since no Meilisearch or Typesense integration tests ran on PHP 8.3/8.4 due to the usage of docblocks. 
Screenshot 2026-02-10 at 10 18 57


gdebrauwer:
@tharropoulos the test that is failing in the Typesense integration group was added by you (https://github.com/laravel/scout/pull/924), but I'm not sure why it fails? 🤔 

edit: never mind, I was able to fix the test.

PR #969, opened 2026-02-10
[11.x] Add support for `where($field, $operator, $value)` on `Builder` class

gdebrauwer:
Currently, there is no easy way to, for example, search the models that were created after a certain date. 
```php
// Before
User::search('foo', function ($engine, $query, options) {
	// You have to try to manipulate the $options to add where statements with operators (<, >, <=, >=, !=).
	// The $options variable is also very different for each engine, so you have to write an engine-specific implementation
})->get();
```
This is because the `where()` method on the Scout `Builder` class only accepts 2 parameters: a field and a value. This PR fixes that by allowing you to use an operator.
```php
// After
User::search('foo')->where('created_at', '>', now()->timestamp)->get();
```

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,29 @@
 
 ## Upgrading To 11.0 From 10.x
 
+### Builder `wheres` Property and Custom Engines
+
+In previous Scout releases, the `wheres` property on the `Builder` instance was a simple key / value associative array. In Scout 11.x, the `wheres` property is now an array of arrays, with each entry containing `field`, `operator`, and `value` keys. This change was made to support comparison operators such as `>`, `<`, `>=`, `<=`, and `!=` via the `where` method:
+
+```php
+User::search('*')->where('age', '>', 30)->get();
+```
+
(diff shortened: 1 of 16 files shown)


taylorotwell:
Does this break every custom engine that expects the `wheres` to be in a different format? If so, would need to go to master branch or be adjusted.

gdebrauwer:
This is indeed a breaking change, that's why I targeted the 11.x branch (there is no master branch in this repository and the current version is 10.x).
Every custom engine will need to be adjusted, but the changes in each engine are not that big, as you can see in the 3 engines that ship with Scout. You just have to change how you loop over the $wheres array.

taylorotwell:
Ah, got it. Didn't notice we were already on the next version branch.

crynobone:
@gdebrauwer Can you check latest `11.x` branch, I had to mark both `test_it_can_filter_with_where_comparisons` on `tests/Integration/MeilisearchSearchableTest.php` and `tests/Integration/TypesenseSearchableTest.php` after merging `10.x` to `11.x`, thank you

See #977

gdebrauwer:
@crynobone I created a PR: https://github.com/laravel/scout/pull/978

PR #970, opened 2026-02-10
[10.x] Make it easy to run Typesense en Meilisearch integration tests locally

gdebrauwer:
Currently, there is no easy way to quickly run the Typesense and Meilisearch integration tests when you are making changes to this package: either you already have Typesense and Meilisearch instances on your machine, or you have to trigger the GitHub actions.

This PR fixes that by adding a docker-compose file and a script to run the tests using those Docker containers on your local machine.
```bash
sh ./bin/test.sh
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,32 @@
+version: '3'
+services:
+    typesense:
+        image: 'typesense/typesense:30.1'
+        ports:
+            - '8108:8108'
+        environment:
+            TYPESENSE_DATA_DIR: '/typesense-data'
+            TYPESENSE_API_KEY: 'xyz'
+            TYPESENSE_ENABLE_CORS: true
+        volumes:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+            timeout: 7s
+    meilisearch:
+      image: 'getmeili/meilisearch:latest'
+      ports:
+          - '7700:7700'
+      healthcheck:
+          test: ["CMD", "wget", "--no-verbose", "--spider",  "http://localhost:7700/health"]
+          retries: 3
+          timeout: 5s
+volumes:
```
cirolosapio, line 30: no meilisearch volume?

gdebrauwer: Adding a meilisearch volume was not needed to get meilisearch running. To get Typesense running, I was required to add a volume.

PR #971, opened 2026-02-19
Use Algolia's filters parameter instead of numericFilters for string filter support

alecgarcia:
## Summary

- Migrates Algolia filter generation from the `numericFilters` parameter to the unified `filters` parameter, which supports both numeric comparisons and facet/string matching
- Fixes string-based filters (e.g. `whereNotIn('status', ['deleted'])`) which previously failed because `numericFilters` only accepts numeric values
- Uses proper Algolia filter syntax: `key=value` for numeric, `key:"value"` for strings, `NOT key:"value"` for string negation, `(... OR ...)` for whereIn groups, joined with `AND`
- Adds comprehensive test coverage for string where/whereIn/whereNotIn, mixed numeric+string filters, quote escaping, floats, zero values, and single-value whereIn

## Test plan

- [x] All existing Algolia v3 and v4 feature tests updated and passing
- [x] New tests added for: string where, string whereIn, string whereNotIn, single-value whereIn, quote escaping, float values, zero/soft-delete values, no-filter search
- [x] Full test suite passes (206 tests)
- [x] Verified filter string format against live Algolia API

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -99,7 +99,7 @@ abstract class AlgoliaEngine extends Engine implements UpdatesIndexSettings
     public function search(Builder $builder)
     {
         return $this->performSearch($builder, array_filter([
-            'numericFilters' => $this->filters($builder),
+            'filters' => $this->filters($builder),
             'hitsPerPage' => $builder->limit,
         ]));
     }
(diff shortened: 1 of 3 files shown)


taylorotwell:
Hey there - I think it would be better to look at this on `11.x` branch as where handling changed in that upcoming major version.

alecgarcia:
@taylorotwell I will look at the changes in `11.x` but I should have mentioned I made this PR because PR #959 actually broke our scout setup not realizing our `whereNotIn` was not actually applying. We were passing in a string and when we did an update we started throwing exceptions. This PR was to get the wheres to actually match up with what the documentation shows in https://laravel.com/docs/12.x/scout#where-clauses

alecgarcia:
Looks like commits `81e699c` and `d8b1be5` on `11.x` make similar changes.

PR #998, opened 2026-06-11
Add PostgreSQL search engine to Scout

WendellAdriel:
This adds a native PostgreSQL search engine to Laravel Scout, giving applications that already run on PostgreSQL a way to perform full-text (and optional trigram similarity) search without standing up a separate search service.

The engine is registered as the `pgsql` driver and uses PostgreSQL's built-in `tsvector` / `tsrank` functions for indexing and ranking. It optionally blends trigram similarity when the `pg_trgm` extension is available, configurable via the same `scout.php` config file used by the other drivers.

Documentation: https://github.com/laravel/docs/pull/11235

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -214,25 +214,25 @@ return [
     |
     | Here you may configure Scout's native PostgreSQL search driver. The driver
     | uses PostgreSQL full-text search by default and may optionally blend in
-    | trigram similarity when the pg_trgm extension is available.
+    | trigram similarity scoring when the pg_trgm extension is available.
     |
     */
 
     'pgsql' => [
-        'language' => env('SCOUT_PGSQL_LANGUAGE', 'english'),
-        'vector_column' => env('SCOUT_PGSQL_VECTOR_COLUMN', 'search_vector'),
+        'language' => 'english',
+        'vector_column' => 'search_vector',
         'column_weights' => [],
-        'query_function' => env('SCOUT_PGSQL_QUERY_FUNCTION', 'plainto_tsquery'),
-        'rank_function' => env('SCOUT_PGSQL_RANK_FUNCTION', 'ts_rank'),
+        'query_function' => 'plainto_tsquery',
+        'rank_function' => 'ts_rank',
         'trigram' => [
-            'enabled' => env('SCOUT_PGSQL_TRIGRAM', false),
-            'threshold' => env('SCOUT_PGSQL_TRIGRAM_THRESHOLD', 0.3),
+            'enabled' => env('PGSQL_TRIGRAM', false),
+            'threshold' => env('PGSQL_TRIGRAM_THRESHOLD', 0.3),
             'columns' => [],
-            'create_extension' => env('SCOUT_PGSQL_CREATE_TRIGRAM_EXTENSION', false),
+            'create_extension' => false,
         ],
         'weights' => [
-            'full_text' => env('SCOUT_PGSQL_FULL_TEXT_WEIGHT', 1.0),
-            'trigram' => env('SCOUT_PGSQL_TRIGRAM_WEIGHT', 0.25),
+            'full_text' => 1.0,
+            'trigram' => 0.25,
         ],
     ],
 
--- a/config/scout.php
+++ b/config/scout.php
@@ -227,10 +227,7 @@ return [
         'trigram' => [
             'enabled' => env('SCOUT_PGSQL_TRIGRAM', false),
             'threshold' => env('SCOUT_PGSQL_TRIGRAM_THRESHOLD', 0.3),
-            'extension' => [
-                'enabled' => env('SCOUT_PGSQL_TRIGRAM_EXTENSION', false),
-                'create' => env('SCOUT_PGSQL_CREATE_TRIGRAM_EXTENSION', false),
-            ],
+            'create_extension' => env('SCOUT_PGSQL_CREATE_TRIGRAM_EXTENSION', false),
         ],
         'weights' => [
             'full_text' => env('SCOUT_PGSQL_FULL_TEXT_WEIGHT', 1.0),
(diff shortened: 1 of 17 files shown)


tests/Integration/PgsqlSearchableTest.php
```diff
@@ -0,0 +1,226 @@
...
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+use Laravel\Scout\Searchable;
+use Orchestra\Testbench\Attributes\RequiresEnv;
+use Orchestra\Testbench\Concerns\WithWorkbench;
+use Orchestra\Testbench\TestCase;
+use PHPUnit\Framework\Attributes\Group;
+
+#[Group('pgsql')]
+#[RequiresEnv('SCOUT_PGSQL_TEST_DATABASE')]
```
benbjurstrom: should we add a github workflow for this?

src/Pgsql/Trigram.php
```diff
@@ -0,0 +1,159 @@
...
+
+        if (array_key_exists($key, $this->availability)) {
+            return $this->availability[$key];
+        }
+
+        try {
+            $result = $connection->selectOne(
+                "select exists (select 1 from pg_extension where extname = 'pg_trgm') as available"
+            );
+        } catch (Throwable) {
```
benbjurstrom: Can we just catch QueryException?

src/Pgsql/Trigram.php
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Get the indexable trigram predicate for the query.
+     *
+     * @param  array  $columns
+     * @return string
+     */
+    public function predicateExpression(array $columns)
+    {
+        return empty($columns) ? 'false' : sprintf(
```
benbjurstrom, line 156: According to Claude, Postgres reads the threshold setting at the start of the index scan, so this set_config may not run until after the index has already filtered rows using the default threshold.

benbjurstrom (commented):
Looks really good, only left a couple of notes. Nice work!

config/scout.php
```diff
@@ -207,4 +207,33 @@
...
+        'rank_function' => env('SCOUT_PGSQL_RANK_FUNCTION', 'ts_rank'),
+        'trigram' => [
+            'enabled' => env('SCOUT_PGSQL_TRIGRAM', false),
+            'threshold' => env('SCOUT_PGSQL_TRIGRAM_THRESHOLD', 0.3),
+            'columns' => [],
+            'create_extension' => env('SCOUT_PGSQL_CREATE_TRIGRAM_EXTENSION', false),
+        ],
+        'weights' => [
+            'full_text' => env('SCOUT_PGSQL_FULL_TEXT_WEIGHT', 1.0),
+            'trigram' => env('SCOUT_PGSQL_TRIGRAM_WEIGHT', 0.25),
```
pushpak1300: nit: All the env names of driver start with respective driver name so for consistency we can use `PGSQL_*`

config/scout.php
```diff
@@ -207,4 +207,33 @@
...
+    | trigram similarity when the pg_trgm extension is available.
+    |
+    */
+
+    'pgsql' => [
+        'language' => env('SCOUT_PGSQL_LANGUAGE', 'english'),
+        'vector_column' => env('SCOUT_PGSQL_VECTOR_COLUMN', 'search_vector'),
+        'column_weights' => [],
+        'query_function' => env('SCOUT_PGSQL_QUERY_FUNCTION', 'plainto_tsquery'),
+        'rank_function' => env('SCOUT_PGSQL_RANK_FUNCTION', 'ts_rank'),
```
pushpak1300: Nit: For some variables, we shouldn’t provide overloaded environment variables.  Most of the time, the default value will be correct and users won’t need to change most of it. We can simply add the default value and then in the published file, users can override it. I feel most of these values shouldn’t be managed via environment variables and should be shared across the entire application, for example: https://github.com/laravel/laravel/blob/a7cc4fe0d0eba5980c29c6c1dd91f81f09032a5b/config/database.php#L87

config/scout.php
```diff
@@ -11,7 +11,7 @@
     | using Laravel Scout. This connection is used when syncing all models
     | to the search service. You should adjust this based on your needs.
     |
-    | Supported: "algolia", "meilisearch", "typesense",
+    | Supported: "algolia", "meilisearch", "typesense", "pgsql",
```
pushpak1300, line 14: also this is very optional but just to add different opinion. pgsql can be get confused with database driver. we should think a specific name. again not super important or highly opiniated i think.

src/Pgsql/Trigram.php
pushpak1300, line 1: we should somehow throw Logging alert or something when user run `scout:import` with pgsql driver.

pushpak1300: or we can run this migration behind the scenes with scout:import so behaviour and ux wise some similarities are there.

> Setting `SCOUT_DRIVER` to `pgsql` does not create any database columns or indexes. You should add Scout's PostgreSQL search vector to each searchable table using a migration.

config/scout.php
```diff
@@ -207,4 +207,33 @@
...
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | PostgreSQL Configuration
+    |--------------------------------------------------------------------------
+    |
+    | Here you may configure Scout's native PostgreSQL search driver. The driver
+    | uses PostgreSQL full-text search by default and may optionally blend in
+    | trigram similarity when the pg_trgm extension is available.
```
pushpak1300:
```suggestion
    | Here you may configure Scout's native PostgreSQL search driver. The driver
    | uses PostgreSQL full-text search by default and may optionally blend in
    | trigram similarity scoring when the pg_trgm extension is available.
```

config/scout.php
WendellAdriel: It makes sense.
I'll ask Joe/Taylor input on this, I think they'll be the best to check on it.

pushpak1300 (approved):

PR #1001, opened 2026-07-06
Allow job timeout and fail on timeout to be configured

### LINKED ISSUE #957: When scout encounters timeout, the jobs never fail in Horizon and block the queue
### Scout Version

10.22.1

### Scout Driver

Typesense

### Laravel Version

12.42.0

### PHP Version

8.4.1

### Database Driver & Version

_No response_

### SDK Version

_No response_

### Meilisearch CLI Version

_No response_

### Description

I encountered a weird issue recently.

I commented my Typesense credentials in my `.env` and it of course failed to send scout requests on deployment (no surprise here).

But what I found out is that because of how the `MakeSearchable` job is configured, when the job receives a timeout, it never fails and is still pending in Horizon.

As a result it just blocked my whole queue and nothing happened with queue messages accumulating.

`./artisan horizon:forget --all` didn't solve anything because all my messages were not failed but pending.

The solution I found was to clear all pending jobs and to change the default scout job:

```php
<?php

// app/Jobs/MakeSearchableWithFailure.php

namespace App\Jobs;

use Laravel\Scout\Jobs\MakeSearchable;

class MakeSearchableWithFailure extends MakeSearchable
{
    /**
     * The number of times the job may be attempted.
     */
    public int $tries = 1;

    /**
     * The maximum number of unhandled exceptions to allow before failing.
     */
    public int $maxExceptions = 1;

    /**
     * Indicate if the job should be marked as failed on timeout.
     */
    public bool $failOnTimeout = true;

    /**
     * The number of seconds the job can run before timing out.
     */
    public int $timeout = 30;
}
```

In my `AppServiceProvider`:

```php
use Laravel\Scout\Scout;

// ...

    /**
     * Bootstrap any application services.
     */
    public function boot(): void
    {
        // ...

        Scout::makeSearchableUsing(\App\Jobs\MakeSearchableWithFailure::class);

        // ...
    }
```

Maybe I wrongly configured horizon and scout, but I couldn't find anything in the scout documentation regarding this phenomenon.

I don't think a temporary timeout should block the whole queue.

### Steps To Reproduce

Remove the driver credentials in env vars, and trigger a few dozens of scout imports.

Bramvzw:
This PR extends the `scout.jobs` config options introduced in #962 with `timeout` and `fail_on_timeout`, applied to `MakeSearchable` and `RemoveFromSearch` through the existing `ConfiguresJobOptions` trait.

Fixes #957.

When a Scout job exceeds its timeout on a queue with unlimited tries (`'tries' => 0`, common in Horizon supervisor configs), the worker is killed but the job is never marked as failed: it is released back onto the queue after `retry_after` and picked up again, indefinitely. `horizon:forget` does not help because the job never reaches `failed_jobs`. That is the behavior described in #957.

I reproduced this against a fresh Laravel 12 app with the database queue driver and a deliberately slow `toSearchableArray()`:

- Without these options (`queue:work --tries=0 --timeout=5`): the worker was killed six times in 50 seconds, the job stayed pending with `attempts = 6`, and `failed_jobs` stayed empty.
- With `'jobs' => ['timeout' => 5, 'fail_on_timeout' => true]` and no `--timeout` flag on the worker: the payload carries `timeout: 5` / `failOnTimeout: true`, and the job failed after five seconds with a `TimeoutExceededException` in `failed_jobs`, unblocking the queue.

Both options default to `null`, so existing behavior is unchanged unless they are explicitly configured, and explicit properties on job subclasses take precedence over config, consistent with #962.

One caveat worth being explicit about: the job timeout relies on `pcntl_alarm`, which cannot interrupt a single blocking I/O call, since PHP only processes signals between instructions. A search engine request that hangs forever will still hang the worker regardless of these options — client-level timeouts on the engine remain the right tool for that case. What `fail_on_timeout` closes is the gap for jobs whose timeout does fire but that are then retried forever under unlimited tries.

### DIFF
--- a/src/Traits/ConfiguresJobOptions.php
+++ b/src/Traits/ConfiguresJobOptions.php
@@ -26,6 +26,20 @@ trait ConfiguresJobOptions
     public $maxExceptions;
 
     /**
+     * The number of seconds the job can run before timing out.
+     *
+     * @var int|null
+     */
+    public $timeout;
+
+    /**
+     * Indicates if the job should be marked as failed on timeout.
+     *
+     * @var bool|null
+     */
+    public $failOnTimeout;
+
+    /**
      * Configure the job.
      *
      * @return void
@@ -45,6 +59,16 @@ trait ConfiguresJobOptions
         if (! isset($this->maxExceptions) &&
             ! is_null($maxExceptions = config('scout.jobs.max_exceptions'))) {
             $this->maxExceptions = $maxExceptions;
+        }
+
+        if (! isset($this->timeout) &&
+            ! is_null($timeout = config('scout.jobs.timeout'))) {
+            $this->timeout = $timeout;
+        }
+
+        if (! isset($this->failOnTimeout) &&
+            ! is_null($failOnTimeout = config('scout.jobs.fail_on_timeout'))) {
+            $this->failOnTimeout = $failOnTimeout;
         }
     }
 }
--- a/tests/Feature/Jobs/MakeSearchableTest.php
+++ b/tests/Feature/Jobs/MakeSearchableTest.php
@@ -37,6 +37,8 @@ class MakeSearchableTest extends TestCase
     #[WithConfig('scout.jobs.tries', 3)]
     #[WithConfig('scout.jobs.backoff', [1, 5, 10])]
     #[WithConfig('scout.jobs.max_exceptions', 2)]
+    #[WithConfig('scout.jobs.timeout', 30)]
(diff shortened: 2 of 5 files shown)


taylorotwell:
I don't really want to introduce more configuration around this. Can we just set reasonable defaults, or just always have fail on timeout = true on this job?

Bramvzw:
Makes sense — opened #1002 which drops the configuration and simply defaults `failOnTimeout = true` on the Scout jobs.

PR #1005, opened 2026-07-22
Fix boolean and inequality Algolia filters

edwei06:
Fixes #1004.

## Summary

Scout's shared Algolia filter compiler currently quotes boolean values and turns string `!=` clauses into equality filters. This produces request payloads such as `is_live:'1'` and `status:'draft'`, which do not match the intended filters.

This change:

- emits booleans as Algolia's unquoted `true` / `false` literals for scalar, `whereIn`, and `whereNotIn` filters
- translates string and boolean inequality to `NOT field:value`
- preserves existing numeric comparison, empty-list sentinel, and grouping behavior
- adds matching request-payload coverage for both supported Algolia client versions

## End-user benefit

Applications can filter boolean facets and exclude string or boolean facet values without empty or inverted result sets. The change is limited to Algolia filter serialization and does not alter public APIs.

## Testing

- `vendor/bin/phpunit tests/Feature/Engines/Algolia4EngineTest.php` (Algolia 4.46.3: 17 tests, 19 assertions)
- `vendor/bin/phpunit tests/Feature/Engines/Algolia3EngineTest.php` (Algolia 3.4.2: 17 tests, 35 assertions)
- `vendor/bin/phpunit` (Algolia 4.46.3: 227 tests, 580 assertions; 1 deprecation, 17 PHPUnit notices)
- `vendor/bin/phpunit` (Algolia 3.4.2: 227 tests, 580 assertions; 1 deprecation, 17 PHPUnit notices)
- `vendor/bin/phpstan --configuration=phpstan.src.neon.dist` (Algolia 4.46.3: no errors)
- `vendor/bin/phpstan --configuration=phpstan.types.neon.dist` (Algolia 4.46.3: no errors)
- `php -l src/Engines/AlgoliaEngine.php`
- `git diff --check upstream/11.x...HEAD`

### DIFF
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -169,9 +169,9 @@ abstract class AlgoliaEngine extends Engine implements UpdatesIndexSettings
                     return '';
                 }
 
-                return '('.collect($values)->map(function ($value) use ($key) {
+                return collect($values)->map(function ($value) use ($key) {
                     return 'NOT '.$key.':'.$this->formatFilterValue($value);
-                })->implode(' OR ').')';
+                })->implode(' AND ');
             })->values();
 
         return $wheres->merge($whereIns)->merge($whereNotIns)->filter()->implode(' AND ');
--- a/src/Engines/AlgoliaEngine.php
+++ b/src/Engines/AlgoliaEngine.php
@@ -135,7 +135,15 @@ abstract class AlgoliaEngine extends Engine implements UpdatesIndexSettings
                 $operator = $where['operator'];
                 $value = $where['value'];
 
-                if (is_string($value) || $operator === '=') {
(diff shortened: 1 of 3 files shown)


src/Engines/AlgoliaEngine.php
```diff
@@ -162,13 +170,24 @@ protected function filters(Builder $builder)
                 }
 
                 return '('.collect($values)->map(function ($value) use ($key) {
-                    return 'NOT '.$key.":'{$value}'";
+                    return 'NOT '.$key.':'.$this->formatFilterValue($value);
                 })->implode(' OR ').')';
```
mos11charlie: Observation closely relating to this PRs original issue (assembling Algolia filters): I believe this line should be `})->implode(' AND ').')';` (`AND` instead of `OR`). Current `OR` behavior for `whereNotIn()` will result in returned/matched records that were intended to be excluded.

Could be argued we open a separate issue to track - can do that if you prefer!

PR #1009, opened 2026-08-24
feat(typesense): add semantic and hybrid search

tharropoulos:
Brings the Typesense engine up to parity with the Meilisearch and Turbopuffer engines for `Builder::semantic()` and `Builder::hybrid()`. Embeddings come from one of two drivers configured per model: `laravel-ai` (the default), where Scout generates vectors through the Laravel AI SDK from `toSearchableEmbedding()` or accepts precomputed ones, embedding the query at search time, or `typesense`, where the collection's native auto-embedding field does everything server-side and Scout only targets it in `query_by`.

**Laravel AI driver (default)**

```php
// config/scout.php
'model-settings' => [
    Product::class => [
        'collection-schema' => [
            'fields' => [
                ['name' => 'name', 'type' => 'string'],
                ['name' => 'embedding', 'type' => 'float[]', 'num_dim' => 1536],
            ],
        ],
        'search-parameters' => ['query_by' => 'name'],
        'embedding' => ['attribute' => 'embedding', 'dimensions' => 1536],
    ],
],

// Scout embeds this at index time, or return an array to supply your own vector...
public function toSearchableEmbedding(): string
{
    return $this->name.' '.$this->description;
}

Product::search('device to type things on')->semantic()->get();
Product::search('a relaxed pet')->options(['vector' => $vector])->semantic(minSimilarity: 0.7)->get();
```

**Native Typesense driver**

```php
// config/scout.php
'model-settings' => [
    Product::class => [
        'collection-schema' => [
            'fields' => [
                ['name' => 'name', 'type' => 'string'],
                [
                    'name' => 'embedding',
                    'type' => 'float[]',
                    'embed' => [
                        'from' => ['name'],
                        'model_config' => ['model_name' => 'ts/all-MiniLM-L12-v2'],
                    ],
                ],
            ],
        ],
        'search-parameters' => ['query_by' => 'name'],
        'embedding' => ['attribute' => 'embedding', 'driver' => 'typesense'],
    ],
],

Product::search('device to type things on')->semantic()->get();
Product::search('desktop copier')->hybrid(textWeight: 1, semanticWeight: 2)->get();
```

- implement `SupportsSemanticSearch` on `TypesenseEngine`, configured via `scout.typesense.model-settings.<model>.embedding`
- generate or accept precomputed document embeddings during indexing, batched through the Laravel AI SDK
- translate `semantic()`/`hybrid()` into `vector_query`: hybrid weights normalize into `alpha`, `minSimilarity` maps to `distance_threshold` (1 - similarity), and `options(['vector' => ...])` supplies a precomputed query vector
- keep `query_by_weights`, `num_typos`, `prefix`, and `infix` count-aligned when the embedding field replaces or extends `query_by`, and disable prefix search on it since remote embedders reject that
- append the embedding attribute to `exclude_fields` so raw vectors never come back in responses
- add unit coverage for every driver and search-mode combination plus a live integration test using user-provided embeddings

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -216,6 +216,10 @@ return [
             //     'search-parameters' => [
             //         'query_by' => 'name'
             //     ],
+            //     'embedding' => [
+            //         'attribute' => 'embedding',
+            //         'dimensions' => 1536,
+            //     ],
             // ],
         ],
         'import_action' => env('TYPESENSE_IMPORT_ACTION', 'upsert'),
--- a/src/EngineManager.php
+++ b/src/EngineManager.php
@@ -168,7 +168,8 @@ class EngineManager extends Manager
 
         return new TypesenseEngine(
             new Typesense($config['client-settings']),
-            $config['max_total_results'] ?? 1000
+            $config['max_total_results'] ?? 1000,
+            $config
         );
     }
 
--- a/src/Engines/TypesenseEngine.php
+++ b/src/Engines/TypesenseEngine.php
@@ -9,6 +9,7 @@ use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Config;
 use Illuminate\Support\LazyCollection;
 use Laravel\Scout\Builder;
+use Laravel\Scout\Contracts\SupportsSemanticSearch;
 use Laravel\Scout\Exceptions\NotSupportedException;
 use Laravel\Scout\Exceptions\ScoutException;
 use stdClass;
@@ -18,7 +19,7 @@ use Typesense\Exceptions\ObjectAlreadyExists;
 use Typesense\Exceptions\ObjectNotFound;
 use Typesense\Exceptions\TypesenseClientError;
 
-class TypesenseEngine extends Engine
+class TypesenseEngine extends Engine implements SupportsSemanticSearch
 {
     /**
      * The Typesense client instance.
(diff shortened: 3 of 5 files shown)


tharropoulos:
@taylorotwell sorry for the ping, but maybe this was missed through your notification list

ahamris:
Two findings from running the current Typesense engine against a Typesense **30.2** server with real embedding dimensions. Both seem relevant here rather than as a separate PR.

### `vector_query` will exceed Typesense's query string limit

`applySemanticSearchParameters()` puts `vector_query` into `$parameters`, which reaches the server through `Documents::search()` — a GET. Typesense caps the query string at 4000 characters:

```
Typesense\Exceptions\RequestMalformed:
Query string exceeds max allowed length of 4000. Use the /multi_search end-point for larger payloads.
```

A serialised vector runs roughly 8–13 characters per component, so the ceiling sits somewhere around 350 dimensions. The example in this PR's description uses `'dimensions' => 1536`, which produces a `vector_query` of 12–20 kB and fails with a 400 on every search. We hit this with `bge-m3` at 1024 dimensions and had to route retrieval through `/multi_search` (a POST) to get past it.

The unit tests use 2-component vectors and the integration test declares `num_dim => 2`, so the limit is never exercised. An integration test with a realistic `num_dim` would catch it — a fixture around 384 (`ts/all-MiniLM-L12-v2`, the model named in the native-driver example) is enough.

The choice isn't cosmetic: `/multi_search` accepts the same search parameters in a POST body, so the fix is mostly about where the parameters are sent rather than how they are built.

### `typesense-php` is pinned to `^4.9.3` in `require-dev`

That release is from April 2024, so CI resolves a client that predates every server feature this PR targets. Applications installing `^6.0` alongside Scout run a combination CI has never seen.

The public API `TypesenseEngine` touches is unchanged across both majors — `Client::getCollections()`, `Collection`, `Documents`, `MultiSearch`, and the `ObjectAlreadyExists` / `ObjectNotFound` / `TypesenseClientError` exceptions all keep their signatures. Diffing the public methods of `Client`, `Collections`, `Collection`, `Documents`, `Document` and `MultiSearch` between v4.9.3 and v6.0.0 shows no differences; the majors add resources rather than change existing ones.

Verified on this repository at `e2b5d9d` with `typesense/typesense-php:v6.0.0` against a Typesense 30.2 server, PHP 8.5.8:

- `tests/Unit/TypesenseEngineTest.php` — 17 passed, 54 assertions
- `tests/Integration/TypesenseSearchableTest.php` — 16 passed, 41 assertions (`--group typesense`)

Which makes it a two-line change, if it is wanted here rather than separately:

```diff
-        "typesense/typesense-php": "^4.9.3",
+        "typesense/typesense-php": "^4.9.3|^5.0|^6.0",
```

```diff
-        "typesense/typesense-php": "Required to use the Typesense engine (^4.9)."
+        "typesense/typesense-php": "Required to use the Typesense engine (^4.9|^5.0|^6.0)."
```

Happy to open that as its own PR instead if you would rather keep this one focused.

tharropoulos:
@ahamris Thank you for the keen eye on the vector query issue. I've since updated both the test and the implementation to route the request through Multisearch's POST request.

As for `typesense-php`, changing it here doesn't fit the scope of the PR. We should open up a separate one for that.


BTW, the test failing has to do with Turbopuffer: https://github.com/laravel/scout/actions/runs/33408825754/job/99543026176?pr=1009#step:5:24

taylorotwell:
Thanks!

PR #1010, opened 2026-09-03
Add Elasticsearch engine

MeRezaRezaei:
## Description

This PR adds a first-party Elasticsearch engine, following the same structure and conventions as the existing Meilisearch and Typesense engines.

**Engine (`Laravel\Scout\Engines\ElasticsearchEngine`)**

- Uses the `elasticsearch/elasticsearch` v8 client (required via `suggest`, guarded at runtime like the other optional clients).
- `update()` / `delete()` use the bulk API, indexing documents under the Scout key (`_id`) and merging the Scout key name into the document body, mirroring the Meilisearch engine.
- Searches use a `multi_match` query with `type: bool_prefix` so partial queries (e.g. `lar`) behave like the Meilisearch/Typesense engines; empty and `*` queries use `match_all`.
- `where` constraints are translated into `bool.filter` clauses (`term` / `range` / `exists`, including `whereNull` semantics), `whereIn` / `whereNotIn` into `terms` / `must_not` clauses, and `orderBy` into `sort`.
- Soft deletes rely on the `__soft_deleted` where clause injected by the Scout builder; the engine maps it to an `integer` field to match the `0` / `1` values written by `pushSoftDeleteMetadata()`.
- Pagination uses `from` / `size`, and the custom search callback receives `(, , )` with the resolved index included.
- Implements `UpdatesIndexSettings`: index settings are synced via `putSettings` / `putMapping` depending on the configured `settings` / `mappings` keys.
- `deleteAllIndexes()` deletes only indexes carrying the Scout prefix and skips dot-prefixed system indexes.

**Configuration**

`config/scout.php` gains an `elasticsearch` block (hosts + HTTP basic credentials + index settings), and the container binding builds the client via `ClientBuilder` when the package is installed.

## Tests

- **Unit** (`tests/Unit/ElasticsearchEngineTest.php`): since the ES `Client` is final, the tests drive a real client through a mocked PSR-18 handler and assert on the resulting wire-level requests (bulk NDJSON payloads, search bodies, sort, pagination, filters, soft-delete metadata, index lifecycle).
- **Feature** (`tests/Feature/Engines/ElasticsearchEngineTest.php`): container wiring and driver registration.
- **Integration** (`tests/Integration/ElasticsearchSearchableTest.php`): runs the shared `SearchableTests` suite against a live Elasticsearch service, with a new CI job mirroring the existing Meilisearch/Typesense service jobs.

The integration suite was verified locally against Elasticsearch 8.19.

Note: `TurbopufferEngineTest::test_update_upserts_models_with_schema_and_scout_ids` currently fails on pristine `11.x` as well; it is unrelated to this change.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -113,6 +113,56 @@ jobs:
           MEILISEARCH_KEY: 'masterKey'
           DB_CONNECTION: 'testing'
 
+  tests-using-elasticsearch:
+    runs-on: ubuntu-22.04
+
+    services:
+      elasticsearch:
+        image: docker.elastic.co/elasticsearch/elasticsearch:8.19.0
+        ports:
+          - 9200:9200
+        env:
+          discovery.type: single-node
+          xpack.security.enabled: false
+          ES_JAVA_OPTS: -Xms512m -Xmx512m
+          cluster.routing.allocation.disk.threshold_enabled: false
+
+    strategy:
+      fail-fast: true
+      matrix:
+          php: ['8.0', 8.1, 8.2, 8.3, 8.4]
+
+    name: PHP ${{ matrix.php }} using Elasticsearch
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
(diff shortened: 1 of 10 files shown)


taylorotwell:

